PR #1, opened 2017-01-20
Feature/remove global event dependencies

dweinber:
Removed all references to the global bitmovin.player objects to access event constants.


src/ts/uimanager.ts
```diff
@@ -117,7 +117,7 @@ export class UIManager {
              * Since this can break functionality of components that rely on this event, we relay the event to the
              * ads UI components with the following call.
              */
-            self.adsUi.getPlayer().fireEventInUI(bitmovin.player.EVENT.ON_AD_STARTED, event);
+            self.adsUi.getPlayer().fireEventInUI(this.player.EVENT.ON_AD_STARTED, event);
```
protyposis: `this` does not work in an inline function without a `bind(this)`. Use `self` instead.

dweinber: Missed that this is inline function. Fixed.

dweinber:
Right, this should be in player-events.d.ts. Moved it and removed the duplicate declaration of events.

PR #13, opened 2017-05-17
Makes the menu closign more intuitive

Succo:
Uses mouseleave event to not close when option list are open.
The main difference is that if an option list is open,
it won't close even if the mouse leaves the menu area.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Detect live streams and time shift availability when configuring dependent components (`PlaybackTimeLabel`, `PlaybackToggleButton`, `SeekBar`) to adjust their mode independently from the player state changes
 
 ### Fixed
-- Fix setting panel closing when an option select box is open
+- Fix settings panel closing when an option select box is open
 - Fix crash of Gulp `serve` task on HTML file changes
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Detect live streams and time shift availability when configuring dependent components (`PlaybackTimeLabel`, `PlaybackToggleButton`, `SeekBar`) to adjust their mode independently from the player state changes
 
 ### Fixed
+- Fix option panel closing when an option list is open
```
protyposis: Please call it "settings panel" and "option select box" (similar to the names of the affected components)

Succo: Yeah make sense, I pushed an update.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Detect live streams and time shift availability when configuring dependent components (`PlaybackTimeLabel`, `PlaybackToggleButton`, `SeekBar`) to adjust their mode independently from the player state changes
 
 ### Fixed
+- Fix setting panel closing when an option select box is open
```
protyposis: setting**s** :)

Succo: :see_no_evil:

PR #16, opened 2017-06-13
Feature/fcc compliant cc

Succo:
Ok, here it is. 

Here are some screenshot of the ui:
settingtoggleleftofmenuclean
The new button to access the settings, won't show if there is no subtitles.
ccsettings
The setting list.

The config options are illustrated (but commented) in index.html, note that for colors even thought the menu only offers 8 option, any valid rgba value can be chosen which fit the 64 values requirement. 

When available local storage is used to remember settings.

It only works with the modern UI both normal and small screen.


src/ts/components/subtitlesettings/fontfamillyselectbox.ts
```diff
@@ -0,0 +1,81 @@
...
+import {StorageUtils} from '../../utils';
+
+/**
+ * A select box providing a selection of different font familly.
+ */
+export class FontFamillySelectBox extends SelectBox {
+
+  private overlay: SubtitleOverlay;
+
+  private fontFamilly: { [key: string]: string} = {
```
protyposis: Consolidate `fontFamily`, `fontVariant` and `fontStyle` into a single object. 3 separate objects with indirectly associated lists make maintenance unnecessarily complicated.

```ts
interface FontFamily {
  name: string;
  family: string;
  variant: string;
  style: string;
}

private fontFamilies: FontFamily[] = [{
  name: 'default',
  family: 'default',
  variant: 'unset',
  style: 'unset',
}, {
  name: 'monospaced serif',
  family: '"Courier New",Courier,"Nimbus Mono L","Cutive Mono",monospace',
  variant: 'unset',
  style: 'unset',
}, ... ];
```

src/ts/components/subtitlesettings/fontfamillyselectbox.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  constructor(config: ListSelectorConfig = {}, overlay: SubtitleOverlay) {
+    super(config);
+    this.overlay = overlay;
+  }
+
+  configure(player: bitmovin.PlayerAPI, uimanager: UIInstanceManager): void {
+    super.configure(player, uimanager);
+
+    this.addItem('default', 'default');
```
protyposis: Iterate over array and add items dynamically:

```ts
for (let fontFamily of this.fontFamilies) {
  this.addItem(fontFamily.name, fontFamily.name);
}

this.selectItem(this.fontFamilies[0].name);
```

src/ts/components/subtitlesettings/backgroundcolorselectbox.ts
```diff
@@ -0,0 +1,48 @@
...
+import {ColorUtils, StorageUtils} from '../../utils';
+
+/**
+ * A select box providing a selection of different background colors.
+ */
+export class BackgroundColorSelectBox extends SelectBox {
+
+  private overlay: SubtitleOverlay;
+
+  constructor(config: ListSelectorConfig = {}, overlay: SubtitleOverlay) {
```
protyposis: Components should receive all configuration properties through the `Config` object, which means you should also pass the `SubtitleOverlay` as config property. You can make an abstract `SubtitlePropertySelectBox` with a `SubtitlePropertySelectBoxConfig` and use that as a base class for all select boxes.

```ts
export interface SubtitlePropertySelectBoxConfig extends ListSelectorConfig { 
  subtitleOverlay: SubtitleOverlay; 
}

export abstract class  SubtitlePropertySelectBox extends SelectBox {

  protected subtitleOverlay: SubtitleOverlay;

  constructor(config: SubtitlePropertySelectBoxConfig) {
    super(config);
    this.subtitleOverlay = config.subtitleOverlay;
    ...
```

src/ts/components/subtitlesettings/settinglist.ts
```diff
@@ -0,0 +1,31 @@
...
+import {SubtitleOverlay} from '../subtitleoverlay'
+import {SubtitlePanelCloser} from '../subtitlesettingtoggle'
+import {Component, ComponentConfig} from '../component'
+
+// Helper function since the subtitle option panel should not change
+let GetSubtitleSettingList = (overlay: SubtitleOverlay): Component<ComponentConfig>[] => {
+    return [
+        new SettingsPanelItem('Font', new FontColorSelectBox({}, overlay)),
+        new SettingsPanelItem('Font opacity', new FontOpacitySelectBox({}, overlay)),
+        new SettingsPanelItem('Background', new BackgroundColorSelectBox({}, overlay)),
```
protyposis: Do the FCC rules mandate these labels? If not, please apply more meaningful labels and remove duplicates (`Background`, `Window`).

src/ts/components/subtitlesettings/fontfamillyselectbox.ts
protyposis: It probably makes sense to use the array index as `key` to make lookup for `FontFamily` objects in `onItemSelected` more straightforward.

src/ts/components/subtitlesettings/settinglist.ts
dweinber: I agree, please use the names `Font family`, `Font color`, `Font size`, `Font opacity`, `Background color`, `Background opacity`, `Window color`, `Window opacity`, `Character Edge Style`.
Also grouping e.g. all font-related settings together could make sense.

dweinber:
I agree with @protyposis that using `< Back` would be a better label than `normal settings`.
Also the empty space for the caption settings on right is definitely too much and should be reduced.

However, I don't like the idea to have a separate line in the main menu for the caption styling configuration, we had this previously and didn't look good. I think the gear icon in the subtitles/captions link is the best option so far (but I'm open for other ideas!). 
@Succo could you please upload a screenshot/mockup with the gear icon on the left of the dropdown?

src/ts/components/subtitlesettings/settinglist.ts
Succo: Ok, for the name change. 

@protyposis Background and Window are two different things, the window is the overall box, containing the captions, whereas background is just "behind" the text. It's explicitly mentioned by the FCC doc.

background-window

@dweinber what do you mean grouping them? Having a sub menu for each?

protyposis: @Succo I know, I was jnust saying that `Background` and `Window` labels are both used twice.

src/ts/components/subtitleoverlay.ts
```diff
@@ -4,30 +4,111 @@ import SubtitleCueEvent = bitmovin.PlayerAPI.SubtitleCueEvent;
...
  * Overlays the player to display subtitles.
  */
-export class SubtitleOverlay extends Container<ContainerConfig> {
+export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
+
+  public config: SubtitleOverlayConfig;
+
+  private subtitleManager: ActiveSubtitleManager;
+  private forceSubtitle: boolean = false;
+  private forcedSubtitle: SubtitleLabel = new SubtitleLabel({text: 'exemple subtitle'});
```
protyposis: Typo => `Example subtitle`

Succo:
@dweinber here is a mockup with the gear on the left
gearonleft

src/ts/components/subtitlesettings/settinglist.ts
dweinber: No, not sub menus for each @Succo. Just reordering them so all `font` settings are below each other, e.g.
1. Font family
1. Font color
1. Font size
1. Font opacity
1. Character Edge Style
1. Background color
1. Background opacity
1. Window color
1. Window opacity

Currently there are two font options at the top of the list, then other settings and at the bottom again a few font options

dweinber:
Sorry, I was probably a bit unclear on the position of the gear icon @Succo. I thought about moving it between the `Subtitiles` text and the dropdown menu.

Succo:
Gotcha, at least we can compare all options.

gearonmiddle

dweinber:
I think the last one looks best, let's use that. But please make sure the bottom of the gear icon is aligned with the bottom of the text.

src/ts/components/subtitlesettings/settinglist.ts
Succo: I've done that. Indeed it was a mess.

However, "Character edge style" and "Background opacity" require increasing the label width to 9em.

dweinber: We could use `Character edge` instead of `Character edge style`.

Not sure there's a good shorter name for `Background opacity` though.

Succo: Yeah, it's frustrating, only misses a little
subsettingslist

dweinber: Any objections regarding the increase of the label width @protyposis?

protyposis: No, give them the space they need

Succo:
@protyposis I think I finished tackling your first round of comments.

I now have a specific panel for subtitles settings, and the two buttons alternate between opening one or the other. It does make the code clearer, although it is a bit cumbersome to initialize the panels first, then create the component and give them references to the panels before adding them to the panel. Same thing than with the CloseButton I guess.

Ready for the next one (unless I missed somethings)

src/ts/components/subtitleoverlay.ts
```diff
@@ -73,6 +162,125 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+  setFontVariant(fontVariant: string) {
+    this.config.fontVariant = fontVariant;
+    this.setItem('fontVariant', this.config.fontVariant);
+    this.updateSubtitleLabelCss();
+  }
+  /**
+   * A helper method to avoid updating the CSS label 3 times in a row
+   * since family, style and variant are normally updated together
+   **/
+  setFont(fontFamily: string, fontStyle: string, fontVariant: string) {
```
protyposis: Call `setFontFamily`,  `setFontStyle`, and `setFontVariant` from here with `update` parameter `false` to prevent duplicate code

src/ts/components/subtitleoverlay.ts
```diff
@@ -73,6 +162,125 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+    this.updateSubtitleLabelCss();
+    this.setItem('characterEdge', this.config.characterEdge);
+  }
+  setFontSize(coefficient: number) {
+    this.config.fontCoefficient = coefficient;
+    this.updateSubtitleLabelCss();
+    this.setItem('fontCoefficient', this.config.fontCoefficient.toString());
+  }
+
+  applyConfToDom(dom: DOM) {
```
protyposis: Make `private`. Rule: Everything is private except when required otherwise

src/ts/components/subtitleoverlay.ts
```diff
@@ -4,30 +4,114 @@ import SubtitleCueEvent = bitmovin.PlayerAPI.SubtitleCueEvent;
 import {Label, LabelConfig} from './label';
 import {ComponentConfig, Component} from './component';
 import {ControlBar} from './controlbar';
+import {ColorUtils, StorageUtils} from '../utils';
+import {DOM} from '../dom';
+
+export interface SubtitleOverlayConfig extends ContainerConfig {
```
protyposis: Simplify to 

```ts
export interface SubtitleOverlayConfig extends ContainerConfig {
  subtitleConfig?: UISubtitleConfig;
}
```

The `subtitleConfig` is optional. If specified, it takes precedence over the global UI config from `UIManager#getConfig()`

src/ts/components/subtitleoverlay.ts
```diff
@@ -4,30 +4,114 @@ import SubtitleCueEvent = bitmovin.PlayerAPI.SubtitleCueEvent;
...
+        this.config.fontCoefficient = Number(fontCoefficient);
+      }
+    }
+    // This css property isn't applied to the subtitle cue
+    // and therefore shoud be applied now
+    this.getDomElement().css('background', this.config.windowColor.toCSS());
+    this.updateSubtitleLabelCss();
   }
 
   configure(player: bitmovin.PlayerAPI, uimanager: UIInstanceManager): void {
```
protyposis: In here you can get the global UI config and extract the `UISubtitleConfig` by calling `uimanager.getConfig().subtitle`

src/ts/components/label.ts
```diff
@@ -22,7 +22,7 @@ export interface LabelConfig extends ComponentConfig {
  */
 export class Label<Config extends LabelConfig> extends Component<LabelConfig> {
 
-  private text: string;
+  protected text: string;
```
protyposis: Keep field `private` and use `getText()`/`setText(text)` methods from outside

src/ts/components/subtitlesettings/backgroundcolorselectbox.ts
```diff
@@ -0,0 +1,48 @@
...
+import {UIInstanceManager} from '../../uimanager';
+import {SubtitleOverlay} from '../subtitleoverlay';
+import {ColorUtils, StorageUtils} from '../../utils';
+
+/**
+ * A select box providing a selection of different background colors.
+ */
+export class BackgroundColorSelectBox extends SelectBox {
+
+  private overlay: SubtitleOverlay;
```
protyposis: This field and the `constructor` are similar in all of your select boxes. Please add an `abstract class SubtitleSettingSelectBox exnteds SelectBox` and use it as a common base class for all your select boxes!

src/ts/components/subtitlesettings/subtitlesetting.ts
```diff
@@ -0,0 +1,7 @@
+import {SubtitleOverlay} from '../subtitleoverlay';
+import {ListSelectorConfig} from '../listselector';
+
+
+export interface SubtitleSettingConfig extends ListSelectorConfig {
```
protyposis: Rename to `SubtitleSettingSelectBoxConfig`, implement the `SubtitleSettingSelectBox` in this file and rename the file to `subtitlesettingselectbox.ts`. Then extend all your select boxes from this class.

The general pattern is that each component gets a separate file, the file is named after the component in lower case, and the component can have an optional config interface named after the component plus "Config" suffix (without a config interface, the component uses the config interface of the parent class).

src/ts/components/subtitlesettings/backgroundcolorselectbox.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (StorageUtils.hasLocalStorage()) {
+      let color = window.localStorage.getItem('backgroundColor');
+      if (color != null) {
+        let col = ColorUtils.colorFromCss(color, ColorUtils.background);
+        col.a = 1;
+        this.selectItem(col.toCSS());
+      }
+    }
+
+
```
protyposis: Remove second empty line

src/ts/components/subtitlesettings/fontsizeselectbox.ts
```diff
@@ -0,0 +1,43 @@
...
+    this.addItem('1', '100%');
+    this.addItem('1.5', '150%');
+    this.addItem('2', '200%');
+    this.addItem('3', '300%');
+    this.addItem('4', '400%');
+
+    this.selectItem('1');
+
+    if (StorageUtils.hasLocalStorage()) {
+      let coef = window.localStorage.getItem('coef');
```
protyposis: Rename to `fontCoefficient`

src/ts/utils.ts
```diff
@@ -347,4 +347,80 @@ export namespace BrowserUtils {
...
+export namespace StorageUtils {
+  let hasLocalStorageCache: boolean;
+  // hasLocalStorage is used to safely ensure we can use localStorage
+  // taken from https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Feature-detecting_localStorage
+  export function hasLocalStorage(): boolean {
+    if (hasLocalStorageCache) {
+      return hasLocalStorageCache;
+    }
+    try {
+        var storage = window['localStorage'],
```
protyposis: `let`

src/ts/components/settingspanel.ts
```diff
@@ -76,7 +78,8 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
       // Attach marker class to last visible item
       let lastShownItem = null;
       for (let component of this.getItems()) {
-        if (component instanceof SettingsPanelItem) {
+        if ((component instanceof SettingsPanelItem)
+            || (component instanceof SubtitlePanelCloser)) {
```
protyposis: The `SettingsPanel` must not know anything about `SubtitlePanelCloser` (or any other specific components). Please find a generic solution. 

General rule: Components must not have explicit relations with any components except subcomponents that they require for their specific functionality (e.g. `SettingsPanelItem` is a subcomponent of `SettingsPanel`)

src/ts/components/subtitlesettingtoggle.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export interface SubtitleSettingsButtonConfig extends ButtonConfig {
+  subtitleSettingsPanel: SubtitleSettingsPanel;
+  settingsPanel: SettingsPanel;
+}
+
+/**
+ * A button that toggles the option menu for subtitles
+ */
+export class SubtitleSettingsOpener extends Button<ButtonConfig> {
```
protyposis: Rename to `SubtitleSettingsOpenButton` and extract to `subtitlesettingsopenbutton.ts`. To use the `SubtitleSettingsButtonConfig` for both buttons, you need to add an `abstract class SubtitleSettingsButton` inside `subtitlesettingsbutton.ts` and use that as a base class for both buttons.

General rule: Every component gets its own file (except subcomponents only used within the component where it is defined)

General rule: Components are named with their base component name as suffix, i.e. every button should be suffixed with `Button`

src/ts/components/subtitlesettingtoggle.ts
```diff
@@ -0,0 +1,130 @@
...
+    let config = <SubtitleSettingsButtonConfig>this.getConfig();
+
+    this.onClick.subscribe(() => {
+      config.subtitleSettingsPanel.show();
+      config.settingsPanel.hide();
+    });
+  }
+}
+
+export interface SubtitleSettingLabelConfig extends LabelConfig {
```
protyposis: Rename to `SubtitleSettingsLabelConfig`

src/ts/components/subtitlesettingtoggle.ts
```diff
@@ -0,0 +1,130 @@
...
+  opener: SubtitleSettingsOpener;
+}
+
+export class SubtitleSettingLabel extends Container<ContainerConfig> {
```
protyposis: Rename to `SubtitleSettingsLabel` and extract to `subtitlesettingslabel.ts` (together with the config interface)

src/ts/components/subtitlesettingtoggle.ts
```diff
@@ -0,0 +1,130 @@
...
+    );
+
+    return labelElement;
+  }
+}
+
+/**
+ * A button that toggles the option menu for subtitles
+ */
+export class SubtitlePanelCloser extends Button<ButtonConfig> {
```
protyposis: Rename to `SubtitleSettingsCloseButton` and extract to `subtitlesettingsclosebutton.ts`

src/scss/skin-modern/components/_settingspanel.scss
```diff
@@ -68,6 +68,56 @@
       display: none;
     }
   }
+
+  .#{$prefix}-ui-subtitlesettingtoggle {
```
protyposis: Rename to `.#{$prefix}-ui-subtitlesettingsopenbutton` and extract to `_subtitlesettingsopenbutton.scss`

src/scss/skin-modern/components/_settingspanel.scss
```diff
@@ -68,6 +68,56 @@
...
+    &:hover {
+      @include svg-icon-shadow;
+    }
+
+    &.#{$prefix}-on {
+      background-image: svg('assets/skin-modern/images/settingsX.svg');
+    }
+  }
+
+  .#{$prefix}-ui-subtitleoptioncloser {
```
protyposis: Rename to `.#{$prefix}-ui-subtitlesettingsclosebutton` and extract to `_subtitlesettingsclosebutton.scss`

src/ts/components/subtitlesettingspanel.ts
```diff
@@ -0,0 +1,32 @@
...
+import {SubtitlePanelCloser} from './subtitlesettingtoggle';
+import {SubtitleOverlay} from './subtitleoverlay';
+import {UIInstanceManager} from '../uimanager';
+import GetSubtitleSettingList from './subtitlesettings/settinglist';
+
+export interface SubtitleSettingsPanelConfig extends SettingsPanelConfig {
+  overlay: SubtitleOverlay;
+}
+
+export class SubtitleSettingsPanel extends SettingsPanel {
```
protyposis: Move this file into the `subtitlesettings` subdirectory

src/ts/components/subtitlesettingtoggle.ts
protyposis: Please also move all extracted components from this file into the `subtitlesettings` subdirectory

src/ts/components/subtitlesettings/subtitlesetting.ts
protyposis: Also remove one of the two empty lines above

Succo:
@protyposis I've fixed all the latest comment.

I've also made SettingsPanelItem more generic as we discussed it, and used it for the back button.

Let me know if you have anymore comments.

Succo:
Ok, this time it should be good.

I've added all components to the namespace, and fixed missing empty lines.

src/scss/skin-modern/components/_subtitlesettingsclosebutton.scss
```diff
@@ -0,0 +1,28 @@
+.#{$prefix}-ui-subtitlesettingsclosebutton {
+  @extend %ui-button;
+
+  position: relative;
+  width: stretch;
```
protyposis: Invalid value, `stretch` does not exist.

src/scss/skin-modern/_skin.scss
```diff
@@ -36,6 +36,8 @@
   @import 'components/airplaytogglebutton';
   @import 'components/pictureinpicturetogglebutton';
   @import 'components/spacer';
+  @import 'components/subtitlesettingsclosebutton';
```
protyposis: Move scss files to subfolder

src/ts/main.ts
```diff
@@ -131,4 +148,20 @@ if (typeof Object.assign !== 'function') {
   VolumeToggleButton,
   VRToggleButton,
   Watermark,
-};
\ No newline at end of file
+  // Subtitle related componenents
```
protyposis: Fix typo, move FCC related components to subpackage

src/ts/components/settingspanel.ts
```diff
@@ -134,58 +136,61 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
...
-  private label: Label<LabelConfig>;
-  private setting: SelectBox;
+  private setting: Component<ComponentConfig>;
 
   private settingsPanelItemEvents = {
     onActiveChanged: new EventDispatcher<SettingsPanelItem, NoArgs>(),
   };
 
-  constructor(label: string, selectBox: SelectBox, config: ContainerConfig = {}) {
+  constructor(setting: Component<ComponentConfig>, config: ContainerConfig = {}) {
```
protyposis: Keep the `label` as first parameter and don't render a label if `null` or `undefined` is passed in. Removing this parameter is a breaking API change that we do not want (some customers are using the framework so the API should be stable). Changing the second parameter from `SelectBox` to `Component` is ok because it widens the type and thus does not break the API.

src/ts/components/subtitleoverlay.ts
```diff
@@ -73,6 +137,193 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
     // Init
     subtitleClearHandler();
   }
+
+  /**
+   * updateSubtitleOverlayConfig updates style for the subtitle overlay based on init config
```
protyposis: Remove (wrong) method name from documentation. The documentation is attached to the function with the name, so it's an unnecessary redundancy to add it to the docs. Also please rephase to a real sentence.

src/ts/components/subtitleoverlay.ts
```diff
@@ -73,6 +137,193 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+    this.fontFamily = subtitlesConfig.fontFamily;
+    this.fontVariant = subtitlesConfig.fontVariant;
+    this.fontStyle = subtitlesConfig.fontStyle;
+    this.fontCoefficient = subtitlesConfig.fontCoefficient;
+    this.characterEdge = subtitlesConfig.characterEdge;
+    this.characterEdge = subtitlesConfig.characterEdge;
+  }
+
+  /**
+   * updateSubtitleOverlayConfig updates styles for the subtitle overlay based local storage values
```
protyposis: Fix documentation

src/ts/components/subtitleoverlay.ts
```diff
@@ -73,6 +137,193 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+   **/
+  private updateSubtitleOverlayFromLocalStorage(): void {
+    if (StorageUtils.hasLocalStorage()) {
+      let store = window.localStorage;
+      let previewSubtitleDom = this.previewSubtitle.getDomElement();
```
protyposis: Remove unused variable

src/ts/components/subtitleoverlay.ts
```diff
@@ -73,6 +137,193 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+
+  enablePreviewSubtitleLabel(): void {
+    this.previewSubtitleActive = true;
+    if (!this.subtitleManager.hasCues) {
+      this.addComponent(this.previewSubtitle);
+      this.updateComponents();
+      this.show();
+    }
+  }
+  removePreviewSubtitleLabel(): void {
```
protyposis: Separate functions with blank lines (everywhere!)

src/ts/components/subtitleoverlay.ts
```diff
@@ -1,33 +1,92 @@
...
+    this.fontStyle = '';
+    this.fontVariant = 'normal';
+    this.characterEdge = '';
+    this.fontCoefficient = 1;
+    this.fontSize = 1.2;
+
+    // Update the config first loading info from UImanager config
+    // then overwrites it with config given to the component if it applies
+    // finally loads user preferences from local storage
+    this.updateSubtitleOverlayFromUIconfig(uimanager.getConfig().subtitles);
```
protyposis: The global `UIConfig` takes precendence over the local `SubtitleOverlayConfig`, please flip the lines.

src/ts/components/subtitleoverlay.ts
```diff
@@ -73,6 +137,193 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+    if (subtitlesConfig.backgroundColor != null) {
+      this.backgroundColor = ColorUtils.colorFromCss(subtitlesConfig.backgroundColor);
+    }
+    if (subtitlesConfig.fontColor != null) {
+      this.fontColor = ColorUtils.colorFromCss(subtitlesConfig.fontColor);
+    }
+    if (subtitlesConfig.windowColor != null) {
+      this.windowColor = ColorUtils.colorFromCss(subtitlesConfig.windowColor);
+    }
+    this.fontFamily = subtitlesConfig.fontFamily;
```
protyposis: Apply all of these properties *conditionally* if set (`if (subtitlesConfig.fontFamily) { this.fontFamily = subtitlesConfig.fontFamily; }`. Currently, when setting just one property of the `UISubtitleConfig`, other unset properties would be `undefined` and overwrite the local config with `undefined`.

src/ts/utils.ts
```diff
@@ -347,4 +347,82 @@ export namespace BrowserUtils {
...
+      this.b = b;
+      this.a = a;
+    }
+
+    toCSS(): string {
+      return `rgba(${this.r}, ${this.g}, ${this.b}, ${this.a})`;
+    }
+  }
+
+  // colorFromCss is used to get a color value from localstorage since an objec can't be stored
```
protyposis: Fix documentation (remove method name, fix typo, correct sentence)

src/ts/components/subtitlesettings/subtitlesettingselectbox.ts
```diff
@@ -0,0 +1,20 @@
+import {SubtitleOverlay} from '../subtitleoverlay';
+import {ListSelectorConfig} from '../listselector';
+import {SelectBox} from '../selectbox';
+
+export interface SubtitleSettingSelectBoxConfig extends ListSelectorConfig {
+  overlay: SubtitleOverlay;
+}
+
+/**
+ * SubtitleSettingSelectBox is a base class for all subtiles settings
```
protyposis: Fix typo, remove class name from documentation

src/ts/components/subtitlesettings/windowopacityselectbox.ts
```diff
@@ -0,0 +1,38 @@
...
+export class WindowOpacitySelectBox extends SubtitleSettingSelectBox {
+
+  constructor(config: SubtitleSettingSelectBoxConfig) {
+    super(config);
+  }
+
+  configure(player: bitmovin.PlayerAPI, uimanager: UIInstanceManager): void {
+    super.configure(player, uimanager);
+
+    this.addItem('1', '100%') ;
```
protyposis: Remove spaces before trailing colons.

src/ts/components/subtitlesettings/characteredgeselectbox.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  constructor(config: SubtitleSettingSelectBoxConfig) {
+    super(config);
+  }
+
+  configure(player: bitmovin.PlayerAPI, uimanager: UIInstanceManager): void {
+    super.configure(player, uimanager);
+
+    this.addItem('none', 'none');
+    this.addItem('0px 0px 4px rgba(0, 0, 0, 0.9), 0px 1px 4px rgba(0, 0, 0, 0.9), 0px 2px 4px rgba(0, 0, 0, 0.9)', 'raised');
```
protyposis: Add line breaks to keep 120 chars text width

src/ts/components/subtitlesettings/settinglist.ts
```diff
@@ -0,0 +1,30 @@
...
+import {BackgroundOpacitySelectBox} from './backgroundopacityselectbox';
+import {WindowColorSelectBox} from './windowcolorselectbox';
+import {WindowOpacitySelectBox} from './windowopacityselectbox';
+import {CharacterEdgeSelectBox} from './characteredgeselectbox';
+import {SettingsPanelItem} from '../settingspanel';
+import {SubtitleOverlay} from '../subtitleoverlay';
+import {Component, ComponentConfig} from '../component';
+
+// Helper function since the subtitle option panel should not change
+let GetSubtitleSettingList = (overlay: SubtitleOverlay): Component<ComponentConfig>[] => {
```
protyposis: Remove the `GetSubtitleSettingList` function and file (that is a bad pattern anyway) and add the subtitle settings items directly within the `SubtitleSettingsPanel` inside the constructor via a `mergeConfig` call (analogous to other components). The list is static and does not need to be changed from the outside.

src/ts/components/subtitlesettings/backgroundopacityselectbox.ts
```diff
@@ -0,0 +1,38 @@
...
+export class BackgroundOpacitySelectBox extends SubtitleSettingSelectBox {
+
+  constructor(config: SubtitleSettingSelectBoxConfig) {
+    super(config);
+  }
+
+  configure(player: bitmovin.PlayerAPI, uimanager: UIInstanceManager): void {
+    super.configure(player, uimanager);
+
+    this.addItem('1', '100%') ;
```
protyposis: Remove spaces before colons (`) ;`)

Succo:
- I think all subtitles related files are in the subfolder apart from `subtitleoverlay` that was already existing, not sure if it should be moved.
- I've moved the component to their own namespace (unless you think I was missing some but I'm not sure which).
- I think I've caught all missing blank lines and unused imports (couldn't the linter track unused import BTW it seems the kind of job it would be better at than me).
- I've fixed the import path.

src/scss/skin-modern/_skin.scss
```diff
@@ -36,6 +36,8 @@
   @import 'components/airplaytogglebutton';
   @import 'components/pictureinpicturetogglebutton';
   @import 'components/spacer';
+  @import 'components/subtitlesettings/subtitlesettingsclosebutton';
```
protyposis: Please create a `'components/subtitlesettings/_subtitlesettings.scss` file that imports these 2 components, then just `@import 'components/subtitlesettings/subtitlesettings';` here. The idea of the subfolders and all that stuff is to make the subtitlesettings a component that could be a plugin or extension later on, so it is nice if as few as possible imports have to be managed.

src/ts/components/subtitleoverlay.ts
```diff
@@ -1,33 +1,92 @@
...
 /**
  * Overlays the player to display subtitles.
  */
-export class SubtitleOverlay extends Container<ContainerConfig> {
+export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
+
+  public config: SubtitleOverlayConfig;
+
+  private subtitleManager: ActiveSubtitleManager;
+  private previewSubtitleActive: boolean = false;
```
protyposis: Please initialize the property values inside the constructor.

src/ts/uimanager.ts
```diff
@@ -385,20 +401,44 @@ export namespace UIManager.Factory {
...
 
   function modernUI() {
+    let subtitleOverlay = new SubtitleOverlay();
+
+    let subtitleSettingsPanel = new SubtitleSettingsPanel({
+      hidden: true,
+      overlay: subtitleOverlay,
+    });
+
+
```
protyposis: Remove one empty line

src/ts/uimanager.ts
```diff
@@ -385,20 +401,44 @@ export namespace UIManager.Factory {
...
       hidden: true,
     });
 
+    let subtitleSettingsOpenButton = new SubtitleSettingsOpenButton({
+      subtitleSettingsPanel: subtitleSettingsPanel,
+      settingsPanel: settingsPanel,
+    });
+    settingsPanel.addComponent(
+      new SettingsPanelItem(
+        new SubtitleSettingsLabel({text: 'Subtiles', opener: subtitleSettingsOpenButton}),
```
protyposis: Fix typo

src/ts/uimanager.ts
```diff
@@ -474,18 +514,40 @@ export namespace UIManager.Factory {
...
       hidden: true,
       hideDelay: -1,
     });
+    let subtitleSettingsOpenButton = new SubtitleSettingsOpenButton({
+      subtitleSettingsPanel: subtitleSettingsPanel,
+      settingsPanel: settingsPanel,
+    });
+    settingsPanel.addComponent(
+      new SettingsPanelItem(
+        new SubtitleSettingsLabel({text: 'Subtiles', opener: subtitleSettingsOpenButton}),
```
protyposis: Fix typo

src/ts/uimanager.ts
```diff
@@ -581,7 +644,6 @@ export namespace UIManager.Factory {
   export function buildModernUI(player: PlayerAPI, config: UIConfig = {}): UIManager {
     // show smallScreen UI only on mobile/handheld devices
     let smallScreenSwitchWidth = 600;
-
```
protyposis: Re-add removed blank line

src/scss/skin-modern/components/subtitlesettings/_subtitlesettingsclosebutton.scss
```diff
@@ -0,0 +1,28 @@
+.#{$prefix}-ui-subtitlesettingsclosebutton {
```
protyposis: Import variables into file, else the usage of  e.g. `$color-primary` leads to warning in IDEs (`@import '../../variables';`)

src/scss/skin-modern/components/subtitlesettings/_subtitlesettingsopenbutton.scss
```diff
@@ -0,0 +1,14 @@
+.#{$prefix}-ui-subtitlesettingsopenbutton {
+  @extend %ui-button;
+
+  background-image: svg('assets/skin-modern/images/settings.svg');
+  padding: .3em;
+
+  &:hover {
+    @include svg-icon-shadow;
```
protyposis: import mixins for this include to not give a warning in IDEs

src/ts/components/subtitlesettings/subtitlesettingsbutton.ts
```diff
@@ -0,0 +1,21 @@
+import {Button, ButtonConfig} from '../button';
+import {SettingsPanel} from '../settingspanel';
+import {SubtitleSettingsPanel} from '../subtitlesettings/subtitlesettingspanel';
```
protyposis: `from './subtitlesettingspanel'`

Succo:
Ok I've fixed all the latest issues, checked and fixed unused imports.

src/ts/components/settingspanel.ts
```diff
@@ -1,9 +1,11 @@
 import {ContainerConfig, Container} from './container';
+import {ComponentConfig, Component} from './component';
 import {SelectBox} from './selectbox';
-import {Label, LabelConfig} from './label';
+import {Label} from './label';
 import {UIInstanceManager} from '../uimanager';
 import {VideoQualitySelectBox} from './videoqualityselectbox';
 import {AudioQualitySelectBox} from './audioqualityselectbox';
+
```
protyposis: Remove empty line

src/ts/components/subtitleoverlay.ts
```diff
@@ -1,33 +1,95 @@
...
-      cssClass: 'ui-subtitle-overlay',
-    }, this.config);
+    this.previewSubtitleActive = false;
+    this.previewSubtitle = new SubtitleLabel({text: 'example subtitle'});
+
+    this.config = this.mergeConfig( config, {
+        cssClass: 'ui-subtitle-overlay',
+      }, this.config);
+
+    config = this.config;
```
protyposis: Remove unused assignment

src/ts/components/subtitlesettings/subtitlesettingsclosebutton.ts
```diff
@@ -0,0 +1,46 @@
+import {SubtitleSettingsButton} from './subtitlesettingsbutton';
+
```
protyposis: Remove empty lines

src/ts/components/subtitlesettings/subtitlesettingsopenbutton.ts
```diff
@@ -0,0 +1,30 @@
+import {SubtitleSettingsButton} from './subtitlesettingsbutton';
+
```
protyposis: Remove empty lines

Succo:
Ok, should be fixed, sorry about those lines, I ran tslint to fix missing import and didn't see that it didn't remove empty lines.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Added
+- Add support for FCC compliant Closed Caption, by adding options on how captions are displayed, and a `SubtitleSettingsPanel` with the possibility to update the settings while playing the video.
```
protyposis: Add support for FCC compliant closed captions. Adds options on how captions are displayed, and a `SubtitleSettingsPanel` with the possibility to update the settings while playing the video.

Succo:
Should be good now.

Succo:
It was meant as a feature because when setting a background color you probably wanted to see it. 

Ideally it would switch to the proper setting in the opacity select box,  I was afraid adding this option would add too much complicated coupling between components for small gains. But if you think it's not clear I can add it.

protyposis:
That is a nice thought but the problem is that it stays this way (e.g. if it would be only visible while hovering the selectbox, it would be ok, but we won't implement that now). 

When opacity is at 0%, it should not be visible. I guess that almost nobody is going to use this feature anyway and people who need it will be used to such settings and will figure out that they need to increase the opacity to get the background visible. Don't think this will be an issue, and we can always improve it later.

protyposis:
As discussed elsewhere, there is still the issue that the settings toggle button is not aware of the subtitle settings panel. When the subtitle settings panel is open, and the settings toggle button pressed, the settings panel opens behind the subtitle settings panel and both are visible at the same time.

We have a feature branch where the settings toggle button is made aware of all panels, but that is a hack I want to avoid: https://github.com/bitmovin/bitmovin-player-ui/tree/Feature/CloseButtonExperiment

A clean solution would be to make the settings panel pageable, i.e. to register pages (`SettingsPanelPage extends Container`) with settings (one page with the default settings, a second page with the subtitle settings), and to add an API that can switch pages (e.g. `settingsPanel.page(pageId: string)`).

PR #20, opened 2017-06-22
Hide toggleplaybutton onPlayEvent

bitmovin-kenny:
As i reverted the ON_PLAY event to the old functionality the togglePlayButton will be hidden onPlay so the user has feedback that his command was accepted.
Added a listener for the new Warning which occurs after the player is ready to play (full buffer) but playback is not starting (mobile devices without autoplay)

### DIFF
--- a/src/ts/components/playbacktogglebutton.ts
+++ b/src/ts/components/playbacktogglebutton.ts
@@ -3,6 +3,7 @@ import {UIInstanceManager} from '../uimanager';
 import PlayerEvent = bitmovin.PlayerAPI.PlayerEvent;
 import {PlayerUtils} from '../utils';
 import TimeShiftAvailabilityChangedArgs = PlayerUtils.TimeShiftAvailabilityChangedArgs;
+import WarningEvent = bitmovin.PlayerAPI.WarningEvent;
 
 /**
  * A button that toggles between playback and pause.
@@ -33,16 +34,31 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
         return;
       }
 
-      if (player.isPlaying()) {
+      // always hide the button on the onPlay Event as the user should get feedback on the press of the play button
+      if (player.isPlaying() || (event && event.type === player.EVENT.ON_PLAY)) {
         this.on();
       } else {
         this.off();
       }
     };
 
+    let playbackWarningHandler = (event: WarningEvent) => {
+      if (event && event.code === 5007) {
+        // Warnings[5007] === 'Playback could not be started, user interaction required'
+        // if playback did not start although the buffer was full pause the video (might not have any effect) and display
+        // the play button again as the most probable explanation is disabled autoplay on mobile devices
+        player.pause();
+        this.off();
+      }
+    };
+
     // Call handler upon these events
     player.addEventHandler(player.EVENT.ON_PLAY, playbackStateHandler);
+    player.addEventHandler(player.EVENT.ON_PLAYING, playbackStateHandler);
     player.addEventHandler(player.EVENT.ON_PAUSED, playbackStateHandler);
+
+    player.addEventHandler(player.EVENT.ON_WARNING, playbackWarningHandler);
+
     // when playback finishes, player turns to paused mode
     player.addEventHandler(player.EVENT.ON_PLAYBACK_FINISHED, playbackStateHandler);
     player.addEventHandler(player.EVENT.ON_CAST_STARTED, playbackStateHandler);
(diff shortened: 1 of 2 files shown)


bitmovin-kenny:
It's just the warning,  pause is not called in this case.

protyposis:
What was the reason for this decision?

src/ts/components/playbacktogglebutton.ts
```diff
@@ -33,16 +34,31 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
...
+      // always hide the button on the onPlay Event as the user should get feedback on the press of the play button
+      if (player.isPlaying() || (event && event.type === player.EVENT.ON_PLAY)) {
         this.on();
       } else {
         this.off();
       }
     };
 
+    let playbackWarningHandler = (event: WarningEvent) => {
+      if (event && event.code === 5008) {
```
protyposis: As this handler always receives an event object, there is no need for the first check (`event && ...`).

src/ts/components/playbacktogglebutton.ts
```diff
@@ -33,16 +34,31 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
...
-      if (player.isPlaying()) {
+      // always hide the button on the onPlay Event as the user should get feedback on the press of the play button
+      if (player.isPlaying() || (event && event.type === player.EVENT.ON_PLAY)) {
         this.on();
       } else {
         this.off();
       }
     };
 
+    let playbackWarningHandler = (event: WarningEvent) => {
```
protyposis: Please move this handler and the belonging `addEventHandler` call below the `addEventHandler` calls for `playbackStateHandler` to keep associated functionality grouped together.

src/ts/components/playbacktogglebutton.ts
```diff
@@ -33,16 +34,31 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
...
+      if (event && event.code === 5008) {
+        // Warnings[5008] === 'Playback could not be started, user interaction required'
+        // if playback did not start although the buffer was full pause the video (might not have any effect) and display
```
protyposis: Please start sentences with uppercase letters and add punctuation to make it more readable (e.g. `... although the buffer was full, pause ...`)

dweinber:
Pause should be called and the warning in addition.

protyposis:
Then you don't need to listen to `ON_WARNING` at all and this whole PR can be scrapped.

bitmovin-kenny:
I will implement the call of pause() then. 
I did not implement it, because somewhen during development, i had a test where a buffering player came out of buffering and went to pause. Did not happen with the current implementation.

protyposis:
No need to implement anything, that is the default behavior.

bitmovin-kenny:
Added pause call in the case of failed autoplay here: https://github.com/bitmovin/bitdash/pull/2254. This pull request is obsolete with that change

protyposis:
@bitmovin-kenny I edited your previous comment. Please do not link from a public to a private repo. If necessary, do it the other way round.

bitmovin-kenny:
Actually there is still a small adjustment to be made in the UI:
in `src/ts/components/playbacktogglebutton.ts` :
```
-      if (player.isPlaying()) {
 +      // always hide the button on the onPlay Event as the user should get feedback on the press of the play button
 +      if (player.isPlaying() || (event && event.type === player.EVENT.ON_PLAY)) {
```

otherwise the Button would not disappear on click. The rest of the changes can be trashed.

protyposis:
@bitmovin-kenny I think that this change should also not be necessary. We should discuss this in tomorrow's meeting if it makes sense to return `true` from `isPlaying()` between `ON_PLAY` and `ON_PLAYING`. Since the player always returned true after the `ON_PLAY` event (except v7.2.0), this might be considered a breaking API change.

protyposis:
@bitmovin-kenny @dweinber I have been thinking about this and I came to the conclusion that `player.isPlaying()` should return `true` between `ON_PLAY` and `ON_PLAYING`. This keeps the 7.1 API behavior where `isPlaying()` returned `true` after calling `play()` and the `ON_PLAY` event. If playback does not start, e.g. because of autoplay on an unmuted mobile device, the player will fire `ON_PAUSED`, and the `ON_PAUSED` should only fire if the player is playing. 

I know this is not entirely correct, since the player is not really playing during this phase, but it retains the API behavior and the `ON_PLAYING` event is "just a nice add-on". We should change the behavior and break the API in player 8.

What do you say?

bitmovin-kenny:
From my relatively naive point of view i would expect the player to return `true` from `isPlaying` where there is actual playback. 
The native player returns `true` when its status is either `STALLED` or `PLAYING`. It does not have a play status and would behave differently here.

protyposis:
How does it behave differently? Both players should behave the same on our API interface.

bitmovin-kenny:
As i implemented that the player will return `isPlaying() === true` during `ON_PLAY` and `ON_PLAYING` this PR can be scrapped

PR #22, opened 2017-06-26
Set thumbnail size from CSS

protyposis:
The thumbnail size was set programmatically in pixels and was directly taken from the sprites file, so it could only be changed by adjusting the sprites file. 

Now the thumbnail size can be a relative or absolute size. It's not programmatically set any more so it can be set in CSS too.

### DIFF
--- a/src/ts/imageloader.ts
+++ b/src/ts/imageloader.ts
@@ -1,0 +1,80 @@
+import {DOM} from './dom';
+
+export interface ImageLoadedCallback {
+  (url: string, width: number, height: number): void;
+}
+
+interface ImageLoaderState {
+  url: string,
+  image: DOM;
+  loadedCallback: ImageLoadedCallback;
+  loaded: boolean;
+  width: number;
+  height: number;
+}
+
+/**
+ * Tracks the loading state of images.
+ */
+export class ImageLoader {
+
+  private state: { [url: string]: ImageLoaderState; } = {};
+
+  /**
+   * Loads an image and call the callback once the image is loaded. If the image is already loaded, the callback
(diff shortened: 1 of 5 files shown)


src/ts/imageloader.ts
```diff
@@ -0,0 +1,80 @@
...
+        image: new DOM('img', {}),
+        loadedCallback: loadedCallback,
+        loaded: false,
+        width: 0,
+        height: 0,
+      };
+      this.state[url] = state;
+
+      // We need to add the image element to the DOM to get the size of the loaded image
+      new DOM('body').append(state.image);
```
HalfbyteHeroes: I think we should set the `display` of the img-tag to `none` before appending it to the body to make sure that it is never visible to the user, even when it is directly removed in `onload`.

protyposis: That does not work. The element is never rendered and width/height stay zero. The whole purpose of this loader is to get the image size. I checked it in every browser and the image is never visible.

HalfbyteHeroes: Have you tried something like [this](https://stackoverflow.com/questions/11442712/get-width-height-of-remote-image-from-url)?
```javascript
function getMeta(url, callback) {
    var img = new Image();
    img.src = url;
    img.onload = function() { callback(this.width, this.height); }
}
getMeta(
  "http://snook.ca/files/mootools_83_snookca.png",
  function(width, height) { alert(width + 'px ' + height + 'px') }
);
```

protyposis: Thanks, works nicely :)

PR #23, opened 2017-06-26
Adds the option to alway show the controls

Succo:
So it pretty much bypass all the logic in `configureUIShowHide` and just fires a `onControlsShow` event as soon as the ui is configured (otherwise component are not subscribed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
-- Add an option to keep the UI always visible by setting the UIContainerConfig#hideTimeout to -1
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Added
+- Add an option to keep the UI always on by setting the hideTimeout to -1
```
protyposis: Please stick to the pattern used in this document so a reader can immediately see which component this config option belongs to: Add an option to keep the UI always visible by setting `UIContainerConfig#hideTimeout` to `-1`

Succo:
Should be good now

PR #25, opened 2017-06-28
Custom UI container

protyposis:
Add `UIConfig#container` config property to specify a custom place in the DOM where the UI will be put into. Can be used to place it somewhere else beside the default player figure.


Succo:
LGTM, I guess the main issue if that if the caller is not careful he can easily break the UI by not passing a proper selector.

protyposis:
If someone plays around with this config property, I must assume he either knows what he is doing or he needs to expect breaking the UI. In the default case it works as previously, that is most important.

PR #27, opened 2017-06-29
Using the new issuer parameter in Events

bitmovin-kenny:
There is an open Pull Request at the player which enables issuer parameters for a multitude of evens:
- mute
- unmute
- volumeChange
- seek
- timeshift

This PR makes use of these new parameters.

### DIFF
--- a/src/ts/components/adclickoverlay.ts
+++ b/src/ts/components/adclickoverlay.ts
@@ -35,7 +35,7 @@ export class AdClickOverlay extends ClickOverlay {
 
     this.onClick.subscribe(() => {
       // Pause the ad when overlay is clicked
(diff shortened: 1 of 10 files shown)


bitmovin-kenny:
I will change it to 'ui' then, i just did it to stay relatively consistent with the `ui-button` from `playbacktogglebutton.ts`. I will change those too then.

protyposis:
Yes please change. No idea why there was a `ui-button` in the first place.

protyposis:
@bitmovin-kenny Is there any reason why this isn't merged yet?

bitmovin-kenny:
Because the pull request in the player wasn't merged yet.

PR #28, opened 2017-07-10
Split utils into to separate files

protyposis:
Splits all utils (`StringUtils`, `BrowserUtils`, etc...) from `utils.ts` into separate files (`stringutils.ts`, `browserutils.ts`, ...).

This is necessary to minimize loading of imported modules. When importing a util from `utils.ts`, e.g. `import {StringUtils} from 'utils.ts';`, all utils inside `utils.ts` would load and execute. This is a problem in certain environments where not all required functionality is available.

Example: `EventDispatcher` needs and imports `ArrayUtils`. When using `EventDispatcher` in a `Node.js` app, it would import `ArrayUtils` from `utils.ts`, which would also implicitely load and execute `BrowserUtils` from the same file, which accesses the `navigator` object, which is not available in Node and therefore fails and breaks the application. To avoid this, we can either apply a hack and define `global.navigator = {};`, or split the utils into separate files to avoid loading of unnecessary modules, which is what this PR does.


Succo:
Package that import this one and require utils would break with this change. (like restarttv ui for example)

It would make sense to consider that with either:

- A warning in the readme change set
-Reexporting the dependencies inside utils.ts until the next major release to avoid "breaking change"

protyposis:
That is an important point and I was hoping someone would bring that up. Also thanks for hinting at the reexport. The question is, which API needs to stay stable, and what is part of this API. 

Surely the public API of UIManager and its components, which users of the library need to use to build and run a custom UI or factory UI. These are all the classes that are exported into the public JS namespace in `main.ts`.

Things like the various utils, event dispatcher, timeout etc. are conceptually internal, would not need to be exported for the library to work, and are only partially exported into the public JS namespace (the utils are exported for some unknown reason). They are still exported as TS modules (a) as a convenience, so other packages can reuse them (e.g. the testing framework), and (b) because I didn't know how to make them importable into other files without exposing them to the "outside".

Considering them internal, this PR would not be a breaking change, but one can argue that the utils are part of the public API since they're importable as TS modules.

This time it is easy to solve as we can just reexport all utils from `utils.ts` to keep the "internal" API stable. Unfortunately exports cannot be marked as deprecated, i.e. signal that the import should use another module (the new one). But this is not the first time and probably not the last time this issue comes up and I think we should clearly define what the public API is that needs to stay stable.

- Does the public API consist of all the classes exported to the public JS namespace, or of all the exports from TS modules?
- Does the public API only cover the names/signatures of exported classes, functions etc., or does it also cover the module names they are exported from?

protyposis:
Due to lack of feedback, I'm deciding on the following rules for API stability:

 * The JS API **must** keep stable
 * The TS API (including module/file names) **should** keep stable, but it's not an absolute necessity because this is "just" the source from which the actual compiled UI library is generated

PR #29, opened 2017-07-12
Fix Container#removeComponents

protyposis:
`Container#removeComponents` did not reliably remove all components if the container contained multiple components. The list of components was iterated and modified at the same time, leading to skipped components that were not removed. 

This issue became noticeable on the subtitle overlay when multiple subtitle cues were displayed at the same time but not all got cleared on a seek.

The fix is to iterate a copy of the components list while modifying the original list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Develop]
+
+### Fixed
+- Fix clearing of container components with `Container#removeComponents` (fixes sticky/duplicate subtitle issue)
+
 ## [2.7.1]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


Succo:
I'm still new to this framework so I might be missing obvious things, but

- Why can't we just to `this.config.components = []`?
- Shouldn't we release removed components?

protyposis:
This is to make sure `removeComponent` is called on every component, which might sooner or later do additional stuff (like releasing the components - good point!). Just resetting the array would not call it. At the moment it would not matter, true; it's just thought ahead for future requirements.

unki2aut:
@protyposis how about something like @Succo suggested?
```typescript
removeComponents(): void {
  this.getComponents().forEach((component) => component.release());
  this.config.components = [];
}
```

protyposis:
Thanks for the input, I'm still going to leave this as-is for now. It's not an issue right now and I think it's a nice future-proof pattern, ready for all kinds of extensions.

 - It's not a performance issue right now, the `SubtitleOverlay` and `RecommendationOverlay` are the only containers that dynamically adds and remove child components. If somebody wanted to add and remove hundreds of components with high frequency, this wouldn't be the right approach anyway (better to directly work directly with DOM elements then).
 - If array clearing ever gets an issue, another approach might be to just mark components as removed and only clear the array every once in a while (kind of garbage collection)
 - I'm not releasing components because they are also never initialized and configured. Only components added before UI startup (e.g. statically added at construction time) are initialized and configured, and only components that are present at UI shutdown will be released. That is because the component lifecycle is currently managed by the `UIInstanceManager`. This is definitely flawed because it can happen that components are added before startup, removed before shutdown, and thus never get released, possibly leaking things like event dispatchers or timeouts. This is an unrelated issue that must be fixed separately.

PR #30, opened 2017-07-13
Feature/subtitle positionning

Succo:
This changes the way subtitles are positioned in general, it's the cleanest way I found to make this work.

Basically subtitles `.ui-subtitle-overlay` now has fixed dimensions

- Normal subtitles are positioned inside it using flex, at the bottom and centered, as before
- Subtitles with cue position, are positioned inside it using the `left` and `top` attribute

As mentioned in #213 it's not really possible to exactly use the precise positioning as we don't have monospace fixed size fonts for our subtitles. This code however makes it work well enough to be able to move subtitles to the left, right top or bottom if needed. 
It will conflict for people using extreme value in the FCC options however, but not much can be done for that.


src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -6,14 +6,22 @@
 
   @include hidden;
 
-  bottom: 2em;
+  align-items: flex-end;
+  display: flex;
```
protyposis: Does this work in all browsers? IE9/10/11? (no subtitles in IE9 I guess?)

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -6,14 +6,22 @@
...
   font-size: 1.2em;
-  left: 3em;
+  height: 86%;
+  justify-content: center;
+  margin: auto 10%;
   pointer-events: none;
   position: absolute;
-  right: 3em;
-  text-align: center;
```
protyposis: The removed text-alignment leads to wrapped text not being centered any more
image

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -29,7 +37,7 @@
   }
 
   &.#{$prefix}-controlbar-visible {
-    bottom: 5em;
-    transition: bottom $animation-duration-short ease-in;
+    height: 70%;
```
protyposis: Why does the relative height here lead to an absolute positioning regarding the control bar, no matter the player height?

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -6,14 +6,22 @@
...
-  left: 3em;
+  height: 86%;
+  justify-content: center;
+  margin: auto 10%;
   pointer-events: none;
   position: absolute;
-  right: 3em;
-  text-align: center;
-  transition: bottom $animation-duration-short ease-out;
+  top: 50%;
```
protyposis: Why did you change to this top/translate centering "hack"? What is the advantage?

src/ts/player-events.d.ts
```diff
@@ -473,11 +473,14 @@ declare namespace bitmovin {
       html?: string;
       region?: string;
       regionStyle?: string;
-      position?: string;
```
protyposis: The position property in the player is also defined as `string`. Is that wrong? Do we need to adjust that in the player API? Or can it be a string too in certain cases?

src/ts/components/subtitleoverlay.ts
```diff
@@ -85,6 +85,13 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
 
       this.applyConfToDom(labelToAdd.getDomElement());
 
+      if (event.position != null) {
+        let domElement = labelToAdd.getDomElement();
+        domElement.css('position', 'fixed');
+        domElement.css('left', `${event.position.column}em`);
```
protyposis: Does this work to right-align a subtitle?

Succo: Not really we discussed it  a bit with @dweinber, comparing with jwplayer and youtube implementation.

It's not really possible to properly align subtitles, because we don't use monospaced fonts.

src/ts/player-events.d.ts
Succo: Hum that seems wrong based on what is described in the issue I linked you, unless there some other case I'm not aware of, could be worth asking @jmsn

src/scss/skin-modern/components/_subtitleoverlay.scss
Succo: Ok, I didn't think of that, will fix it.

src/scss/skin-modern/components/_subtitleoverlay.scss
Succo: Before `ui-subtitle-overlay` had no intrinsic height unless a `subtitle-label` was inserted in it. 

The best solution I could think of to be able to position the subtitles was to make the `ui-subtitles-overlay` div of fixed dimension relative to the player, center it, and then position the `subtitle-label` inside it using `top` and `left`. If I kept the `ui-subtitles-overlay` as before `left` and `top` would not have worked properly.

Then I tried different options to center the `ui-subtitles-overlay` both vertically and horizontally, this is the cleanest I found, but there maybe other solutions.

src/scss/skin-modern/components/_subtitleoverlay.scss
Succo: Not sure I understand the question.

Are you saying that in some case the control bar and the overlay could superpose each other? If that's the issue, in my testing I found that those case were rare, (really small player) and the default was not really better as it tended to superpose with the description.

I wished I was able to not use a fixed height and just have a `bottom` and `top` property but that didn't work for me.

src/ts/components/subtitleoverlay.ts
protyposis: So that basically means that CEA-608 is incompatible to the FCC requirements. To solve the font problem, we could create a grid and render letters separately into it... it'd work with any font, but you couldn't control the font size any more. 

Maybe we shouldn't allow FCC settings on CEA-608 captions, or limit the settings to those that do not interfere (e.g. color settings)? We could also apply a monospaced default profile to the FCC settings to get correct rendering. I'm not really happy with the current solution (not your fault!) and I'm going to open an issue about potential improvements.

src/scss/skin-modern/components/_subtitleoverlay.scss
protyposis: As long as there's no label inserted, why does it matter that it doesn't have a height in that case? Did relative positioning not work?

src/scss/skin-modern/components/_subtitleoverlay.scss
protyposis: You are using a relative height here, so I was assuming that the distance from the control bar to the subtitle would change depending on the height of the player (because 70% of the player height converted to pixels changes with the player height). I tested that and observed that the distance is always the same though, as if you specified the position absolutely (like it was before in `em`). 
And like it often happens while discussing an issue, while writing this I just remembered that the percentage is taken from the player width, even for the height property. That explains why the offset from the control bar didn't change, no matter what height I set the player to :)

protyposis:
I just retested with the Sintel stream (without positioning data), and the positioning is off:

image

vs

image

It gets worse (more noticeable) the larger the player window is.

protyposis:
With positioning data:
image

src/ts/components/subtitleoverlay.ts
Succo: Ok, I agree that there is incompatibility and this solution is far from perfect. 

To properly separate the two we would need this info before, since from the UI we only know if CEA-608 happen when the caption arrives.

There is more context and discussion in the related issue.

protyposis:
CEA-608 rendering
-----------------------------

CEA-608 subtitles are to be rendered "Uppercase white text on black box background" in a 32x16 grid ([source](http://www.3playmedia.com/2015/08/11/whats-the-difference-between-cea-608-line-21-captions-cea-708-captions/)). 

Please make sure that CEA-608 captions are rendered correctly according to the standard. You can use a monospaced font if that makes it easier. Here are also some hints on [CSS styling](https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html). As the subtitle styling feature only applies to CEA-708 captions (which we do not really support yet), I propose that you configure a fitting default style for CEA-608 so they are positioned and rendered correctly. If a user wants to play with the subtitle settings, and by that screws up subtitle rendering, it is his own fault. For now it is only important that they are rendered correctly with the default settings. You could add a reset button to the subtitle settings to revert screwed up settings.

So the styling should be left as-is for all subtitles except for CEA-608 captions, where the style should be switched to monospaced uppercase text and the positioning should be correctly applied.

The `SubtitleCueEvent` will get a `type` or `kind` property, that tells us which kind of subtitle we are dealing with. For the moment, you can assume that cues with a `position` are CEA-608 captions.

src/ts/components/subtitleoverlay.ts
protyposis: Instead of a separate issue I have posted a proposed solution below.

protyposis: The subtitle style should be able dynamically adapt (e.g. when switching the subtitle track from CEA-608 to something else), so detecting CEA-608 and adjusting the style when a caption arrives is fine.

Succo:
Ok, properly splitting the two case should indeed make it easier to implement.

I guess the issue is that we don't really fully support CEA-608 or 708 and try to selectively implement features from both.

Succo:
Using the `left: 0.6em` and using a monospace font as in the latest update works well regarding precise left positioning, as illustrated with the log I've temporary added. I mean that two text element will be perfectly positioned one relatively to the other.

The issues are making sure that the `em` base size scales such as there is always 32 character fitting in the width on the `subtitle-overlay` otherwise absolute alignment can show issue depending on the player size.

@protyposis I can't reproduce the case with normal subtitles not being centered, do you have more details?

src/ts/components/subtitleoverlay.ts
```diff
@@ -86,6 +86,20 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
 
       this.applyConfToDom(labelToAdd.getDomElement());
 
+      // In this block all the logic for CEA-608 captions is applied
+      // It's based on the assertion that a cue with position metadata
+      // is CEA-608
+      if (event.position != null) {
+        let domElement = labelToAdd.getDomElement();
+        domElement.css('position', 'fixed');
```
protyposis: As discussed offline, please move all static styles into a stylesheet. Add a marker class `cea-608` that carries the styles, and add the marker when CEA-608 captions are detected (and remove it otherwise). An examples for adding a marker class can be found in `PlaybackTimeLabel`. Add the styling into `subtitlesettings/_subtitleoverlay-cea608.scss` and apply the marker directly to subtitle overlay (instead of each `SubtitleLabel`).

src/ts/components/subtitleoverlay.ts
```diff
@@ -86,6 +86,20 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
...
+        domElement.css('left', `${event.position.column * 0.6}em`);
+        // 6.66 = 100/15 the number of possible lines
+        domElement.css('top', `${event.position.row * 6.66}%`);
+        domElement.css('text-transform', 'uppercase');
+        domElement.css('font-family', '"Courier New",Courier,"Nimbus Mono L","Cutive Mono",monospace');
+        console.log(event.position.column, event.position.column + event.text.length, event.text);
```
protyposis: Remove console logging

protyposis:
@Succo Due to the fixed character grid of the CEA-608 captions, the font size needs to scale with the video size for the monspaced font to always fit the grid. Would it be possible to calculate the required font size and adjust it with `ON_PLAYER_RESIZE` events?

src/ts/components/subtitleoverlay.ts
Succo: Yes of course, just a temporary helper

src/ts/components/subtitleoverlay.ts
Succo: Ok, I guess the lest and top property still need to be added directly as CSS, I don't think creating 31 classes is a good idea.

protyposis: Yes they can be directly added

Succo:
@protyposis  I'm looking into this, it seems complicated but doable.

The main challenge is that font size applies to the height of the font and the width of the font depends on the specific font implementation (at least it's constant with a monospace font :smile: ).

So we can either assume what the font width to height ratio is (might be wrong on some devices), or calculate it. This would require getting the wdth of an element using this font, we can use an invisible element, but it might be a bit messy.

protyposis:
Then how about playing with `text-align` for horizontal positioning? We know the coordinates on the grid and the number of characters. Maybe we can convert that to relative positioning with `text-align`?

Examples:
* col 0 / 1 character: text-align left
* col 0 / 32 characters: text-align center
* col 31 / 1 character: text-align right
* col 5 / 5 characters: text-align left + left offset
* col 28 / 5 chars: text-align right
* col 27 / 5 chars: text-align right + right offset

This way we can approximate the desired position, and right aligned text will be right aligned, centered text will be centered.

Succo:
If I understand correctly text-align applies to the text inside the container element, so it would require the `subtitle-label` to have the width of the parent container.

I guess beside that it would work, but it requires a lot of if to handle all cases, and there is probably some time where bug would happen (two lines meant to be aligned to each other, not being aligned).

In the meantime I've managed to implement a dynamic width calculator if you want to take a look at it. It works pretty nicely I think, just have a small issue with height to fix.

Succo:
I think in this state it works ok enough for CEA-608.

I'm starting to wonder if it wouldn't be easier to actually have another overlay just for them.

It would ensure that we keep the actual behaviour for "normal" subtitles, while avoiding having too many hack with CEA-608 subtitles.

And it would be easier to add/remove the functionality.

What do you think @protyposis ?

src/scss/skin-modern/components/_subtitleoverlay-cea608.scss
```diff
@@ -0,0 +1,8 @@
+.#{$prefix}-ui-subtitle-overlay {
+  .#{$prefix}-cea608 {
+    font-family: monospace;
+    font-size: 3.1%;
```
protyposis: 3% font size? What's the purpose?

src/scss/skin-modern/components/_subtitleoverlay.scss
protyposis: Again the question. Did you test in all browsers, especially IE11 (IE11 has flexbox issues)?

src/ts/components/subtitleoverlay.ts
```diff
@@ -81,10 +81,55 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
     let subtitleManager = new ActiveSubtitleManager();
     this.subtitleManager = subtitleManager;
 
+    let cea608CssClass = this.prefixCss('cea608');
+    let cea608fontSize = 1;
+
+    let generateCea608Ratio = () => {
+      let label = new SubtitleLabel({
+        // One letter label used to calculate the height wifth ratio of the font
```
protyposis: Grammar/typo: `A label used to ... height/width ratio ...` ... `This works because ...`

protyposis: Please add punctiation to the whole comment block.

src/ts/components/subtitleoverlay.ts
```diff
@@ -81,10 +81,55 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
...
+        // Works because we are using a monospace font for cea 608
+        // Using a longer string increases precision due to width being an integer
+        text: 'aaaaaaaaaa',
+      });
+      let domElement = label.getDomElement();
+      domElement.addClass(cea608CssClass);
+      domElement.css('color', 'rgba(0, 0, 0, 0)');
```
protyposis: Apply multiple CSS properties at once with 

```js
domElement.css({
  'property1', 'value1',
  'property2', 'value2',
});
```

src/ts/components/subtitleoverlay.ts
```diff
@@ -81,10 +81,55 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
...
 
+    let cea608CssClass = this.prefixCss('cea608');
+    let cea608fontSize = 1;
+
+    let generateCea608Ratio = () => {
+      let label = new SubtitleLabel({
+        // One letter label used to calculate the height wifth ratio of the font
+        // Works because we are using a monospace font for cea 608
+        // Using a longer string increases precision due to width being an integer
+        text: 'aaaaaaaaaa',
```
protyposis: Move the dummy string to a variable to use the length property below instead of the hardcoded `10`

src/ts/components/subtitleoverlay.ts
```diff
@@ -81,10 +81,55 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
...
+      let domElement = label.getDomElement();
+      domElement.addClass(cea608CssClass);
+      domElement.css('color', 'rgba(0, 0, 0, 0)');
+      domElement.css('font-size', '1px');
+      this.addComponent(label);
+      this.updateComponents();
+      this.show();
+
+      // 10 is the length of the string used above
+      let width = domElement.width() / 10;
```
protyposis: `let width = domElement.width() / dummyText.length;`

src/ts/components/subtitleoverlay.ts
```diff
@@ -81,10 +81,55 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
...
+
+      this.removeComponent(label);
+      this.updateComponents();
+      if (!subtitleManager.hasCues) {
+        this.hide();
+      }
+
+      let ratio =   1 / width;
```
protyposis: spacing

src/ts/components/subtitleoverlay.ts
```diff
@@ -81,10 +81,55 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
...
+      cea608fontSize = (new DOM(player.getFigure()).width()) * (0.6 / 32) * ratio;
```
protyposis: Please move all harcoded values into static named fields (with meaningful names), e.g. 
```js
private static readonly CEA608_NUM_COLUMNS = 32;
```

src/ts/components/subtitleoverlay.ts
```diff
@@ -81,10 +81,55 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
...
+    player.addEventHandler(player.EVENT.ON_PLAYER_RESIZE, generateCea608Ratio);
+
     player.addEventHandler(player.EVENT.ON_CUE_ENTER, (event: SubtitleCueEvent) => {
       let labelToAdd = subtitleManager.cueEnter(event);
 
-      this.applyConfToDom(labelToAdd.getDomElement());
+      // In this block all the logic for CEA-608 captions is applied
+      // It's based on the assertion that a cue with position metadata
+      // is CEA-608
+      if (event.position != null) {
```
protyposis: Introduce an additional valriable to make the meaning more verbose:

```js
let isCEA608 = event.position != null;

if (isCEA608) { ...
```

src/ts/components/subtitleoverlay.ts
```diff
@@ -81,10 +81,55 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
...
+      cea608fontSize = (new DOM(player.getFigure()).width()) * (0.6 / 32) * ratio;
+    };
+    uimanager.onConfigured.subscribeOnce(generateCea608Ratio);
+    player.addEventHandler(player.EVENT.ON_PLAYER_RESIZE, generateCea608Ratio);
+
     player.addEventHandler(player.EVENT.ON_CUE_ENTER, (event: SubtitleCueEvent) => {
       let labelToAdd = subtitleManager.cueEnter(event);
 
-      this.applyConfToDom(labelToAdd.getDomElement());
+      // In this block all the logic for CEA-608 captions is applied
```
protyposis: Add punctuation

src/scss/skin-modern/components/_subtitleoverlay-cea608.scss
Succo: Just a test, need to be removed, it's overwritten in js.

Succo:
Ok, so now it works with ie11 and resize properly.

I only update the cea608 captions on resizing.

Anything else to test?

src/scss/skin-modern/components/_subtitleoverlay-cea608.scss
```diff
@@ -0,0 +1,7 @@
+.#{$prefix}-ui-subtitle-overlay {
+  .#{$prefix}-cea608 {
```
protyposis: Change to `&.#{$prefix}-cea608 ..#{$prefix}-ui-subtitle-label` so you can apply the `cea608` directly to the `SubtitleOverlay` (see comment below)

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -6,14 +6,19 @@
...
+  margin: auto 20%;
   pointer-events: none;
   position: absolute;
-  right: 3em;
   text-align: center;
-  transition: bottom $animation-duration-short ease-out;
+  top: 50%;
+  transform: translate(0%, -50%);
+  transition: height $animation-duration-short ease-out;
+  width: 60%;
```
protyposis: This won't work well when the video aspect ratio is e.g. 1:2 (portrait) and the width rather small, there's not much space left for the subtitles and 40% of the width are wasted. Please don't assume that all players are in 16:9 format.

protyposis: Also please move all styling that only concerns CEA-608 to the `cea608` style.

src/ts/components/subtitleoverlay.ts
```diff
@@ -33,9 +38,24 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
       let labelToAdd = subtitleManager.cueEnter(event);
-
+      // If there is positionning data we assume we have CEA-608 style captions
+      labelToAdd.isCEA608 = event.position != null;
+
+      // In this block all the logic for CEA-608 captions is applied.
+      if (labelToAdd.isCEA608) {
+        let domElement = labelToAdd.getDomElement();
+        domElement.css({
+          'left': `${event.position.column * 0.6}em`,
```
protyposis: Please replace all hardcoded values with constants.

src/ts/components/subtitleoverlay.ts
```diff
@@ -33,9 +38,24 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
     let subtitleManager = new ActiveSubtitleManager();
     this.subtitleManager = subtitleManager;
 
+    this.configureCea608Captions(player, uimanager);
+
     player.addEventHandler(player.EVENT.ON_CUE_ENTER, (event: SubtitleCueEvent) => {
       let labelToAdd = subtitleManager.cueEnter(event);
-
+      // If there is positionning data we assume we have CEA-608 style captions
+      labelToAdd.isCEA608 = event.position != null;
```
protyposis: Please move all of the CEA-608 related code into `configureCea608Captions`. You can do this by adding a `getCues(event: SubtitleCueEvent): ActiveSubtitleCue[]` method to the `ActiveSubtitleManager`, subscribe to the `ON_CUE_ENTER` event in `configureCea608Captions`, call `getCues` there and apply the necessary CSS properties. Take care that `configureCea608Captions` must be called after subscribing to `ON_CUE_ENTER` in `configure` for the subtitle cue objects to exist when `ON_CUE_ENTER` in `configureCea608Captions` is called.

src/ts/components/subtitleoverlay.ts
```diff
@@ -33,9 +38,24 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+      // In this block all the logic for CEA-608 captions is applied.
+      if (labelToAdd.isCEA608) {
+        let domElement = labelToAdd.getDomElement();
+        domElement.css({
+          'left': `${event.position.column * 0.6}em`,
+          // 6.66 = 100/15 the number of possible lines
+          'top': `${event.position.row * 6.66}%`,
+          'font-size': `${this.cea608fontSize}px`,
+        });
+        domElement.addClass(this.prefixCss(SubtitleOverlay.CLASS_CEA_608));
```
protyposis: Apply this class directly to the `SubtitleOverlay` instead of every `SubtitleLabel` (see comment above).

src/ts/components/subtitleoverlay.ts
protyposis: This block also contains duplicate block with the CSS application code in `configureCea608Captions`... move that to a common closure/handler that is called from both places. It does not matter that you set a few CSS properties more often (e.g. left, top), clean code is more important than such micro-optimizations.

Succo:
Ok I've made the separation between the two use cases event more explicite.

I've also moved back to using 80% of width. The issue with that is that it tended to make `font-size` too big and character on two differents line would overwrite slightly. A workaround I found is using a small `letter-spacing`, to increase the effective width of the text. Ratio calculation have been addapted.

src/scss/skin-modern/components/_subtitleoverlay-cea608.scss
```diff
@@ -0,0 +1,29 @@
+// sass-lint:disable force-element-nesting
```
protyposis: Please avoid disabling lint rules that can easily be fulfilled. That should only be done in cases where there is no (easy) solution.

src/scss/skin-modern/components/_subtitleoverlay-cea608.scss
```diff
@@ -0,0 +1,29 @@
...
+  height: 86%;
+  justify-content: center;
+  left: unset;
+  margin: auto 10%;
+  pointer-events: none;
+  right: unset;
+  top: 50%;
+  transform: translate(0%, -50%);
+  transition: height $animation-duration-short ease-out;
+  width: 80%;
```
protyposis: Increasing the width and at the same time increasing the letter-spacing does not gain us any space. It just makes the text look odd. 

Please remove the letter-spacing. I think you then end up with something similar to [this guide](https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html#positioning-in-cea-608)?

src/ts/components/subtitleoverlay.ts
```diff
@@ -13,8 +14,17 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
   private subtitleManager: ActiveSubtitleManager;
   private previewSubtitleActive: boolean;
   private previewSubtitle: SubtitleLabel;
+  private cea608lineHeight: number;
+  private isCEA608: boolean = false;
 
   private static readonly CLASS_CONTROLBAR_VISIBLE = 'controlbar-visible';
+  private static readonly CLASS_CEA_608 = 'cea608';
+  private static readonly CEA608_NUM_COLUMN = 32;
+  private static readonly CEA608_WIDTH = 0.8;
```
protyposis: Where are the values of all these constants coming from? Please add comments (similar to `CEA608_LINE_COEF `) so others can understand why they have these values.

src/ts/components/subtitleoverlay.ts
```diff
@@ -86,10 +101,96 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
       }
     });
 
+    this.configureCea608Captions(player, uimanager);
     // Init
     subtitleClearHandler();
   }
 
+  configureCea608Captions(player: bitmovin.PlayerAPI, uimanager: UIInstanceManager): void {
+
```
protyposis: remove empty line

src/ts/components/subtitleoverlay.ts
```diff
@@ -86,10 +101,96 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+        'left': `${event.position.column / ratio}em`,
+        'top': `${event.position.row * SubtitleOverlay.CEA608_LINE_COEF}%`,
+        'font-size': `${SubtitleOverlay.CEA608_LINE_TO_FONT_SIZE * this.cea608lineHeight}px`,
+      });
+
+      this.addLabel(labelToAdd);
+    });
+
+    player.addEventHandler(player.EVENT.ON_SOURCE_UNLOADED, () => {
+      this.getDomElement().removeClass(SubtitleOverlay.CLASS_CEA_608);
```
protyposis: This also needs to be done when the subtitle track changes and the new track is not a CEA-608 track.

src/ts/components/subtitleoverlay.ts
```diff
@@ -174,6 +285,21 @@ class ActiveSubtitleManager {
   }
 
   /**
+   * Returns the label associated with an already added cue.
+   * @param event
+   * @return {SubtitleLabel}
+   */
+  getCue(event: SubtitleCueEvent): SubtitleLabel {
```
protyposis: Change to `getCues()` and apply style to all cues that match the calculated id, not just the first one.

src/ts/components/subtitleoverlay.ts
```diff
@@ -36,13 +46,18 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
     player.addEventHandler(player.EVENT.ON_CUE_ENTER, (event: SubtitleCueEvent) => {
       let labelToAdd = subtitleManager.cueEnter(event);
 
-      if (this.previewSubtitleActive) {
-        this.removeComponent(this.previewSubtitle);
+      if (this.isCEA608) {
```
protyposis: Remove all CEA-608 stuff from `configure` and handle it only in `configureCea608Captions`. Adding of the label can stay here like it was. The other handler in `configureCea608Captions` should just get the existing cue(s) with `getCues` and apply the style.

src/ts/components/subtitleoverlay.ts
```diff
@@ -86,10 +101,96 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+              'font-size': `${SubtitleOverlay.CEA608_LINE_TO_FONT_SIZE * this.cea608lineHeight}px`,
+            });
+          }
+        }
+      }
+    };
+    player.addEventHandler(player.EVENT.ON_PLAYER_RESIZE, generateCea608Ratio);
+
+    player.addEventHandler(player.EVENT.ON_CUE_ENTER, (event: SubtitleCueEvent) => {
+      let labelToAdd = this.subtitleManager.getCue(event);
```
protyposis: Like written above, `getCue` (which should be `getCues`) returns an already added label, so you just have to apply CEA-608 styling here. No need to add it here, so no need for the `addLabel` function. Unless something speaks against it?

src/ts/components/subtitleoverlay.ts
Succo: Don't you think showing the cue with the wrong styling and then, styling and moving it can be an issue?

src/scss/skin-modern/components/_subtitleoverlay-cea608.scss
Succo: If I remove with the letter spacing, as explained the  `font-size` required to get 32 character in 80% of the player width (the size of the box inside which cea-608 captions must be displayed).
It's too big meaning that there is no way to fit 15 lines inside 80% the the player height, especially, since we don't really have 80% of the player height available when the menu are shown. So two successive lines will overlap.

If I just reduce the `font-size` then the character width is not 1/32 of the total width and text is not properly right aligned.

src/ts/components/subtitleoverlay.ts
protyposis: That should of course not happen, hence my question below :) Still it does not make sense to extract cea608 handling to a separate configure function if it is not self-contained.

src/ts/components/subtitleoverlay.ts
Succo: Ok I assumed all tracks in a video would follow the same convention.

Succo:
Ok I've fixed all issues I understood.

I'm not sure about `letter-spacing` maybe we can discuss it tomorrow.

src/scss/skin-modern/components/_subtitleoverlay-cea608.scss
protyposis: Please change `.#{$prefix}-cea608` to `&.#{$prefix}-cea608`!

src/ts/components/subtitleoverlay.ts
```diff
@@ -86,10 +101,89 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+    let generateCea608Ratio = () => {
+      let dummyText = 'aaaaaaaaaa';
+      let label = new SubtitleLabel({
+        // One letter label used to calculate the height wifth ratio of the font
+        // Works because we are using a monospace font for cea 608
+        // Using a longer string increases precision due to width being an integer
+        text: dummyText,
+      });
+      let domElement = label.getDomElement();
+      domElement.addClass(this.prefixCss(SubtitleOverlay.CLASS_CEA_608));
```
protyposis: The class must be added to the overlay, not to the label. Please take a look at the CSS to understand why.

src/ts/components/subtitleoverlay.ts
```diff
@@ -117,13 +211,23 @@ interface ActiveSubtitleCueMap {
 
 class SubtitleLabel extends Label<LabelConfig> {
 
+  private _isCEA608: boolean = false;
```
protyposis: I think this isn't used any more, right? If so, please remove.

src/ts/components/subtitleoverlay.ts
```diff
@@ -86,10 +101,89 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
     });
 
+    this.configureCea608Captions(player, uimanager);
     // Init
     subtitleClearHandler();
   }
 
+  configureCea608Captions(player: bitmovin.PlayerAPI, uimanager: UIInstanceManager): void {
+    let ratio = 1;
+    let generateCea608Ratio = () => {
```
protyposis: Rename to `updateCea608FontSize`. This function is not *generating* anything, it is basically calculating the line height and then updating the font size with it.

protyposis:
I just freshly checked out the branch and it does not seem to work. The font size is extremely small and the subtitles never positioned.

image

Succo:
Yeah, forgetting the `&` in the css broke everything should be good now.

src/ts/components/subtitleoverlay.ts
```diff
@@ -86,10 +101,88 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+
+      let labels = this.subtitleManager.getCues(event);
+
+      if (!this.isCEA608) {
+        this.isCEA608 = true;
+        this.getDomElement().addClass(this.prefixCss(SubtitleOverlay.CLASS_CEA_608));
+        updateCEA608FontSize();
+      }
+      for (let label of labels) {
+        let domElement = label.getDomElement();
```
protyposis: Unnecessary variable declaration, can be removed (it's only used in the line below)

src/ts/components/subtitleoverlay.ts
```diff
@@ -86,10 +101,88 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+    player.addEventHandler(player.EVENT.ON_PLAYER_RESIZE, updateCEA608FontSize);
+
+    player.addEventHandler(player.EVENT.ON_CUE_ENTER, (event: SubtitleCueEvent) => {
+      if (event.position == null) {
+        return;
+      }
+
+      let labels = this.subtitleManager.getCues(event);
+
+      if (!this.isCEA608) {
```
protyposis: Add CEA-608 detection. The CEA styling is currently applied to all subtitles, no matter if they have positioning data or not.

protyposis: Also the flag is never reset.

Succo: What do you mean, remove the `CLASS_CEA_608` from the component if one label isn't `CEA-608`? That would break other label not yet removed if they are positioned.

protyposis: I mean that you are applying the CEA-608 style to every subtitle, also to non-CEA subtitles. There is no detection anywhere, e.g. `isCea608Subtitle = cue.position !== null`. Instead, you just set `isCEA608` to `true` if it is `false`.

Succo: Hum, two line above I check if `event.position` is null, so if the subtitles is non-CEA 608 I never reach this line.

If it is, then I set the class to the overlay, calculate the font size, and set the flag to add the class or calculate the font size anymore.

protyposis: Oh sorry, missed that check. Please make it more verbose like it was before

```ts
let isCea608Cue = event.position !== null;

if (!isCea608Cue) {
  // Skip all non-CEA-608 cues
  return;
}
```

src/ts/components/subtitleoverlay.ts
```diff
@@ -13,8 +14,22 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
   private previewSubtitleActive: boolean;
   private previewSubtitle: SubtitleLabel;
+  private cea608lineHeight: number;
+  private isCEA608: boolean = false;
 
   private static readonly CLASS_CONTROLBAR_VISIBLE = 'controlbar-visible';
+  private static readonly CLASS_CEA_608 = 'cea608';
+  // The number of column in a cea608 line
+  private static readonly CEA608_NUM_COLUMN = 32;
+  // 80% is the width of the space wher CEA608 captions are displayed
```
protyposis: typo

src/ts/components/subtitleoverlay.ts
```diff
@@ -13,8 +14,22 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
   private subtitleManager: ActiveSubtitleManager;
   private previewSubtitleActive: boolean;
   private previewSubtitle: SubtitleLabel;
+  private cea608lineHeight: number;
+  private isCEA608: boolean = false;
 
   private static readonly CLASS_CONTROLBAR_VISIBLE = 'controlbar-visible';
+  private static readonly CLASS_CEA_608 = 'cea608';
+  // The number of column in a cea608 line
+  private static readonly CEA608_NUM_COLUMN = 32;
```
protyposis: Rename to `CEA608_NUM_COLUMNS` (plural)

src/ts/components/subtitleoverlay.ts
```diff
@@ -13,8 +14,22 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
   private subtitleManager: ActiveSubtitleManager;
   private previewSubtitleActive: boolean;
   private previewSubtitle: SubtitleLabel;
+  private cea608lineHeight: number;
+  private isCEA608: boolean = false;
 
   private static readonly CLASS_CONTROLBAR_VISIBLE = 'controlbar-visible';
+  private static readonly CLASS_CEA_608 = 'cea608';
+  // The number of column in a cea608 line
```
protyposis: `columns`

src/ts/components/subtitleoverlay.ts
```diff
@@ -86,10 +101,90 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
     // Init
     subtitleClearHandler();
   }
 
+  configureCea608Captions(player: bitmovin.PlayerAPI, uimanager: UIInstanceManager): void {
+    let ratio = 1;
+    let updateCEA608FontSize = () => {
+      let dummyText = 'aaaaaaaaaa';
+      let label = new SubtitleLabel({
+        // One letter label used to calculate the height wifth ratio of the font
```
protyposis: typo

src/ts/components/subtitleoverlay.ts
```diff
@@ -86,10 +101,90 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+        label.getDomElement().css({
+          'left': `${event.position.column / ratio}em`,
+          'top': `${event.position.row * SubtitleOverlay.CEA608_LINE_COEF}%`,
+          'font-size': `${SubtitleOverlay.CEA608_LINE_TO_FONT_SIZE * this.cea608lineHeight}px`,
+        });
+      }
+    });
+
+    let reset = () => {
+      this.getDomElement().removeClass(SubtitleOverlay.CLASS_CEA_608);
```
protyposis: Class removal does not work, CSS prefix missing

Succo:
Yeah the intersection of cea608 and cea708 is tricky. Unless explicitly required by a client I think the safest would be to not support both.

PR #38, opened 2017-07-17
Improve SubtitleOverlay style handling

protyposis:
* Refactors, optimizes and simplifies the code
* fixes a hardcoded CSS class
* ~~~fixes flawed style application from the config~~~
* moves all styling to CSS and removes JS style config
* retains default CSS style
* adds a reset button


src/ts/components/subtitleoverlay.ts
```diff
@@ -241,104 +244,115 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
...
   setFontSize(coefficient: number): void {
-    this.fontCoefficient = coefficient;
-    this.updateSubtitleLabelCss();
-    this.setItem('fontCoefficient', this.fontCoefficient.toString());
+    this.subtitleStyleSetting.fontCoefficient = coefficient;
+    this.applyStyleToLabels();
+    this.saveSubtitleStyle();
+  }
+
+  private saveSubtitleStyle(): void {
```
Succo: Using this means that if one settings is changed by the user all other values, whether it's the normal default values or values set in the player config get stored as user preferences and will overwrite default value next time the player is used on the same domain.

It could cause issues if somebody used the player sometimes with special styling and sometimes not.

protyposis:
@Succo your identified issues should be fixed now, thanks.

I have identified additional issues:
 * `characterEdge` style is never loaded in `CharacterEdgeSelectBox`
 * colors and opacities must be split and saved separately (also split in the config), else it is not possible to configure/select a color *or* an opacity, it always saves and loads both

Can you please implement that in this branch?

Succo:
Yes, no problem.

Let me look into that.

protyposis:
I have now moved all styling back to CSS. The JS code only applies classes to the `SubtitleOverlay`. This enables customers to customize the styling via CSS and also does not override the default style.

protyposis:
@Succo please review again

src/scss/skin-modern/components/subtitlesettings/_subtitleoverlay-settings.scss
```diff
@@ -0,0 +1,134 @@
...
+    'raised': '0px 0px 4px rgba(0, 0, 0, 0.9), 0px 1px 4px rgba(0, 0, 0, 0.9), 0px 2px 4px rgba(0, 0, 0, 0.9)',
+    'depressed': 'rgba(0, 0, 0, 0.8) 0px -2px 1px',
+    'uniform': '-2px 0px 1px rgba(0, 0, 0, 0.8), 2px 0px 1px rgba(0, 0, 0, 0.8), 0px -2px 1px rgba(0, 0, 0, 0.8), 0px 2px 1px rgba(0, 0, 0, 0.8), -1px 1px 1px rgba(0, 0, 0, 0.8), 1px 1px 1px rgba(0, 0, 0, 0.8), 1px -1px 1px rgba(0, 0, 0, 0.8), 1px 1px 1px rgba(0, 0, 0, 0.8)',
+    'dropshadowed': ' 0px 2px 1px rgba(0, 0, 0, 0.8)',
+  );
+
+  $font-sizes: (
+    50: .5,
+    75: .75,
+    100: 1,
```
Succo: Defaut value used to be 1.2 em.

src/ts/colorutils.ts
```diff
@@ -7,17 +7,47 @@ export namespace ColorUtils {
     a: number;
```
Succo: Should we remove all of colorUtils? It's only used by storageUtils but the object we set don't have a `Color` anymore.

src/ts/components/subtitlesettings/fontopacityselectbox.ts
```diff
@@ -1,37 +1,41 @@
...
-    this.selectItem('1');
+    this.onItemSelected.subscribe((sender, key: string) => {
+      this.settingsManager.fontOpacity.value = key;
 
-    if (StorageUtils.hasLocalStorage()) {
-      let color = window.localStorage.getItem('fontColor');
-      if (color != null) {
-        let col = ColorUtils.colorFromCss(color, ColorUtils.foreground);
-        this.selectItem(col.a.toString());
+      // Color and opacity go together, so we need to...
```
Succo: Setting the opacity to default reset the font color. I think that's not super intuitive if you set a color and then try different opacity.

src/scss/skin-modern/components/subtitlesettings/_subtitleoverlay-settings.scss
protyposis: It still is. The default is set to `1.2em` on the `SubtitleOverlay`. These sizes are set on the `SubtitleLabel` child elements. To the final size is `1.2em * #{$font-size}em`.

src/ts/colorutils.ts
protyposis: Good point, we can remove them.

src/ts/components/subtitlesettings/fontopacityselectbox.ts
protyposis: Opacity and color always need to be set together. Due to CSS `rgba` coloring, it is not possible to *only* set the opacity (opacity cannot be set separately in CSS and we cannot "inherit" the base color) *or* the color (color cannot be set separately from the opacity).
Therefore, I select a default opacity when a color is selected, or the other way round. Afterwards, then can be set independently, i.e. color is not reset to default every time an opacity is set. Or did you make an opposite observation?

Succo: For the background and window, with a default of 0%,it works ok.

For the font where the default opacity is 100%, setting the opacity to default should not reset the color I think.

src/scss/skin-modern/components/subtitlesettings/_subtitleoverlay-settings.scss
Succo: Ok, sorry.

src/ts/components/subtitlesettings/fontopacityselectbox.ts
protyposis: It *must*, because "default" means basically "whatever is specified as default CSS style". I cannot set only the selected color and keep the default opacity, because the default opacity is specified somewhere in CSS (e.g. a `style` tag in the html document) and I cannot know its value.

Also, a default opacity of 0% does not make sense. When I select a color I want to see it. With a default opacity of 0% I can select any color, do not see anything, and am wondering if this is broken.

protyposis: Or do you have an idea how this can be achieved?

Succo: Well by a default opacity of 0% I meant that in default setting there is no background which is kind of equivalent to 0% opacity. Once a background is set it goes to 100%, but if the user click on default it's not shown anymore, which is equivalent to clicking 0%. Unless there is a style tag settings different values.

The usecase that I don't like, is 

1. The user activate subtitles
2. He changes the opacity
3. Then the color
4. Actually prefer the "normal" opacity so click on default

=> color resets to white.

But I understand there is no easy workaround, so I guess for now it's ok.

Succo:
I guess it's good to go for me.
The small usecase specified above won't impact too many users I think.

src/ts/components/subtitlesettings/fontopacityselectbox.ts
protyposis: I understand, unfortunately we need to compromise here.

A nicer solution would be to make the opacity a subsetting of the color, i.e. disable the opacity box while the color is set to "default", and enable the opacity selection while a color is selected. This way we would not need the "default" opacity option, but it is too much mork now so we'll keep it as a nice-to-have-someday.

PR #41, opened 2017-07-19
Feature/small screen mute button

Succo:
That was a really easy one.

When using the small screen UI the user must be able to at least mute/unmute the volume (no volume slider).


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Fix clearing of container components with `Container#removeComponents` (fixes sticky/duplicate subtitle issue)
 
+### Added
```
protyposis: The order is always Added > Changed > Fixed, so Added must come before Fixed.

PR #44, opened 2017-07-25
Feature/remove subtitles on unload

bitmovin-kenny:
Fixed an issue where the select controls would not correctly reset when the player was reloaded without being destroyed.

### DIFF
--- a/src/ts/components/playbackspeedselectbox.ts
+++ b/src/ts/components/playbackspeedselectbox.ts
@@ -26,5 +26,15 @@ export class PlaybackSpeedSelectBox extends SelectBox {
     this.onItemSelected.subscribe((sender: PlaybackSpeedSelectBox, value: string) => {
       player.setPlaybackSpeed(parseFloat(value));
     });
+
+    // when the player hits onReady again, adjust the playback speed selection with fallback to default 1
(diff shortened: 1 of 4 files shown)


src/ts/components/playbackspeedselectbox.ts
```diff
@@ -26,5 +26,15 @@ export class PlaybackSpeedSelectBox extends SelectBox {
...
     });
+
+    // when the player hits onReady again, adjust the playback speed selection with fallback to default 1
+    player.addEventHandler(player.EVENT.ON_READY, (): void => {
+      let playbackSpeed = String(player.getPlaybackSpeed());
+      if (!this.selectItem(playbackSpeed)) {
+        playbackSpeed = '1';
+        this.selectItem(playbackSpeed);
+      }
+      this.updateDomItems(playbackSpeed);
```
protyposis: This is automatically called by `selectItem` above and should be removed

bitmovin-kenny: You are right, i was confused because `onItemSelectedEvent` is only called when there is an actual selection, but '1' has to exist in the predefined values, so it's always called. I will remove it.

PR #49, opened 2017-07-31
Fix init components on ready

unki2aut:
Make sure `ON_READY` is triggered even if the UI is created when player is already set up.
Fixes components that rely on `ON_READY` to initialize.


protyposis:
This is not going to work correctly because the `UIManager` manages multiple UI instances through `UIInstanceManager` instances. `ON_READY` handling needs to be done in the `UIInstanceManager` for every single UI instance, and this solution only does it for the `currentUi` instance. Also, it adds a state field which is unnecessary because a UI instance is only loaded once and the ready state can be checked there.

I submitted #50 as an alternative.

unki2aut:
Close this PR in favor of #50

PR #50, opened 2017-07-31
Fix init components on ready v2

protyposis:
Fires a `ON_READY` event into the UI when the UI is loaded after the player is ready and will thus not receive an `ON_READY` from the player. This is done because some components wait for `ON_READY` to initialize their state.

### DIFF
--- a/src/ts/uimanager.ts
+++ b/src/ts/uimanager.ts
@@ -368,7 +368,7 @@ export class UIManager {
 
     // Some components initialize their state on ON_READY. When the UI is loaded after the player is already ready,
     // they will never receive the event so we fire it from here in such cases.
-    if (!player.isReady()) {
+    if (player.isReady()) {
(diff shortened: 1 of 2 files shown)


src/ts/uimanager.ts
```diff
@@ -358,12 +358,20 @@ export class UIManager {
...
+
     ui.configureControls();
     /* Append the UI DOM after configuration to avoid CSS transitions at initialization
      * Example: Components are hidden during configuration and these hides may trigger CSS transitions that are
      * undesirable at this time. */
     this.uiContainerElement.append(dom);
 
+    // Some components initialize their state on ON_READY. When the UI is loaded after the player is already ready,
+    // they will never receive the event so we fire it from here in such cases.
+    if (!player.isReady()) {
```
unki2aut: This should be done, when the player is ready, but the UI missed the `ON_READY` event.

PR #52, opened 2017-08-01
Improve Gulpfile reuse

protyposis:
Moves the JS output filename to the ``paths` config object and exports the object so it can be customized from other gulpfiles that import and reuse this project's gulpfile. Helpful for packages use this package as their base and modify the framework to create a custom build.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -32,6 +32,7 @@ var browserSync = require('browser-sync');
 var merge = require('merge2');
 var nativeTslint = require('tslint');
 var npmPackage = require('./package.json');
+var path = require('path');
 
 var paths = {
   source: {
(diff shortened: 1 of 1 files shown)


protyposis:
Good point, I'll add it

protyposis:
Are you sure you even need to redeclare the `sass` task? Don't you just have to customize the `paths.source.sass` and `paths.target.css` properties?

Succo:
Because in the sass task I add
```
.pipe(sass({
    includePaths:[path.join(__dirname, 'node_modules')]
  })
```
to be able to import from the bitmovin-player-ui files. We could also make it by default I guess.

protyposis:
What if you 

```scss
@import '../../node_modules/bitmovin-player-ui/dist/css/bitmovinplayer-ui-legacy';
```

instead?

protyposis:
Nevermind. I'll just add it by default.

Succo:
Yep it seems to work. I can't remember why we decided on using the other method.

protyposis:
It definitely looks nicer without the relative path to `node_modules`.

Succo:
ok, lgtm :+1:

PR #59, opened 2017-09-25
Feature/legacy volume seekbar fixes

Succo:
The overflow is not hidden to show the button when at the bottom
A small timeout has been added to match the scss animation

The `overflow: hidden` only applied to the legacy volume seekbar, didn't seem relevant anymore.

The timeout is because `seekBar.height()` here https://github.com/bitmovin/bitmovin-player-ui/blob/828caef2eb00fd82215d467026fe1526195f81d5/src/ts/components/seekbar.ts#L705 is always 0 when the position was calculated because the css animation was still taking place. So the position marker was alway set to the top of the volume seekbar.

Not 100% about this one @protyposis, should I only add a timeout for the legacy volume toggle the only other case where a seekbar visibility is toggled is when switching from and to livestream and the delay doesn't seem to affect it in a sensible manner.


src/scss/skin-legacy/components/_volumecontrolbutton.scss
```diff
@@ -20,7 +20,6 @@
       bottom: .5em;
       height: auto;
       left: .3em;
-      overflow: hidden;
```
protyposis: This is required, else the slider isn't centered.

src/ts/components/seekbar.ts
```diff
@@ -848,6 +848,7 @@ export class SeekBar extends Component<SeekBarConfig> {
     // value calculated from the size. This required size is not known when it is hidden.
     // For such cases, we refresh the position here in onShow because here it is guaranteed that the component knows
     // its size and can set the position correctly.
-    this.refreshPlaybackPosition();
+    // Some css component use an animation to setup, to make sure it's finished we add a timeout of 300ms which correspond to the 0.3s value (see $animation-duration in _variable.scss)
+    setTimeout( this.refreshPlaybackPosition, 300);
```
protyposis: This is a hack that can't make it into the code I'm afraid. We cannot hardcode things like the animation time that are variable and customizable by definition, as all CSS properties are. I recognize the issue here, but we need to find a proper solution.

Succo: We could try to not refresh the playback position then in the case if the volume slider ( or the legacy ui) since this is only really useful when switching from and to livestream.

But this can't be done by implementing a `onShowEvent` inside volumeslider since it still needs to to call `super.onShowEvent()`.

Succo: i guess the seekbar and the volume slider should implement a generic `slider`, or we could reuse the `configureSeek` flag from the `configure` method (making it a class property probably).

protyposis: Yes there should be a common `Slider` base class, the volume bar in it's current state is kind of a hack too.

If leaving out the `refreshPlaybackPosition` from `onShowEvent` works, you can solve it in `VolumeSlider` by overriding `refreshPlaybackPosition`:

```ts
protected refreshPlaybackPosition(refresh: boolean = false) {
  if (refresh) {
    super.refreshPlaybackPosition();
  }
}
```

And then change all `refreshPlaybackPosition()` calls in `VolumeSlider` to `refreshPlaybackPosition(true)`. The call from `onShowEvent` will then default to `refreshPlaybackPosition(true)` and not refresh the position.

Succo: Yeah that should work, although this feel even more hacky.

Succo:
@protyposis so I've updated it following your proposition.

src/scss/skin-legacy/components/_volumecontrolbutton.scss
```diff
@@ -20,7 +20,7 @@
...
+      overflow-x: hidden;
```
protyposis: it's required to hide all overflow, else I can sometimes see scroll bars (in Chrome)

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- Fix legacy volume seekbar not visible when set at the lowest
```
protyposis: "Vertical `VolumeSlider` knob in legacy skin was not visible when set to a low volume"

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
...
+- Fix legacy volume seekbar always defaulting to 100% after hiding and showing again
```
protyposis: "Legacy skin's `VolumeSlider` knob was always rendered at 100% when appearing after being hidden"

src/ts/components/volumeslider.ts
```diff
@@ -70,19 +70,29 @@ export class VolumeSlider extends SeekBar {
...
-      this.refreshPlaybackPosition();
+      this.refreshPlaybackPosition(true);
     });
 
     // Init volume bar
     volumeChangeHandler();
   }
 
+  /**
+   * Refreshes the playback position. This is needed because subclassing {@link seekBar} is a hack
```
protyposis: "Refreshes the volume slider's position. This is needed because we want to avoid the default behavior of {@link SeekBar.onShowEvent}, which is to refresh the slider position when it appears because that screws up our volume slider's position. To avoid it we override this method and add a conditional flag that defaults to `false`, avoiding the update from all calls coming from the base class ({@link SeekBar}). We can still request updates if we supply `true`."

protyposis: Lol, there is a GitHub user called link. So everytime someone posts some JavaDoc/JSDoc/etc... without a surrounding code block with an @link tag in it, they receive a notification. That must be a lot of notifications 😁

Succo: Haha poor guy hope he has good filtering setup.

Succo:
Removing the animation would fix the issue about it rendering improperly but not about it hiding at the bottom, I think this is partially because this https://github.com/bitmovin/bitmovin-player-ui/blob/b52b95b834a5c750bb44b08b193593cbcb3e9849/src/scss/skin-legacy/components/_seekbar.scss#L142 is not applied actually.

Succo:
Using a pseudo element that I translate I can kind of get it to work.

It is still half hidden in the top and bottom position, the issue is that we position the `.#{$prefix}-seekbar-playbackposition-marker` from 0 to the height of the seekbar ignoring the height of the marker...

Doesn't cause an issue with the other bar because they can overflow.

Succo:
I've updated to make the position calculation take into account the element height, this is the only solution if we can't afford to overflow I think.

src/scss/skin-legacy/components/_volumecontrolbutton.scss
```diff
@@ -8,7 +8,7 @@
   position: relative;
 
   .#{$prefix}-ui-volumeslider {
-    @include animate-slide-in-from-bottom(6em, $animation-duration / 2);
+    //@include animate-slide-in-from-bottom(6em, $animation-duration / 2);
```
protyposis: Please remove this line (and the following empty line) entirely and add an additional entry into the `Changed` section of the `CHANGELOG`:

"Removed `VolumeControlButton`'s `VolumeSlider` slide-in animation in the legacy skin to fix the slider knob at 100% bug"

(This is a change in the skin's behavior and it's not obvious from the bugfix entry)

src/ts/components/volumeslider.ts
```diff
@@ -70,19 +70,30 @@ export class VolumeSlider extends SeekBar {
...
     // Init volume bar
     volumeChangeHandler();
   }
 
+  /**
+   * Refreshes the volume slider's position.
+   * This is needed because we want to avoid the default behavior of {@link SeekBar.onShowEvent}, which is to refresh the slider position when it appears because that screws up our volume slider's position.
+   * To avoid it we override this method and add a conditional flag that defaults to false, avoiding the update from all calls coming from the base class ({@link SeekBar}). We can still request updates if we supply true.
+   */
+  protected refreshPlaybackPosition(refresh: boolean = false) {
```
protyposis: I think we can remove this hack, can't we?

src/ts/components/seekbar.ts
```diff
@@ -702,9 +702,10 @@ export class SeekBar extends Component<SeekBarConfig> {
     this.setPosition(this.seekBarPlaybackPosition, percent);
 
     // Set position of the marker
-    let px = (this.config.vertical ? this.seekBar.height() : this.seekBar.width()) / 100 * percent;
+    let totalSize = (this.config.vertical ? (this.seekBar.height() - this.seekBarPlaybackPositionMarker.height()) : this.seekBar.width());
+    let px = (totalSize) / 100 * percent;
     if (this.config.vertical) {
-      px = this.seekBar.height() - px;
+      px = this.seekBar.height() - px - this.seekBarPlaybackPositionMarker.height();
```
protyposis: Taking the height into account here introduces another issue.... the position marker is not centered around the mouse cursor any longer. Noticeable when you click at the top or bottom. You'll see that the center position of the marker and the cursor do not match.

protyposis: It's not ideal but let's leave it like this. It's definitely an improvement to the issues we had before 👍

Succo: Yes I guess it aligns just below the mouse.

Succo:
Done

PR #60, opened 2017-10-11
Fix item selection in ItemSelectionList

protyposis:
The `ItemSelectionList` rebuilt it's DOM element on every item selection only to move the `selected` class to the newly selected item. Besides being ineffective, it caused an issue that essentially broke hiding of the UI, because the rebuilt and replaced DOM below the user's mouse pointer caused an unexpected `mouseenter` event.

This PR cleans up the code a little bit and moves the `selected` class without recreating the DOM.


protyposis:
Yeah surprisingly someone used it :) That's how we found the bug.

PR #67, opened 2017-10-23
Feature/onplaying listener

bitmovin-kenny:
Added an event listener to ON_PLAYING in addition to ON_PLAY to the PlaybackToggleButton.

This is needed as in some weird edge cases the player can fire `ON_PLAY` before `ON_READY` and the UI would not have notices the event in that case and the play button would not have disappeared.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## develop
+- `PlaybackToggleButton` now also listens to ON_PLAYING in addition of ON_PLAY
+
 ## [2.10.3]
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## develop
-- `PlaybackToggleButton` now also listens to ON_PLAYING in addition of ON_PLAY
+
+### Changed
+- `PlaybackToggleButton` now also listens to `ON_PLAYING` in addition to `ON_PLAY`
 
 ## [2.10.3]
 
--- a/src/ts/components/playbacktogglebutton.ts
+++ b/src/ts/components/playbacktogglebutton.ts
@@ -43,7 +43,10 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
     // Call handler upon these events
     player.addEventHandler(player.EVENT.ON_PLAY, playbackStateHandler);
     player.addEventHandler(player.EVENT.ON_PAUSED, playbackStateHandler);
-    player.addEventHandler(player.EVENT.ON_PLAYING, playbackStateHandler);
+    if (player.EVENT.ON_PLAYING) {
+      // Since player 7.3. Not really necessary but just in case we ever miss the ON_PLAY event.
+      player.addEventHandler(player.EVENT.ON_PLAYING, playbackStateHandler);
+    }
     // when playback finishes, player turns to paused mode
     player.addEventHandler(player.EVENT.ON_PLAYBACK_FINISHED, playbackStateHandler);
     player.addEventHandler(player.EVENT.ON_CAST_STARTED, playbackStateHandler);
--- a/src/ts/components/playbacktogglebutton.ts
+++ b/src/ts/components/playbacktogglebutton.ts
@@ -43,6 +43,7 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
     // Call handler upon these events
     player.addEventHandler(player.EVENT.ON_PLAY, playbackStateHandler);
     player.addEventHandler(player.EVENT.ON_PAUSED, playbackStateHandler);
+    player.addEventHandler(player.EVENT.ON_PLAYING, playbackStateHandler);
     // when playback finishes, player turns to paused mode
     player.addEventHandler(player.EVENT.ON_PLAYBACK_FINISHED, playbackStateHandler);
     player.addEventHandler(player.EVENT.ON_CAST_STARTED, playbackStateHandler);


protyposis:
Adding a listener for `ON_PLAYING` probably won't hurt, but
 * if `ON_PLAY` is fired before `ON_READY` that's an issue in the player that also needs to be fixed there
 * the `PlaybackToggleButton` checks `player.isPlaying()` at startup specifically to cover the case where the UI is created late and has missed the `ON_PLAY` event. Does `player.isPlaying()` return `false` between `ON_PLAY` and `ON_PLAYING`?

protyposis:
Answering my own question, I now remember that we discussed in the player meeting a few weeks ago that `isPlaying()` must return `true` between `ON_PLAY` and `ON_PLAYING` as to not break the v7 API and keep consistency with how it was before `ON_PLAYING` was introduced. And as far as I remember we specifically implemented it that way.

So the fact that this workaround is required in the UI hints at something being wrong in the player which definitely needs to be fixed there, as I cannot image a case where listening for `ON_PLAY` and checking `isPlaying()` would still miss out on the playing state.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## develop
+- `PlaybackToggleButton` now also listens to ON_PLAYING in addition of ON_PLAY
```
protyposis: * Please adhere to the CHANGELOG format and put the entry under the "Changed" subsection
 * Put event names into `code` tags
 * Grammar: "in addition to"

bitmovin-kenny:
The Problem here seems to be that the UI component is loaded during a microstall during player startup with autoplay. When i put a breakpoint into the `PlaybackToggleButton.configure` function i see that the following events have already been fired by the player:

```log
Event Type:"onReady", isPlaying: true
Event Type:"onPlay", isPlaying: true
Event Type:"onAudioAdaptation", isPlaying: true
Event Type:"onAudioDownloadQualityChange", isPlaying: true
Event Type:"onVideoAdaptation", isPlaying: true
Event Type:"onVideoDownloadQualityChange", isPlaying: true
Event Type:"onMuted", isPlaying: true
Event Type:"onDownloadFinished", isPlaying: true
Event Type:"onSourceLoaded", isPlaying: true
Event Type:"onDownloadFinished", isPlaying: true
Event Type:"onSegmentRequestFinished", isPlaying: true
Event Type:"onVideoDownloadQualityChanged", isPlaying: true
Event Type:"onDownloadFinished", isPlaying: true
Event Type:"onSegmentRequestFinished", isPlaying: true
Event Type:"onAudioDownloadQualityChanged", isPlaying: true
Event Type:"onStallStarted", isPlaying: false
Event Type:"onVideoAdaptation", isPlaying: false
Event Type:"onAudioAdaptation", isPlaying: false
```
Here the setup of the UI is happening... after it finished the following events occur until the `onPlaying` event:

```log
Event Type:"onVolumeChanged", isPlaying: false
Event Type:"onUnmuted", isPlaying: false
Event Type:"onVolumeChanged", isPlaying: false
Event Type:"onMuted", isPlaying: false
Event Type:"onDownloadFinished", isPlaying: false
Event Type:"onSegmentRequestFinished", isPlaying: false
Event Type:"onCastAvailable", isPlaying: false
Event Type:"onDownloadFinished", isPlaying: false
Event Type:"onSegmentRequestFinished", isPlaying: true
Event Type:"onPlayerResize", isPlaying: true
Event Type:"onVideoAdaptation", isPlaying: true
Event Type:"onAudioAdaptation", isPlaying: true
Event Type:"onStallEnded", isPlaying: true
Event Type:"onDownloadFinished", isPlaying: true
Event Type:"onSegmentRequestFinished", isPlaying: true
Event Type:"onDownloadFinished", isPlaying: true
Event Type:"onSegmentRequestFinished", isPlaying: true
Event Type:"onVideoAdaptation", isPlaying: true
Event Type:"onAudioAdaptation", isPlaying: true
Event Type:"onTimeChanged", isPlaying: true
Event Type:"onVideoPlaybackQualityChanged", isPlaying: true
Event Type:"onSegmentPlayback", isPlaying: true
Event Type:"onAudioPlaybackQualityChanged", isPlaying: true
Event Type:"onSegmentPlayback", isPlaying: true
Event Type:"onPlaying", isPlaying: true
```
so i would suggest that we either need the `onPlaying` event handler or an additional `onStallEnded`

protyposis:
Interesting. Shouldn't the player also return `true` for `isPlaying()` during stalls? I mean it's basically in the playing state and would play if it could / didn't have to buffer.

bitmovin-kenny:
Hmmm, valid point, i think i broke that in my feature branch, while letting the player react to pause on the video element, i needed to know if we were stalling and introduced a new Status.
I did not think the player should return true for `isPlaying` when it's stalled, i will change that.

protyposis:
@bitmovin-kenny I don't know if you broke it or if it has always been like this (didn't test it), just wanted to discuss it.

bitmovin-kenny:
@protyposis as far as i can tell (from the code and from testing dev) the player returned `isPlaying` === true during stalls:

```log
Event Type:"onPlay", isPlaying: true
Event Type:"onVideoDownloadQualityChange", isPlaying: true
Event Type:"onVideoAdaptation", isPlaying: true
Event Type:"onAudioDownloadQualityChange", isPlaying: true
Event Type:"onAudioAdaptation", isPlaying: true
Event Type:"onReady", isPlaying: true
Event Type:"onDownloadFinished", isPlaying: true
Event Type:"onSegmentRequestFinished", isPlaying: true
Event Type:"onVideoDownloadQualityChanged", isPlaying: true
Event Type:"onDownloadFinished", isPlaying: true
Event Type:"onSegmentRequestFinished", isPlaying: true
Event Type:"onAudioDownloadQualityChanged", isPlaying: true
Event Type:"onStallStarted", isPlaying: true
Event Type:"onVideoAdaptation", isPlaying: true
Event Type:"onAudioAdaptation", isPlaying: true
Event Type:"onDownloadFinished", isPlaying: true
Event Type:"onSegmentRequestFinished", isPlaying: true
```

as i fixed this bug in my feature branch this PR is no longer necessary (although i think it wouldn't harm to add the additional listener)

PR #76, opened 2017-12-20
Update playback position during paused live streams with DVR window

protyposis:
When pausing a live stream with a DVR window, the playback position stays still and thus moves away from the live edge. The visual playback position indicator on the seekbar should therefore move towards the back.

Because this case was not handled, the playback position during a paused live stream with timeshift window only updated "accidentally" for a few seconds while the buffer was filled, but the buffer is usually < 1 minute while the timeshift window can be much longer (e.g. multiple hours).

This PR adds a timer that is activated once a live stream with timeshift window is paused, and deactivated once playback continues. It updates the playback position once per second.

### DIFF
--- a/src/ts/components/seekbar.ts
+++ b/src/ts/components/seekbar.ts
@@ -198,6 +198,7 @@ export class SeekBar extends Component<SeekBarConfig> {
     // update playback position of Cast playback
     player.addEventHandler(player.EVENT.ON_CAST_TIME_UPDATED, playbackPositionHandler);
 
+    this.configureLivePausedTimeshiftUpdater(player, uimanager, playbackPositionHandler);
 
     // Seek handling
     player.addEventHandler(player.EVENT.ON_SEEK, () => {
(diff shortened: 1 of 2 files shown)


src/ts/components/seekbar.ts
```diff
@@ -317,6 +318,26 @@ export class SeekBar extends Component<SeekBarConfig> {
...
+
+    // Start updater when a live stream with timeshift window is paused
+    player.addEventHandler(player.EVENT.ON_PAUSED, () => {
+      if (player.isLive() && player.getMaxTimeShift() < 0) {
+        pausedTimeshiftUpdater.start();
+      }
+    });
+
+    // Stop updater when playback continues (no matter if the updater was started before)
+    player.addEventHandler(player.EVENT.ON_PLAY, () => pausedTimeshiftUpdater.clear());
```
bitmovin-kenny: i think it won't hurt to add an additional `ON_PLAYING` handler

protyposis: It wouldn't hurt but I say it's unnecessary because
 * it's the player's job to fire the correct events, not the UI's job to fix player bugs
 * the UI is backwards compatible with all 7.x players and `ON_PLAYING` has only been added in 7.3

protyposis:
> Additionally the UI still marks the stream as being on the live edge

This is unrelated and will be fixed in a separate PR.

src/ts/components/seekbar.ts
protyposis: I think there is no way to exit the paused state without an `ON_PLAY` event.

PR #81, opened 2018-01-17
Rework volume changeability detection

bitmovin-kenny:
The volumeslider would try setting the volume on the player to see if it actually changes to detect if we are on a device which does not support setting of volume.
When the player was in muted autoplay and unmuted autoplay was disallowed by the browser this lead to a pause on the video element. The play call afterwards sometimes did not do anything as the player might still have been in playing state (it seems the video element does not fire the pause event immediately).
To avoid this strange behaviour i changed the code to probe a dummy video element with a minimal audio source.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
-### Fixed
-- Positioning of `SeekBar` markers was broken due to style changes in 2.11.0
-
 ### Changed
 - Execute volume control availability test on dummy media element to prevent unexpected interference with muted autoplay
+
+### Fixed
+- Positioning of `SeekBar` markers was broken due to style changes in 2.11.0
 
 ## [2.11.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Positioning of `SeekBar` markers was broken due to style changes in 2.11.0
 
 ### Changed
-- The volume slider now does not probe the volume setting capabilities on the player but on a dummy element to avoid muted autoplay getting paused
+- Execute volume control availability test on dummy media element to prevent unexpected interference with muted autoplay
 
 ## [2.11.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Positioning of `SeekBar` markers was broken due to style changes in 2.11.0
 
+### Changed
+- The volume slider now does not probe the volume setting capabilities on the player but on a dummy element to avoid muted autoplay getting paused
+
 ## [2.11.0]
 
 ### Added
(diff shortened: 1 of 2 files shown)


src/ts/components/volumeslider.ts
```diff
@@ -83,44 +93,20 @@ export class VolumeSlider extends SeekBar {
...
-        return true;
-      }
-    } else {
-      // Volume is not 100, so we're definitely not on a volume-control-restricted iOS device
-      return true;
-    }
+    // as muted autoplay gets paused as soon as we unmute it, we may not touch the volume of the actual player so we
+    // probe a dummy video element with a minimal audio source
+    const dummyVideoElement = document.createElement('video');
+    const dummyAudioSource = document.createElement('source');
```
protyposis: Setting the volume will also work without an audio source so you can remove it

src/ts/components/volumeslider.ts
```diff
@@ -20,6 +20,16 @@ export class VolumeSlider extends SeekBar {
...
+  /**
+   * Small MP3 from https://mrcoles.com/detecting-html5-audio-autoplay/
+   * @type {string}
+   */
+  public static readonly dummyAudioSource: string = 'data:audio/mpeg;base64,/+MYxAAAAANIAUAAAASEEB/jwOFM/0MM/90b/+R' +
+    'hST//w4NFwOjf///PZu////9lns5GFDv//l9GlUIEEIAAAgIg8Ir/JGq3/+MYxDsLIj5QMYcoAP0dv9HIjUcH//yYSg+CIbkGP//8w0bLVjUP///' +
+    '3Z0x5QCAv/yLjwtGKTEFNRTMuOTeqqqqqqqqqqqqq/+MYxEkNmdJkUYc4AKqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq' +
+    'qqqqqqqqqqqqqqqqqqqqqqqqq';
+
+
```
protyposis: Duplicate empty line

src/ts/components/volumeslider.ts
bitmovin-kenny: You were right, tested it today and removed the dummyAudioSource

protyposis: But the source element itself is still needed?

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Positioning of `SeekBar` markers was broken due to style changes in 2.11.0
 
+### Changed
+- The volume slider now does not probe the volume setting capabilities on the player but on a dummy element to avoid muted autoplay getting paused
```
protyposis: `Execute volume control availability test on dummy media element to prevent unexpected interference with muted autoplay`

src/ts/components/volumeslider.ts
bitmovin-kenny: Changed it to probe an `audio` element, i though it would be more safe to probe a video element with an audio source but it seems to be an overhead

protyposis: Why did you change it to an `audio` element?

bitmovin-kenny: Because we want to check if we can change the audio volume?

PR #82, opened 2018-01-17
Add smooth source config

bitmovin-kenny:
As we support smooth streaming in the player, i added the `smooth` part of the source config

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
+### Added
+- Added Smooth Streaming to the source config 
+
 ### Fixed
 - Positioning of `SeekBar` markers was broken due to style changes in 2.11.0
 
--- a/src/ts/player-config.d.ts
+++ b/src/ts/player-config.d.ts
@@ -391,6 +391,11 @@ declare namespace bitmovin {
        */
       hls?: string;
       /**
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
+### Added
```
protyposis: We don't add type definition improvements to the changelog because it's not really related to the UI

src/ts/player-config.d.ts
```diff
@@ -390,6 +390,11 @@ declare namespace bitmovin {
        * HLS content can easily and for free be generated using our encoding solution bitcodin.
        */
       hls?: string;
+      /**
+       * An URL to a Microsoft Smooth Streaming Manifest (normally ends with .ism/Manifest but can also be a .xml)
+       * Supported by player version 7.5+
```
protyposis: `@since v7.5`

src/ts/player-config.d.ts
```diff
@@ -390,6 +390,11 @@ declare namespace bitmovin {
        * HLS content can easily and for free be generated using our encoding solution bitcodin.
        */
       hls?: string;
+      /**
+       * An URL to a Microsoft Smooth Streaming Manifest (normally ends with .ism/Manifest but can also be a .xml)
+       * Supported by player version 7.5
```
protyposis: Please use the `@since` JSDoc/TypeDoc tag instead of a plain text sentence: `@since v7.5`

Issue #65, opened 2017-10-19
Add localization for strings


protyposis:
Text strings in the UI are currently all hardcoded in English. Most strings can be changed by creating a custom UI structure and setting the labels, but not all of them.

A decent localization interface would not require users at all to create a custom UI structure. It should work with the player's default UI by supplying localized strings to the player config.

protyposis:
Here's a simple script to localize (static) labels of UI from the outside, without such a feature actually being implemented in the UI:

```js
// German translations
var translationMap = {
  // Settings menu
  'Video Quality': 'Videoqualität',
  'Audio Quality': 'Audioqualität',
  'Audio Track': 'Audiospur',
  'Speed': 'Geschwindigkeit',
  // Subtitle settings menu
  'Subtitles': 'Untertitel',
  'Font size': 'Größe',
  'Font family': 'Schriftart',
  'Font color': 'Farbe',
  'Font opacity': 'Deckkraft',
  'Character edge': 'Ränder',
  'Background color': 'Hintergrundfarbe (Text)',
  'Background opacity': 'Hintergrunddeckkraft (Text)',
  'Window color': 'Hintergrundfarbe (Textbox)',
  'Window opacity': 'Hintergrunddeckkraft (Textbox)',
  'Back': 'Zurück',
  'Reset': 'Zurücksetzen',
};

function translateElementContent(element) {
  var currentText = element.innerHTML;
  var translatedText = translationMap[currentText];

  // If a translation exists, apply it
  if (translatedText) {
    element.innerHTML = translatedText;
  }
}

function translateUi(uicontainer) {
  // Gets the labels that are "Label" components
  var labelComponents = Array.from(uicontainer.getElementsByClassName('bmpui-ui-label'));
  // Gets labels that aren't distinct components but part of other components' DOM tree
  var innerLabels = Array.from(uicontainer.getElementsByClassName('bmpui-label'));

  // Concatenate all elements into a single array so we can iterate over all of them easily
  var elements = labelComponents.concat(innerLabels);

  // Translate the content of all elements
  // A label is usually a <span> with text inside
  elements.forEach(function(element) {
    translateElementContent(element);

    // Workaround for the "Subtitles" label
    // This label unfortunately does not follow the label-pattern and carries
    // its text inside a child <span>, so we check that too
    var innerSpans = element.getElementsByTagName('span');
    if (innerSpans.length > 0) {
      translateElementContent(innerSpans[0]);
    }
  });
}

[...]

PR #83, opened 2018-01-24
Experimental suggestion on i18n support

molszanski:
In reference to an issue: "**Add localization for strings**" https://github.com/bitmovin/bitmovin-player-ui/issues/65.

Within the current player-ui architecture I see 3 possible implementation. With their own pros and cons. 
- Replace text approach
- String label access based approach, i18next alike
- Object / key based access

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Beside the Git repository, the UI framework is also available through the follow
 
 The UI framework and default skin bundled with the latest player release are always available via CDN. This is the recommended way if you just want to work with the predefined UI components. All components will be available in the `bitmovin.playerui` namespace.
 
- * JavaScript library: `//bitmovin-a.akamaihd.net/bitmovin-player/stable/7/bitmovinplayer-ui.js` 
+ * JavaScript library: `//bitmovin-a.akamaihd.net/bitmovin-player/stable/7/bitmovinplayer-ui.js`
  * CSS default skin: `//bitmovin-a.akamaihd.net/bitmovin-player/stable/7/bitmovinplayer-ui.css`
 
 ### NPM
@@ -33,7 +33,7 @@ The UI framework is also available in the NPM repository and comes with all sour
   * `gulp serve` to open test page in browser, build and reload changed files automatically
   * `gulp lint` to lint TypeScript and SASS files
   * `gulp build-prod` to build project with minified files into `dist` directory
-  
+
 To just take a look at the project, also run `gulp serve`. For changes, check our [CHANGELOG](CHANGELOG.md).
 
 ## Contributing
@@ -42,7 +42,7 @@ Pull requests are welcome! Please check the [contribution guidelines](CONTRIBUTI
 
 ## Introduction
 
-This repository contains the Bitmovin Player UI framework introduced with the 7.0 release. 
+This repository contains the Bitmovin Player UI framework introduced with the 7.0 release.
 It is designed as a flexible and modularized layer on the player API that replaces the old integrated monolithic UI, enabling customers and users of the player to easily customize the UI to their needs in design, structure, and functionality. It makes it extremely easy and straightforward to add additional control components and we encourage our users to proactively contribute to our codebase.
 
 The framework basically consists of a `UIManager` that handles initialization and destruction of the UI, and components extending the `Component` base class. Components provide specific functionality (e.g. `PlaybackToggleButton`, `ControlBar`, `SeekBar`, `SubtitleOverlay`) and usually consist of two files, a TypeScript `.ts` file containing control code and API interaction with the player, and a SASS `.scss` file containing the visual style.
@@ -150,12 +150,12 @@ Here is an example on how to display a special UI in fullscreen mode:
 ```js
 bitmovin.player('player-id').setup(config).then(function (player) {
   var myUiManager = new bitmovin.playerui.UIManager(player, myWindowUi);
-  
+
   player.addEventHandler(player.EVENT.ON_FULLSCREEN_ENTER, function () {
     myUiManager.release();
     myUiManager = new bitmovin.playerui.UIManager(player, myFullscreenUi);
   });
-  
+
   player.addEventHandler(player.EVENT.ON_FULLSCREEN_EXIT, function () {
     myUiManager.release();
     myUiManager = new bitmovin.playerui.UIManager(player, myWindowUi);
@@ -240,6 +240,44 @@ var myUiManager = new bitmovin.playerui.UIManager(player, myUi, myUiConfig);
 ```
 
 All of the configuration properties are optional. If `metadata` is set, it overwrites the metadata of the player configuration. If `recommendations` is set, a list of recommendations is shown in the `RecommendationOverlay` at the end of playback. For this to work, the UI must contain a `RecommendationOverlay`, like the default player UI does.
+
+### UI Translations
+
+You can use either  a built-in locale or provide your own translation.
+
+**Buil-it locales**
+
+Right now `en` and `de` locales are supported. For example
+```js
+var config = {...};
+var uiConfig = { locale: 'de' };
+
+bitmovin.player('player-id').setup(config).then(function (player) {
+  var myUiManager = bitmovin.playerui.UIManager.Factory.buildDefaultUI(player, uiConfig);
+});
+```
+
+**Custom translation**
+
+Alternatively, you can provide your own translation with mappings for the default UI language. Take a look at `src/ts/translations/en.ts` file. Any missing translations will use english as the default.
+
+```js
+var config = {...};
+var uiConfig = {
+  ...,
+  translations: {
+    settings: {
+      //videoQuality: 'Videoqualität',
+      audioQuality: 'Audioqualität'
+    }
+  }
+};
+
+bitmovin.player('player-id').setup(config).then(function (player) {
+  var myUiManager = bitmovin.playerui.UIManager.Factory.buildDefaultUI(player, uiConfig);
+});
+// Since videoQuality key is not provided, it will use english translation 'Video Quality' as a default
+```
 
 ### UI Playground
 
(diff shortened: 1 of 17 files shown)


src/ts/i18n.ts
```diff
@@ -0,0 +1,106 @@
...
+const predifinedLanguages: any = {
+  en: en,
+  de: de,
+};
+
+/**
+ * @prop q - quick translation
+ * @prop en - default english map of strings
+ */
+export class BtTranslate {
```
protyposis: Why don't you just call it `I18n`?

src/ts/translations/de.ts
```diff
@@ -0,0 +1,41 @@
+import {Translation} from './translation-types';
+
+let lang: Translation = {
```
protyposis: I like the general idea of the structure of this `Translation` definition, I'd flatten the hierarchy a bit though

protyposis: Also you can directly do a `export default <Translation>{` here

src/ts/translations/de.ts
```diff
@@ -0,0 +1,41 @@
...
+  language: 'german',
+  settings: {
+    panel: {
```
protyposis: Remove the `panel` and move the panel strings directly into the `settings`

src/ts/translations/de.ts
```diff
@@ -0,0 +1,41 @@
...
+      fontColor: 'Farbe',
+      fontOpacity: 'Deckkraft',
+      characterEdge: 'Ränder',
+      backgroundColor: 'Hintergrundfarbe (Text)',
+      backgroundOpacity: 'Hintergrunddeckkraft (Text)',
+      windowColor: 'Hintergrundfarbe (Textbox)',
+      windowOpacity: 'Hintergrunddeckkraft (Textbox)',
+    },
+  },
+  labels: {
```
protyposis: All these labels IMO also belong into the `settings`

src/ts/translations/de.ts
```diff
@@ -0,0 +1,41 @@
...
+    subtitles: 'Untertitel',
+    subtitleSettings: 'Subtitles settings',
+    reset: 'Zurücksetzen',
+    back: 'Zurück',
+    normal: 'Normal',
+  },
+  messages: {
```
protyposis: Why is this called `messages`, whats the differentiation to the other non-messages?

src/ts/translations/de.ts
```diff
@@ -0,0 +1,41 @@
...
+    remainingTime: 'Ad: {remainingTime} secs',
```
protyposis: I'd put that into an `ads` section

src/ts/translations/de.ts
```diff
@@ -0,0 +1,41 @@
...
+    connectingTo: function(castDeviceName: string): string {
```
protyposis: This and the next should go into a `cast` section

protyposis: I like this function-based pattern but I think the downside is that this prohibits us from supplying `Translation` objects as JSON. If they were pure JSON objects then they could much easier be used with translation software. The object property hierarchy could easily be converted into i18next-type strings and back if necessary, e.g. `en.settings.audioQuality` <=> `'en.settings.audioQuality'`.

So in the light of that, it might make more sense to keep a `Translation` object function free and provide a separate function on the `ì18n` object that must be used where placeholders are desired.

src/ts/uimanager.ts
```diff
@@ -50,6 +50,10 @@ import {Spacer} from './components/spacer';
 import {UIUtils} from './uiutils';
 import {ArrayUtils} from './arrayutils';
 import {BrowserUtils} from './browserutils';
+import i18n from './i18n';
+
+// Interfaces and types
+import { Panel } from './translations/translation-types';
```
protyposis: All imports should be in a single block at the top of the file... else IntelliJ will do some weird stuff on reformatting

src/ts/uimanager.ts
```diff
@@ -396,30 +400,43 @@ export class UIManager {
     this.managerPlayerWrapper.clearEventHandlers();
   }
 }
+// For the lack of a better place, a wip helper to configure a UI language
+function configureUiLanguage(config: any): void {
+  if (config.locale !== undefined) {
```
protyposis: We usually use the `if (config.locale) { ...` pattern

src/ts/uimanager.ts
```diff
@@ -396,30 +400,43 @@ export class UIManager {
...
+  if (config.translation !== undefined) {
+    i18n.addTranslation(config.translation);
+  }
+}
+
 
 export namespace UIManager.Factory {
 
   export function buildDefaultUI(player: PlayerAPI, config: UIConfig = {}): UIManager {
+    configureUiLanguage(config);
```
protyposis: This should probably go directly into the `UIManager` or `UIInstanceManager`

test/tansl.spec.coffee
```diff
@@ -0,0 +1,154 @@
...
+assert = chai.assert
+path = require 'path'
+
+Translation = require '../src/ts/i18n.ts'
+# process.env.DEBUG = 'bt-player-ui:*'
+
+swallowConsoleWarnings = false
+
+try
+	describe 'Translation tests: ', ->
```
protyposis: Please write the tests in TypeScript as well... I can see how the CS syntax is much cleaner but all our projects are currently pure TypeScript and introducing another language would slow us down and increase the complexity

protyposis: Btw. great that you wrote tests 👍

src/ts/i18n.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  constructor() {
+    return this;
+  }
+
+  public g(): Translation {
+    return this.q;
+  }
+
+  public addTranslation(newTranslation: any): void {
```
protyposis: Does `Partial<Translation>` instead of `any` work, or does the partial only apply to the top level of the object?

src/ts/i18n.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  Future work:
+    - It is possible to make use of getters / setters:
+      - will lead to better IDE support
+      - would be possible to utilize typescript built-in
+        interface and enum support. This will also reduce
+        a bundle size
+*/
+
+const predifinedLanguages: any = {
```
protyposis: Typo: `predefinedLanguages`

protyposis: Please don't use `any` types... better to omit it and let TS infer the type, e.g. here it will probably be `{ [name: string]: Translation }`... but ideally the type is correctly defined

src/ts/components/subtitlesettings/subtitlesettingspanel.ts
```diff
@@ -33,34 +37,36 @@ export class SubtitleSettingsPanel extends SettingsPanel {
     this.overlay = config.overlay;
 
     let manager = new SubtitleSettingsManager();
+    // Quick lang getters
+    const i18fonts: FontPanelLabels = i18n.q.settings.fonts;
 
     this.config = this.mergeConfig(config, {
       components: <Component<ComponentConfig>[]>[
-        new SettingsPanelItem('Font size', new FontSizeSelectBox({
+        new SettingsPanelItem(i18fonts.fontSize, new FontSizeSelectBox({
```
protyposis: I would prefer if `FontPanelLabels` wasn't a separate interface but defined inline in the `Translation` interface so it purposefully cannot be used like this here and the strings would always have to be specified with their "fully qualified name", e.g. `i18n.q.settings.fonts.fontSize` instead of `i18fonts.fontSize`. It's more verbose, true, but better resembles the "usual" pattern of translation frameworks and might be more intuitive to understand.

src/ts/uimanager.ts
protyposis: If no locale is set in the `UIConfig` then we should try to detect the language from the `Accept-Language` header.

src/ts/uimanager.ts
```diff
@@ -396,30 +400,43 @@ export class UIManager {
     this.managerPlayerWrapper.clearEventHandlers();
   }
 }
+// For the lack of a better place, a wip helper to configure a UI language
+function configureUiLanguage(config: any): void {
```
protyposis: `UIConfig` instead of `any`

protyposis:
A few i18n libraries for inspiration:
https://github.com/fnando/i18n-js
https://github.com/airbnb/polyglot.js
https://github.com/i18next/i18next

protyposis:
Recap of today's call:

 * There's various pattern on how to store and retrieve strings and how to process placeholders, and all of them can be easily implemented and will work, so this is pretty straightforward
 * The i18n mechanism must not be a singleton, i.e. the language should be configurable per UI instance to allow multiple UIs with different languages, similar to how multiple UI instances can generally be configured differently
 * The actual issue is how to integrate the translations into the current architecture, especially in cases where the text is set in the configuration object, which can either be inside a `Component` constructor or outside in a factory method or whereever a UI structure is defined
 * It might be necessary to extend the architecture to allow component rerendering when some state, e.g. the config object, changes
 * Output size is currently not an issue but the increase should of course be kept minimal, i.e. large 3rd party libraries should be avoided
 * Intuitiveness / code readability goes over microoptimizations

README.md
```diff
@@ -241,6 +241,44 @@ var myUiManager = new bitmovin.playerui.UIManager(player, myUi, myUiConfig);
 
 All of the configuration properties are optional. If `metadata` is set, it overwrites the metadata of the player configuration. If `recommendations` is set, a list of recommendations is shown in the `RecommendationOverlay` at the end of playback. For this to work, the UI must contain a `RecommendationOverlay`, like the default player UI does.
 
+### UI Translations
+
+You can use either  a built-in locale or provide your own translation.
+
+**Buil-it locales**
```
aramin: spelling `Built-In`

PR #92, opened 2018-04-06
Use ON_PLAYBACK_SPEED_CHANGED event

dweinber:
There was no event in the player to notify about playback speed changes. This event will be introduced in v7.7. The UI now listens to the event (if the player version supports it) and changes the selector accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [development]
 
 ### Added
-- Subscribe to the `ON_PLAYBACK_SPEED_CHANGED` event to display the correct speed in the selector
+- Subscribe to the `ON_PLAYBACK_SPEED_CHANGED` event to display the correct speed in the `PlaybackSpeedSelectBox`
 
 ## [2.13.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [development]
+
+### Added
+- Subscribe to the `ON_PLAYBACK_SPEED_CHANGED` event to display the correct speed in the selector
+
 ## [2.13.0]
 
 ### Changed
--- a/src/ts/components/playbackspeedselectbox.ts
+++ b/src/ts/components/playbackspeedselectbox.ts
@@ -32,7 +32,7 @@ export class PlaybackSpeedSelectBox extends SelectBox {
     player.addEventHandler(player.EVENT.ON_READY, setDefaultValue);
 
     if (player.EVENT.ON_PLAYBACK_SPEED_CHANGED) {
-      // Since player 7.7.0
+      // Since player 7.8.0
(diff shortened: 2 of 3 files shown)


src/ts/components/playbackspeedselectbox.ts
```diff
@@ -38,5 +33,33 @@ export class PlaybackSpeedSelectBox extends SelectBox {
...
     player.addEventHandler(player.EVENT.ON_READY, setDefaultValue);
+
+    if (player.EVENT.ON_PLAYBACK_SPEED_CHANGED) {
+      // Since player 7.7.0
+      player.addEventHandler(player.EVENT.ON_PLAYBACK_SPEED_CHANGED, () => {
+        const speed = String(player.getPlaybackSpeed());
+
+        if (!this.selectItem(speed)) {
+          // a playback speed was set which is not in the list, add it to the list to show it to the user
+          this.clearItems();
```
bitmovin-kenny: Why do we clear everything here and then add the default items + the new one? Wouldn't it be enough to just add the new Item.
We should also add the playback speed to the list when we set the initial value in line 26:
```ts
       let playbackSpeed = String(player.getPlaybackSpeed());
       if (!this.selectItem(playbackSpeed)) {
         playbackSpeed = '1';
```
Furthermore i think the we should do a sorted insert to improve the usability

dweinber: If multiple "non-standard" values are selected, each time the item would be added and the list could grow. Therefore I decided to clear the old ones, but open for discussion.

Good catch with the initial value in L26, will update.

Sorted insert would have been nice... 😉

dweinber: Updated to have sorted insert as well.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [development]
+
+### Added
+- Subscribe to the `ON_PLAYBACK_SPEED_CHANGED` event to display the correct speed in the selector
```
protyposis: `PlaybackSpeedSelectBox` (in code) instead of "selector"

src/ts/components/playbackspeedselectbox.ts
```diff
@@ -6,37 +6,60 @@ import {UIInstanceManager} from '../uimanager';
...
     player.addEventHandler(player.EVENT.ON_READY, setDefaultValue);
+
+    if (player.EVENT.ON_PLAYBACK_SPEED_CHANGED) {
+      // Since player 7.7.0
+      player.addEventHandler(player.EVENT.ON_PLAYBACK_SPEED_CHANGED, setDefaultValue);
+    }
+  }
+
+  setSpeed(speed: string): void {
+    if (!this.selectItem(speed)) {
```
protyposis: I would handle the speeds as `number` everywhere except where it needs to be a string, i.e. convert it to `String` in `addItem` and `selectItem` and leave it `number` everywhere else. Because now it is `ON_PLAYBACK_SPEED_CHANGED` > `setDefaultValue` > `number` to `string` (`player.getPlaybackSpeed()`) > `setSpeed` > `string` to `number` (`addDefaultItems`) > `addDefaultItems` > `number` to `string` (`addItem`).

PR #93, opened 2018-04-17
Use on/off instead of addEventHandler/removeEventHandler when available

bitmovin-kenny:
Since Player Version 7.7 add/removeEventHandler are deprecated. This PR introduces the adaptation to the new on + off methods while keeping a fallback to the old functionality to stay compatible with older versions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [development]
 
 ### Added
-- Support for `on`/`off` replacing `addEventHandler`/`removeEventHandler` in player Version 8
+- Prefer `on`/`off` over `addEventHandler`/`removeEventHandler` with player version 7.8+ to avoid deprecation log messages
 
 ## [2.13.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [development]
+
+### Added
+- Support for `on`/`off` replacing `addEventHandler`/`removeEventHandler` in player Version 8
+
 ## [2.13.0]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


src/ts/player.d.ts
```diff
@@ -57,8 +57,19 @@ declare namespace bitmovin {
      *
      * @param eventType The type of event to subscribe to.
      * @param callback The event callback handler that will be called when the event fires.
+     *
+     * @deprecated since 7.7, use {@link PlayerAPI#on} instead
```
protyposis: The `#` notation is Javadoc syntax and does not work in TypeDoc (even though it works in the IntelliJ editor), you need to use a `.` for static and instance accessors: `{@link PlayerAPI.on}`

src/ts/player.d.ts
```diff
@@ -326,9 +337,20 @@ declare namespace bitmovin {
      *
      * @param eventType The event to remove the handler from
      * @param callback The callback handler to remove
+     *
+     * @deprecated since 7.7, use {@link PlayerAPI#off} instead
```
protyposis: As above

src/ts/uimanager.ts
```diff
@@ -1118,7 +1118,13 @@ class PlayerWrapper {
 
     // Explicitly add a wrapper method for 'addEventHandler' that adds added event handlers to the event list
     wrapper.addEventHandler = (eventType: EVENT, callback: PlayerEventCallback) => {
-      player.addEventHandler(eventType, callback);
+      // in player V8 addEventHandler was replaced by on
+      if (player.on) {
+        (player as any).on(eventType, callback);
```
protyposis: Since you added the definitions, the `any` cast should not be necessary

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [development]
+
+### Added
+- Support for `on`/`off` replacing `addEventHandler`/`removeEventHandler` in player Version 8
```
protyposis: It is player 7.7 (or 7.8) I guess which introduces on/off? There will be a separate branch and UI 3.0 for player v8 so it is confusing to mention it here

protyposis: Also `version` in lowercase, its not a proper noun.

protyposis:
This got accidentally merged into `master` instead of `develop`. Fixed by #96.

PR #98, opened 2018-04-25
fix blocked autoplay

bitmovin-kenny:
When a play call was tried but blocked due to disabled autoplay on some devices the `play` event would be fired, but the player would never transition into playing. This breaks the hugeplaybacktogglebutton as it only waits for the first `play` event to start the the waiting for a potential double click for potential fullscreen handling. This handling destroys the user interaction needed for the play call to succeed.

As a fix we now reset this first play flag on an `onWarning` event to use the first click as a user interaction to call play.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@ License change from LGPLv3 to MIT.
 - Prefer `on`/`off` over `addEventHandler`/`removeEventHandler` with player version 7.8+ to avoid deprecation log messages
 - `data-bmpui-volume-level-tens` attribute on `VolumeToggleButton` for more granular styling of the volume icon
 - `onClass`/`offClass` configuration properties in `ToggleButtonConfig` to allow customizing the state marker CSS class names
-- handling for blocked autoplay in the `hugeplaybacktogglebutton`
 
 ### Changed
 - Removed `bmpui-low` marker class from `VolumeToggleButton` (replaced by `data-bmpui-volume-level-tens` attribute)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ License change from LGPLv3 to MIT.
 - Prefer `on`/`off` over `addEventHandler`/`removeEventHandler` with player version 7.8+ to avoid deprecation log messages
 - `data-bmpui-volume-level-tens` attribute on `VolumeToggleButton` for more granular styling of the volume icon
 - `onClass`/`offClass` configuration properties in `ToggleButtonConfig` to allow customizing the state marker CSS class names
+- handling for blocked autoplay in the `hugeplaybacktogglebutton`
```
protyposis: Please move to `Fixed` section and call it: User interaction passthrough from `HugePlaybackToggleButton` to player when autoplay is blocked

src/ts/components/hugeplaybacktogglebutton.ts
```diff
@@ -102,6 +103,14 @@ export class HugePlaybackToggleButton extends PlaybackToggleButton {
       firstPlay = false;
     });
 
+    player.addEventHandler(player.EVENT.ON_WARNING, (ev: WarningEvent) => {
```
protyposis: Please call the parameter `event`... there is no need to shorten the parameters and make them less readable because UglifyJS takes care of that in the production build

PR #99, opened 2018-04-25
Fix seekbar in Android 4.4 WebView

protyposis:
Missing `-webkit` prefixes on `transform` properties led to a dysfunctional `SeekBar` in Android 4.4's `WebView` (which is based on an old `WebKit` engine).


protyposis:
There was a small but serious bug that I fixed in fae9508

Issue #101, opened 2018-05-22
Add API to switch UI variants


protyposis:
Currently it is possible to define different UI variants that are switched according to criteria provided by the `UIConditionContext` object when certain player events happen. This means it is only possible to set the criteria for certain UI variants at setup, but switching is done autonomously and internally in the `UIManager` without any means to interfere from outside. Desired is an API that also allows to trigger switching from outside at any desired time.

The `UIConditionContext` upon which the target UI variant is evaluated:
https://github.com/bitmovin/bitmovin-player-ui/blob/d8a6978f593ad229e96f051257b7d518d9687619/src/ts/uimanager.ts#L80-L118

Example UI variants and criteria setup:
https://github.com/bitmovin/bitmovin-player-ui/blob/d8a6978f593ad229e96f051257b7d518d9687619/src/ts/uimanager.ts#L654-L672

The events that trigger the evaluation of the context and potentially switch the UI variant:
https://github.com/bitmovin/bitmovin-player-ui/blob/d8a6978f593ad229e96f051257b7d518d9687619/src/ts/uimanager.ts#L339-L349

protyposis:
Since the UI variants do not have distinct IDs, it is not possible to directly switch to a specific variant, e.g. `uimanager.switchToVariant(UIVariant.Ads)`. The API will thus somehow have to work with the `UIConditionContext` and allow setting the conditions so the switching code evaluates to the desired UI variant.

PR #102, opened 2018-05-22
Add API to switch UI variants

protyposis:
Implements #101 and provides multiple ways to switch between UI variants via the public `UIManager` API:

```ts
interface UIConfig {
  /** 
   * A configuration flag that allows to turn off automatic UI variant switching to allow purely 
   * externally triggered switching through `resolveUiVariant` and/or directly through 
   * `switchToUiVariant` (default `true`)
   */
  autoUiVariantResolve?: boolean;
}

class UIManager {
  /**
   * An event that is triggered just before the next UI variant is resoved and the UI variant 
   * is switched. It is fired when the switch is triggered from an automatic switch and when 
   * calling `resolveUiVariant` and can be used to modify the `UIConditionContext` before 
   * the UI variant is resoved.
   */
  onUiVariantResolve: EventDispatcher<UIManager, UIConditionContext>;

  /*
   * Returns the list of UI variants as passed into the constructor of `UIManager`.
   */
  getUiVariants(): UIVariant[];
  /*
   * Switches to a passed in UI variant taken from the list returned by `getUiVariants()`.
   */
  switchToUiVariant(uiVariant: UIVariant): void;
  /*
   * Triggers a UI variant switch according to the current `UIConditionContext` and allows 
   * to overwrite internal context properties.
   */
  resolveUiVariant(context: Partial<UIConditionContext> = {}): void;
}
```

## Examples

All the following examples can be combined in all kinds of combinations. Im many cases it will make sense to disable automatic switching and handle the switching purely through the API with `switchToUiVariant`, but there might be cases where automatic switching is wanted but certain modifications in the behavior are desired. In the latter case it makes sense to combine `resolveUiVariant` and/or `switchToUiVariant` with `onUiVariantResolve`.

```ts
const uiConfig: UIConfig = {};

const uiManager = new UIManager(player, [{
  // Add a UI variant for ads
  ui: buildAdsUi(),
  condition: (context: UIConditionContext) => {
    return context.isAd;
  },
}, {
  // Add a UI variant for the main content
  ui: buildDefaultUi(),
}], uiConfig);
```

### Example 1: Modify automatic switches

```ts
uiManager.onUiVariantResolve.subscribe((sender, context) => {
  // Avoid switching to the ads UI depending on our custom `preventAdsUi` flag
  if (context.isAd && preventAdsUi === true) {
    context.isAd = false;
  }
});
```

### Example 2: Trigger a switch
```ts
// Switch to the ads UI independently from the content and player events
// This will reset as soon as any monitored player event is fired (e.g. `ON_PAUSED`)
uiManager.resolveUiVariant({ isAd: true });
```

### Example 3: Disable automatic switching and trigger a switch by player events
```ts
// Disable automatic switching
uiConfig = {
  autoUiVariantResolve: false,
};

// Show ads UI when ad starts
player.addEventHandler(player.EVENT.ON_AD_STARTED, () => {
  uiManager.resolveUiVariant({ isAd: true });
});

// Switch back to default UI when ad is finished
player.addEventHandler(player.EVENT.ON_AD_FINISHED, () => {
  uiManager.resolveUiVariant({ isAd: true });
});
```

### Example 4: Switch directly to desired variant

```ts
const myUiVariants = uiManager.getUiVariants();
// We know the index because we have defined the UI variant list and passed it into the UIManager
const myAdsUiVariant = myUiVariants[0];

// Show ads UI when ad starts
player.addEventHandler(player.EVENT.ON_AD_STARTED, () => {
  uiManager.switchToUiVariant(myAdsUiVariant);
});
```

### DIFF
--- a/src/ts/uimanager.ts
+++ b/src/ts/uimanager.ts
@@ -314,36 +314,15 @@ export class UIManager {
     return this.config;
   }
 
-  resolveUiVariant(context: Partial<UIConditionContext> = {}, onShow?: (context: UIConditionContext) => void): void {
-    // Determine the current context for which the UI variant will be resolved
-    const defaultContext: UIConditionContext = {
-      isAd: false,
-      isAdWithUI: false,
-      adClientType: null,
-      isFullscreen: this.player.isFullscreen(),
-      isMobile: BrowserUtils.isMobile,
-      isPlaying: this.player.isPlaying(),
-      width: this.uiContainerElement.width(),
-      documentWidth: document.body.clientWidth,
-    };
+  getUiVariants(): UIVariant[] {
+    return this.uiVariants;
+  }
 
-    // Overwrite properties of the default context with passed in context properties
-    const switchingContext = { ...defaultContext, ...context };
+  switchToUiVariant(uiVariant: UIVariant, onShow?: () => void): void {
+    let uiVariantIndex = this.uiVariants.indexOf(uiVariant);
 
-    // Fire the event and allow modification of the context before it is used to resolve the UI variant
-    this.events.onUiVariantResolve.dispatch(this, switchingContext);
-
-    let nextUi: InternalUIInstanceManager = null;
+    const nextUi: InternalUIInstanceManager = this.uiInstanceManagers[uiVariantIndex];
     let uiVariantChanged = false;
-
-    // Select new UI variant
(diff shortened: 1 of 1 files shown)


protyposis:
API discussed and accepted in the player architecture meeting 2018-05-22.

Issue #107, opened 2018-06-13
Don't animate HugePlaybackToggleButton on initialization


protyposis:
When the `HugePlaybackToggleButton` is initialized and rendered for the first time, the button fades in with an animation if the player is paused, or fades out with an animation if the player is player.

The button should simple be displayed or not displayed without any animation when it is initially rendered.

protyposis:
Released in `2.16.0`

PR #109, opened 2018-06-13
Animate HugePlaybackToggleButton only on state changes

protyposis:
Implements #107

This does not work in all cases because it depends on when the UI is initialized and which events from the player it receives. It works in some cases though, so it's an improvement for some users.

### DIFF
--- a/src/ts/components/hugeplaybacktogglebutton.ts
+++ b/src/ts/components/hugeplaybacktogglebutton.ts
@@ -139,4 +139,19 @@ export class HugePlaybackToggleButton extends PlaybackToggleButton {
 
     return buttonElement;
   }
+
+  /**
+   * Enables or disables the play state transition animations of the play button image. Can be used to suppress
+   * animations.
+   * @param {boolean} enabled true to enable the animations (default), false to disable them
+   */
+  protected setTransitionAnimationsEnabled(enabled: boolean): void {
+    const noTransitionAnimationsClass = this.prefixCss('no-transition-animations');
+
+    if (enabled) {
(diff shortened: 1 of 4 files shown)


src/ts/components/hugeplaybacktogglebutton.ts
```diff
@@ -124,6 +124,14 @@ export class HugePlaybackToggleButton extends PlaybackToggleButton {
     player.addEventHandler(player.EVENT.ON_CAST_START, castInitializationHandler);
     player.addEventHandler(player.EVENT.ON_CAST_STARTED, castInitializationHandler);
     player.addEventHandler(player.EVENT.ON_CAST_STOPPED, castInitializationHandler);
+
+    // Hide the play button animation when the UI is loaded (it should only be animated on state changes)
+    this.setTransitionAnimationsEnabled(false);
+
+    // Enable the transition animations when the play state changes for the first time
+    this.onToggle.subscribeOnce(() => {
+      this.setTransitionAnimationsEnabled(true);
```
bitmovin-kenny: woulnd't it be more simple to add the fade animation class here instead of supressing it before with a different css class?

protyposis: It would save one line of code but wouldn't be as nicely backwards compatible. This solution can be easily merged into custom branches of customers because it only adds an additional CSS rule, your proposal would make it necessary to modify existing rules that would create merge conflicts.

protyposis: I have evaluated both approaches and the result is the same so I'm sticking with the current one.

Issue #110, opened 2018-06-13
Autoplay handling


protyposis:
The UI should render the play button in state playing (display pause icon in play button, hide play overlay in huge playback button) when an autoplay video is starting.

The player works as follows:
 * "Normal" play: `new Player()` -> `player.setup({ source })` -> `player.play()` -> `ON_PLAY` -> `ON_PLAYING` or `ON_PAUSED`+`ON_WARNING 5008` (when playback is blocked, e.g. `play()` call without user interaction)
* Autoplay: `new Player()` -> `player.setup({ source, playback: { autoplay: true }})` -> `ON_PLAYING` or `ON_WARNING 5008` (if autoplay is blocked, e.g. due to audio)

In case of autoplay, there is no `ON_PLAY` event because playback is triggered implicitly, i.e. without an API `play()` call, and also no `ON_PAUSED` call if it fails because `ON_PAUSED` can only come after an `ON_PLAY`. So if autoplay works, we get an `ON_PLAYING` event, and if it fails, we get an `ON_WARNING` event with code `5008`.

When the UI is initialized after the player setup has finished, there is no way to check if the autoplay attempt has already happened and failed, or if it is upcoming, so we cannot rely on the `autoplay` config flag when the player is ready. In this case, we simply rely on the playing state.

protyposis:
Released in `2.16.0`

PR #113, opened 2018-06-14
Hide play button when autoplay is enabled

protyposis:
Implements #110.

### DIFF
--- a/src/ts/components/hugeplaybacktogglebutton.ts
+++ b/src/ts/components/hugeplaybacktogglebutton.ts
@@ -133,30 +133,23 @@ export class HugePlaybackToggleButton extends PlaybackToggleButton {
       this.setTransitionAnimationsEnabled(true);
     });
 
-    // Hide the play button when autoplay is enabled
-    // (This only works when the UI is initialized fast enough to receive the ON_READY event from the player setup)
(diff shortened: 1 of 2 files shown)


src/ts/components/hugeplaybacktogglebutton.ts
```diff
@@ -125,13 +125,38 @@ export class HugePlaybackToggleButton extends PlaybackToggleButton {
...
-    // Enable the transition animations when the play state changes for the first time
-    this.onToggle.subscribeOnce(() => {
-      this.setTransitionAnimationsEnabled(true);
-    });
+    // Hide the play button animation when the UI is loaded (it should only be animated on state changes)
+    suppressPlayButtonTransitionAnimation();
+
+    const isAutoplayEnabled = player.getConfig().playback && Boolean(player.getConfig().playback.autoplay);
+    // We only know if an autoplay attempt is upcoming if the player is not yet ready. It the player is already ready,
+    // the attempt might be upcoming or might have already happened, and we have no way of knowing that.
```
bitmovin-kenny: i would suggest to simplify this line of the comment to `we can rely on the returned value of isPlaying and do not need to worry about autplay`

Issue #103, opened 2018-05-23
Add API to dynamically manage timeline markers


protyposis:
Currently it is possible to add generic `TimelineMarker` markers to the timeline (i.e. `SeekBar`), but only via the `UIConfig` at construction time of the `UIManager`.

The time line marker interface:
https://github.com/bitmovin/bitmovin-player-ui/blob/d8a6978f593ad229e96f051257b7d518d9687619/src/ts/uimanager.ts#L61-L64

The UI config interface that takes an array of timeline markers:
https://github.com/bitmovin/bitmovin-player-ui/blob/d8a6978f593ad229e96f051257b7d518d9687619/src/ts/uimanager.ts#L66-L78

The constructor that takes the config:
https://github.com/bitmovin/bitmovin-player-ui/blob/d8a6978f593ad229e96f051257b7d518d9687619/src/ts/uimanager.ts#L167

We need an API that allows dynamic management of these markers, i.e. add and remove during the lifetime of the `UIManager`. Additionally we want to be able to define different kinds of markers, e.g. chapter and ad markers that need to look and possibly work differently.

Currently these markers mark points in time. We also want to be able to mark time intervals. While this should be easily achievable to mark intervals of the main content, it is going to be more difficult for inserted ad breaks that modify the timeline, so we may want to handle this in a separate follow-up issue once necessary. The API for this would be quite simple, but the question of how this is rendered on a seek bar is to be discussed (e.g. how do we transform the timeline of the seekbar if an ad interval is inserted? how is the mapping from the seekbar to the content done? how do we even know in advance how long an ad break is going to be?).

protyposis:
Proposed API extension:

```ts
interface TimelineMarker {
  /**
   * Optional CSS class that is applied to the marker and can be used to differentiate 
   * different types of markers by their style (e.g. different color of chapter markers and 
   * ad break markers).
   */
  cssClass?: string;
}

class UIManager {
  /**
   * Returns the list of all added markers in undefined order.
   */
  getMarkers(): TimelineMarker[];
  /**
   * Adds a marker to the timeline. Does not check for duplicates/overlaps at the `time`.
   */
  addMarker(timelineMarker: TimelineMarker): void;
  /**
   * Removes a marker from the timeline (by reference) and returns `true` if the marker has 
   * been part of the timeline and successfully removed, or `false` if the marker could not
   * be found and thus not removed.
   */
  removeMarker(timelineMarker: TimelineMarker): boolean;
}
[...]

PR #114, opened 2018-06-21
API to dynamically manage timeline markers

protyposis:
Implements #103

### DIFF
--- a/src/ts/components/seekbar.ts
+++ b/src/ts/components/seekbar.ts
@@ -41,6 +41,11 @@ export interface SeekPreviewEventArgs extends SeekPreviewArgs {
   scrubbing: boolean;
 }
 
+export interface SeekBarMarker {
+  marker: TimelineMarker;
+  position: number;
+}
+
 /**
  * A seek bar to seek within the player's media. It displays the current playback position, amount of buffed data, seek
  * target, and keeps status about an ongoing seek.
@@ -69,7 +74,7 @@ export class SeekBar extends Component<SeekBarConfig> {
 
   private label: SeekBarLabel;
 
-  private timelineMarkers: TimelineMarker[];
+  private timelineMarkers: SeekBarMarker[];
 
   /**
    * Buffer of the the current playback position. The position must be buffered in case the element
@@ -410,18 +415,27 @@ export class SeekBar extends Component<SeekBarConfig> {
   private configureMarkers(player: bitmovin.PlayerAPI, uimanager: UIInstanceManager): void {
     let clearMarkers = () => {
       this.timelineMarkers = [];
-      this.updateMarkers(player);
+      this.updateMarkers();
     };
 
     let setupMarkers = () => {
       clearMarkers();
 
+      const duration = player.getDuration();
+
+      if (duration === Infinity) {
+        // Don't generate timeline markers if we don't yet have a duration
+        // The duration check is for buggy platforms where the duration is not available instantly (Chrome on Android 4.3)
+        return;
+      }
+
       for (let marker of uimanager.getConfig().metadata.markers) {
-        this.timelineMarkers.push(marker);
+        const position = 100 / duration * marker.time; // convert absolute time to percentage
+        this.timelineMarkers.push({ marker, position });
       }
 
       // Populate the timeline with the markers
-      this.updateMarkers(player);
+      this.updateMarkers();
     };
 
     // Add markers when a source is loaded
@@ -523,7 +537,7 @@ export class SeekBar extends Component<SeekBarConfig> {
       seeking = false;
 
       // Fire seeked event
-      this.onSeekedEvent(snappedChapter ? snappedChapter.time : targetPercentage);
+      this.onSeekedEvent(snappedChapter ? snappedChapter.position : targetPercentage);
     };
 
     // A seek always start with a touchstart or mousedown directly on the seekbar.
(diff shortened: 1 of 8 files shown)


src/ts/components/seekbar.ts
```diff
@@ -416,24 +422,18 @@ export class SeekBar extends Component<SeekBarConfig> {
...
-      if (markers && player.getDuration() !== Infinity) {
-        for (let marker of markers) {
-          this.timelineMarkers.push({
-            time: 100 / player.getDuration() * marker.time, // convert time to percentage
-            title: marker.title,
-          });
-        }
+      const duration = player.getDuration();
+
+      if (duration === Infinity) {
```
bitmovin-kenny: I think live videos have `duration == infinity` too, this might break this functionality

bitmovin-kenny: I just read in the comment section of the isse that `We don't support markers on live streams right now` ... so i guess this is ok

src/ts/components/seekbar.ts
```diff
@@ -444,6 +444,11 @@ export class SeekBar extends Component<SeekBarConfig> {
     player.addEventHandler(player.EVENT.ON_READY, setupMarkers);
     // Remove markers when unloaded
     player.addEventHandler(player.EVENT.ON_SOURCE_UNLOADED, clearMarkers);
+    // Remove markers when unloaded
```
bitmovin-kenny: duplicate comment (does not match the `onResize` subscription)

src/ts/components/seekbar.ts
```diff
@@ -607,30 +612,49 @@ export class SeekBar extends Component<SeekBarConfig> {
...
+        'data-marker-time': String(marker.marker.time),
+        'data-marker-title': String(marker.marker.title),
+      }).css(cssProperties));
     }
   }
 
-  protected getMarkerAtPosition(percentage: number): TimelineMarker | null {
-    let snappedMarker: TimelineMarker = null;
-    let snappingRange = 1;
+  protected getMarkerAtPosition(percentage: number): SeekBarMarker | null {
```
bitmovin-kenny: should it be possible to have overlapping markers? if so we must not break in the for loop bur rather collect all matching markers

src/ts/uimanager.ts
```diff
@@ -461,6 +540,35 @@ export class UIManager {
...
+   */
+  getTimelineMarkers(): TimelineMarker[] {
+    return this.config.metadata.markers;
+  }
+
+  /**
+   * Adds a marker to the timeline. Does not check for duplicates/overlaps at the `time`.
+   */
+  addTimelineMarker(timelineMarker: TimelineMarker): void {
+    this.config.metadata.markers.push(timelineMarker);
```
bitmovin-kenny: if we do not allow overlapping markers, we need to check the ranges here

src/ts/components/seekbar.ts
```diff
@@ -607,30 +612,49 @@ export class SeekBar extends Component<SeekBarConfig> {
...
-          snappedMarker = marker;
-          break;
+        if (marker.duration > 0
+          && percentage >= marker.position - snappingRange
+          && percentage <= marker.position + marker.duration + snappingRange) {
+          return marker;
+        }
+        else if (percentage >= marker.position - snappingRange
+          && percentage <= marker.position + snappingRange) {
+          return marker;
```
Reevn: I'm not a fan of multi-line if statements. How about something more expressive and clean?
```ts
if (this.timelineMarkers.length > 0) {
  for (let marker of this.timelineMarkers) {
    const isMarkerInStartRange = percentage >= marker.position - snappingRange;
    const isMarkerInEndRange = percentage <= marker.position + marker.duration + snappingRange;

    if (marker.duration > 0 && isMarkerInStartRange && isMarkerInEndRange) {
      return marker;
    } else if (isMarkerInStartRange && percentage <= marker.position + snappingRange) {
      return marker;
    }
  }
}
```

The variable names could of course be chosen better etc. And the example could be written better but you get my point 👍

src/ts/uimanager.ts
```diff
@@ -461,6 +540,35 @@ export class UIManager {
...
+
+  /**
+   * Returns the list of all added markers in undefined order.
+   */
+  getTimelineMarkers(): TimelineMarker[] {
+    return this.config.metadata.markers;
+  }
+
+  /**
+   * Adds a marker to the timeline. Does not check for duplicates/overlaps at the `time`.
```
Reevn: Is there a use-case for not filtering exact duplicates before adding markers?

src/ts/components/seekbar.ts
protyposis: This infinity check was there before, I just rearranged the code. Never seemed to be an issue, but you're right regarding live streams. Since we don't support them yet, I'd say we leave that for another PR.

src/ts/components/seekbar.ts
protyposis: It is possible (overlaps are not prevented), but it does not make sense and I leave the responsibility to manage that correctly to the user. Also, which `SeekBarLabel` for which marker should be displayed then? All of them? Let's also leave this open until it becomes an issue.

src/ts/uimanager.ts
protyposis: As explained above.

src/ts/components/seekbar.ts
protyposis: I wrote it on purpose the way it is because in your proposal the `isMarkerInEndRange` is always calculated although only needed in case `marker.duration > 0` 😝 But you're right, it's hard to follow.

src/ts/uimanager.ts
protyposis: Yes, I leave that responsibility to the user. If he really wants to have overlaps he can have them. But it does not make sense and will probably never happen in practice, because chapters or ad breaks seldomly overlap. That is unless somebody figures out that these markers can be used for some other creative stuff I didn't think of, but then we can always improve later.

PR #117, opened 2018-06-27
Add filter & translator functions to ListSelector

protyposis:
Adds `ListSelectorConfig.filter` and `ListSelectorConfig.translator` to allow filtering items and translating labels of items added through `ListSelector.addItem`. 

While this does not seem to make a lot of sense in situations where the programmer has control over what is added via `addItem`, it is very useful for subclasses that automatically populate themselves (`SubtitleSelectBox`, `AudioTrackSelectBox`, ...), where no control over the added items was possible until now:

```ts
new SubtitleSelectBox({
  // Filter out all French subtitles (sorry my French friends, we still like you!)
  filter: (listItem) => listItem.key !== 'fr' && !listItem.label.includes('Fran'),

  // Translate the Spanish label but no other
  translator: (listItem) => {
    if (listItem.key === 'es') {
      return 'Aiaiai!';
    }
    return listItem.label;
  },
})
```

### DIFF
--- a/src/ts/components/listselector.ts
+++ b/src/ts/components/listselector.ts
@@ -13,9 +13,8 @@ export interface ListItem {
 /**
  * Filter function that can be used to filter out list items added through {@link ListSelector.addItem}.
  *
- * Using this does not make a lot of sense when items are added from outside because the added items can be
- * controlled anyway, but it does make sense to be used in conjunction with subclasses that populate
- * themselves automatically via the player API, e.g. {@link SubtitleSelectBox}.
+ * This is intended to be used in conjunction with subclasses that populate themselves automatically
+ * via the player API, e.g. {@link SubtitleSelectBox}.
  */
 export interface ListItemFilter {
   /**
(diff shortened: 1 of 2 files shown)


src/ts/components/listselector.ts
```diff
@@ -10,11 +10,45 @@ export interface ListItem {
   label: string;
 }
 
+/**
+ * Filter function that can be used to filter out list items added through {@link ListSelector.addItem}.
+ *
+ * Using this does not make a lot of sense when items are added from outside because the added items can be
```
offnertho: I would remove the judgement of the usage of this interface. Just state what it does. Users should decide on their own whether to use it or not.

src/ts/components/listselector.ts
```diff
@@ -10,11 +10,45 @@ export interface ListItem {
...
+ * controlled anyway, but it does make sense to be used in conjunction with subclasses that populate
+ * themselves automatically via the player API, e.g. {@link SubtitleSelectBox}.
+ */
+export interface ListItemFilter {
```
offnertho: Why don't we add a more generic interface. Like: 
```
export interface Predicate<T> {
(t: T): boolean;
}
```

src/ts/components/listselector.ts
protyposis: This statement explains the usecase.. I will improve it

src/ts/components/listselector.ts
protyposis: Because we don't need it anywhere else. It can be added later if desired, it's just an interface.

src/ts/components/listselector.ts
```diff
@@ -10,11 +10,45 @@ export interface ListItem {
...
+   */
+  (listItem: ListItem): string;
+}
+
 /**
  * Configuration interface for a {@link ListSelector}.
  */
 export interface ListSelectorConfig extends ComponentConfig {
   items?: ListItem[];
+  filter?: ListItemFilter;
```
offnertho: Maybe do a more generic thing such as: 
```
filter?: Predicate<ListItem>;
translator: Producer<ListItem, string>;
```

protyposis: As explained above, this is the only place where these interfaces are used so I think it's not too bad if the names match.

PR #120, opened 2018-07-03
Add demo for extracted audio tracks and subtitle settings from the settings panel direct into the controlBar

stonko1994:


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Added
+- Add demo for extracted audio tracks and subtitle settings from the settings panel direct into the controlBar
```
protyposis: When you add an entry into the `Added` section, please don't start the entry with `Add`, as it is totally redundant. For example ``Demo with separate `SettingsPanel`s for subtitles and audio tracks`` would be sufficient.

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
...
+- Add CSS class on `SubtitleSelectBox` when a subtitle is selected
```
protyposis: What is this necessary for?

src/ts/components/subtitleselectbox.ts
```diff
@@ -51,4 +54,18 @@ export class SubtitleSelectBox extends SelectBox {
     // Populate subtitles at startup
     updateSubtitles();
   }
+
+  /**
+   * Add a css class when subtitles are activated
+   * @param {boolean} activated
+   */
+  private setSubtitleActivatedClass(activated: boolean): void {
```
protyposis: Why is this necessary for the subtitles but not for the audio tracks? Wouldn't this be better placed in the `SelectBox` base class? Is this even necessary or can't the `SelectBox` be natively styled as needed?

Also it seems that you need this only for styling in the demo style file. Please don't build functionality into the core components that is only used for a specific demo, i.e. you're implementing functionality here but styling within the demo file. That does not fit together. When implementing something into UI components, it must always be generic and be of use to all users of the framework, or added externally and specifically for the use case.

src/ts/components/subtitleselectbox.ts
```diff
@@ -34,7 +34,10 @@ export class SubtitleSelectBox extends SelectBox {
     };
 
     this.onItemSelected.subscribe((sender: SubtitleSelectBox, value: string) => {
+      // called twice? first time with null as String second time with NULL
```
protyposis: Where/how is it called twice?

src/ts/uimanager.ts
```diff
@@ -661,7 +661,7 @@ export namespace UIManager.Factory {
     });
   }
 
-  function modernAdsUI() {
+  export function modernAdsUI() {
```
protyposis: What are these exports needed for?

src/ts/uimanager.ts
```diff
@@ -1001,6 +1001,97 @@ export namespace UIManager.Factory {
   }
 }
 
+// Separate the demos from the other factories.
+// The others are more like different 'styles' than some variants with different features
+export namespace UIManager.Demo.Factory {
```
protyposis: Please put this into a separate file. I understand why you put it here, but it's actually not desired to have everything within this file. The `Factory` and other classes will be extracted into separate files with UI v3. They have just been remaining here for the time being (UI v2) to minimize merge conflicts for customers with custom branches when they update the UI framework.

src/ts/uimanager.ts
```diff
@@ -1001,6 +1001,97 @@ export namespace UIManager.Factory {
...
+          components: [
+            new PlaybackToggleButton(),
+            new VolumeToggleButton(),
+            new VolumeSlider(),
+            new Spacer(),
+            new PictureInPictureToggleButton(),
+            new AirPlayToggleButton(),
+            new CastToggleButton(),
+            new VRToggleButton(),
+            new AudioTrackSelectBox({
```
protyposis: All settings must be within a `SettingsPanel`. It doesn't look nice to display a `SelectBox` out of nowhere without a container.

protyposis: See #121

src/ts/components/subtitleselectbox.ts
stonko1994: We wanted to display the 'on-state' icon when a subtitle is selected.
As far as i understand there is always an audio track but you can turn off / on subtitles.

For now it is only used in the demo style file but we could use it too in the future. I think it would be nice to have an indicator in the ui when subtitles are active.

I will have a look at this use-case again in #121

CHANGELOG.md
stonko1994: See [Comment](https://github.com/bitmovin/bitmovin-player-ui/pull/120#discussion_r200019722)

src/ts/components/subtitleselectbox.ts
stonko1994: I forgot to dig into it.

src/ts/components/subtitleselectbox.ts
protyposis: Then why is this implemented in the `SubtitleSelectBox`? Shouldn't this be a feature of the `SettingsToggleButton`?

src/ts/components/subtitleselectbox.ts
stonko1994: It's an issue within the `SelectBox` and its subclasses because the `onItemSelectedEvent` is triggered 1) when the `selectBoxValue` get changed or 2) the player events `ON_SUBTITLE_CHANGED` `ON_AUDIO_CHANGED`.

So far no problem but we set the values in the player (`setAudio` `setSubtitle`) when 1) is triggered and so 2) gets triggered immediately again. 
There is a mechanism in `ListSelector` to prevent these infinite event loop problem by checking the `selectedValue` but these value is null in the first time.

A fix would be to set the `selectedItem` in the `SelectBox` when the value of the DOM element changed.

_I hope this is understandable_

protyposis: As discussed offline, this is known behavior that isn't ideal, but not problematic either.

src/ts/uimanager.ts
stonko1994: Is it possible to easy split it?

The root namespace and the class have the same name -> `UIManager`. 
When you split it, you have to import the `UIManager` class but this will results in an error `Individual declarations in merged declaration 'UIManager' must be all exported or all local.` because of the naming conflict.

Also the export is needed (comment above).

Or am i missing something here.

protyposis: Just move it to a file called `demofactory.ts` and `export namespace DemoFactory` in there (an add the export into `main.ts`. There's (hacky) ways to achieve what you attempted, but it does not have to be under the `UIManager` namespace so let's take the clean approach.

protyposis:
The icons look better now but still odd. The baseline seems off.

image

Shouldn't the red and green lines on the bottom match, similar to the yellow one on top?

stonko1994:
We know that but we said we will leave it as it is until the new designer have time for this.

stonko1994:
This pull request is now part of #122

Issue #121, opened 2018-07-03
Implement ListBox


protyposis:
Add an implementation of the `ListSelector` as a DOM element-based `ListBox`. In contrast to the `SelectBox`, it does not render a dropdown panel but displays the items in a stacked list one above the other. The items could be implemented by reusing the `Button` component.

The `ListBox` should fill a `SettingsPanel` with its items, and the use-case is for example a `SettingsPanel` that opens and directly offers a selection of all subtitles or audio tracks, without the user having to navigate into another "deeper level" like he has to with the dropdown panel of a `SelectBox` (a `SettingsPanel` with a single `SelectBox` is bad UX and unnecessarily complicated). It will thus make sense to implement subclasses like `SubtitleListBox` and `AudioTrackListBox` etc., analogous to `SubtitleSelectBox`, `AudioTrackSelectBox` etc. Ideally, common code would be shared across both implementations because they both build upon the same `ListSelector` interface.

The `SettingsPanelItem` with its label needs to be revised for this, as it makes more sense to have the label on top instead of on the left, wasting a lot of precious space. Or it might be completely left away because the purpose of the panel might be obvious from the button that opens it or the content it displays.

protyposis:
Released in `2.18.0`

Issue #119, opened 2018-07-03
Support paging in SettingsPanel


protyposis:
#### Current state

Different setting groups need to be added into different `SettingsPanel` instances, and these panels need to be switched on/off to switch between the setting groups.

Example: `SettingsToggleButton` opens a `SettingsPanel` with a `SubtitleSettingsOpenButton` that opens another `SettingsPanel` instance, the `SubtitleSettingsPanel`. The `SubtitleSettingsOpenButton` hides the `SettingsPanel` and shows the `SubtitleSettingsPanel`, and the `SubtitleSettingsCloseButton` does the inverse. This results in a odd UX and has the side effect that the `SettingsToggleButton` is only in the active state while it`s "own" `SettingsPanel` is shown, but it does not know anything about "sub-settings panels".

#### Desired state

A `SettingsPanel` should be a container for multiple pages/panes, which can be navigated. In our example, we could have a `MainSettingsPage` and a `SubtitleSettingsPage` within the `SettingsPanel` and switch between them with navigation actions (e.g. via API or buttons). There could even be nice animated transitions. The UX would be nicer and the problem with the wrongly inactive `SettingsToggleButton` solved too.

protyposis:
Released in `3.0.0`

PR #122, opened 2018-07-05
Add list box #121

stonko1994:
Added a new UI Element discussed [here](https://github.com/bitmovin/bitmovin-player-ui/pull/121)


src/ts/components/listbox.ts
```diff
@@ -0,0 +1,173 @@
+import {UIInstanceManager} from '../uimanager';
+import {Button, ButtonConfig} from './button';
+import {ListSelector, ListSelectorConfig} from './listselector';
+import {DOM} from '../dom';
+import {Timeout} from '../timeout';
+
+export interface ListBoxConfig extends ListSelectorConfig {
```
protyposis: The `ListBox` should be a component analogous to `SelectBox`, i.e. it would be used for example as a child inside a `SettingsPanel`. A `SettingsPanel` can contain one or many `ListBox` instances. A `SelectBox` also does not have a cointainer or cares about hiding, that's the business of whatever `Container` the `SelectBox` or `ListBox` night be put into. This is also important in regards to paging (#119), where a `SettingsPanel` will be able to contain multiple `SettingsPanel`s that can be navigated.

It seems what you have created here is a hybrid of a `SettingsPanel` and a `ListBox`. This goes against the design principle of component reuse and composition.

src/ts/components/listboxtogglebutton.ts
```diff
@@ -0,0 +1,68 @@
+import {ToggleButton, ToggleButtonConfig} from './togglebutton';
+import {ListBox} from './listbox';
+import {UIInstanceManager} from '../uimanager';
+
+export interface ListBoxToggleButtonConfig extends ToggleButtonConfig {
```
protyposis: Continuing the comment from above, if the `ListBox` is a child of a `SettingsPanel`, there's also no need for this component but the already existing `SettingsToggleButton` will be used. Of course there can be a `SubtitleSettingsToggleButton` if a button with extended functionality specifically for the subtitle settings panel is required.

src/ts/components/listbox.ts
```diff
@@ -0,0 +1,119 @@
...
+ *       <div class='ui-listboxitem'>
+ *         <button class='ui-listboxitem-button'>label</button>
+ *       </div>
+ *     </div>
+ *     ...
+ *   </div
+ * </code>
+ */
+export class ListBox extends ListSelector<ListSelectorConfig> {
+  private listBoxElement: DOM;
```
protyposis: This field is unnecessary. The `DOM` object is already stored in `Component` and can be retrieved with `getDomElement()`

src/ts/components/listbox.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    this.listBoxElement = listBoxElement;
+    this.updateDomItems();
+
+    return listBoxElement;
+  }
+
+  protected updateDomItems(selectedValue: string = null) {
+    // Delete all children
+    this.listBoxElement.empty();
```
protyposis: Use `this.getDomElement()` instead of `tihs.listBoxElement`

src/ts/components/listbox.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    // Add updated children
+    for (let item of this.items) {
+      let rowElement = new DOM('div', {
+        'class': this.prefixCss(ListBox.LIST_BOX_ITEM_WRAPPER_CLASS),
```
protyposis: What is this wrapper necessary for? Why can't you directly work with the `ListBoxItemButton`?

src/scss/skin-modern/components/_listbox.scss
```diff
@@ -0,0 +1,35 @@
...
+        background-color: transparentize($color-highlight, .15);
+      }
+
+      &:last-child {
+        border-bottom: 0;
+      }
+    }
+  }
+}
+
```
protyposis: You are missing the actual class declaration here:

```scss
.#{$prefix}-ui-listbox {
  @extend %ui-listbox;
}
```

Without that, a basic `ListBox` would not be styled.

src/scss/skin-modern/components/_subtitlesettingspanel.scss
```diff
@@ -0,0 +1,53 @@
+@import '../variables';
+@import '../mixins';
+
+.#{$prefix}-ui-subtitles-settings-panel {
```
protyposis: What is this file necessary for? Why is the styling not covered by `_settingspanel.scss`?

Basically, for every SCSS in the `components` folder, there must be a `Component`, which in this case is violated. You should generify the `SettingsPanel` so it covers also the subtitle settings.

src/scss/skin-modern/components/_subtitlesettingspaneltogglebutton.scss
```diff
@@ -0,0 +1,11 @@
+// demo for extracted audio tracks and subtitle settings from the settings panel direct into the controlBar
+@import '../variables';
```
protyposis: Again there is no equivalent `Component` class. In this case, you can create a `SubtitleSettingsPanelToggleButton extends SettingsPanelToggleButton`.

src/ts/components/subtitleslistbox.ts
```diff
@@ -0,0 +1,55 @@
...
+  constructor(config: ListSelectorConfig = {}) {
+    super(config);
+    this.config = this.mergeConfig(config, {
+      cssClass: 'ui-subtitles-listbox',
+    }, this.config);
+  }
+
+  configure(player: bitmovin.PlayerAPI, uimanager: UIInstanceManager): void {
+    super.configure(player, uimanager);
+    let selectCurrentSubtitle = () => {
```
protyposis: This is basically the same code as in the `SubtitleSelectBox`, so the code should be shared via a helper/utility class.

src/ts/components/subtitleslistbox.ts
```diff
@@ -0,0 +1,55 @@
+import {ListBox} from './listbox';
+import {ListSelectorConfig} from './listselector';
+import {UIInstanceManager} from '../uimanager';
+
+/**
+ * A element that is similar to a select box where the user can select a subtitle
+ */
+export class SubtitlesListBox extends ListBox {
```
protyposis: The name should be in singular, i.e. `SubtitleListBox`, Analogous to `SubtitleSelectBox` (filename too)

src/scss/skin-modern/components/_subtitlesettingspaneltogglebutton.scss
stonko1994: I had one but i was not sure because the only different to the `SettingsPanelToggleButton` was a different `cssClass`.

src/scss/skin-modern/components/_subtitlesettingspanel.scss
stonko1994: It's needed for the different positioning and the little arrow to its toggle button.

src/ts/components/listbox.ts
stonko1994: This will result in an infinite loop because the `updateDomItems` are called from the `getDomElement` function to handle initial state configuration.

Similar to the `SelectBox`

src/scss/skin-modern/components/_subtitlesettingspaneltogglebutton.scss
protyposis: Yes both is not really ideal. Ok then please leave the file in its place but extend the `ui-settingstogglebutton` instead of `ui-button`.

src/scss/skin-modern/components/_subtitlesettingspanel.scss
protyposis: Ok but why is that whole file necessary to just change the positioning? Shouldn't that be much simpler?

src/ts/components/listbox.ts
protyposis: Ok then it's fine :)

src/scss/skin-modern/components/_subtitlesettingspanel.scss
stonko1994: Should be. We can consider using something like a wrapper over the `SettingsToggleButton` and the `SettingsPanel` to 'group' then together. Than the positioning and the little arrow should be easier.

But i would say to do this in a separate issue and then first extract the `Panel` and introduce it also for the `SettingsPanel`

src/ts/components/subtitleslistbox.ts
stonko1994: Hm what about a `SubtitleSwitchHandler` and pass the `bitmovin.PlayerAPI` and a subclass of `ListSelector` and handle all events where something is to update (Player Events & ListSelector Events) within this class?

src/scss/skin-modern/components/_subtitlesettingspanel.scss
protyposis: Even without a wrapper, I don't think it's necessary to replicate that much of the `_settingspanel.scss`. This file also should not contain styles for a `ListBox`.

src/ts/components/subtitleslistbox.ts
protyposis: Sounds good!

src/scss/skin-modern/components/_settingstogglebutton.scss
```diff
@@ -1,15 +1,25 @@
 @import '../variables';
 @import '../mixins';
 
-.#{$prefix}-ui-settingstogglebutton {
+%ui-togglebutton {
```
protyposis: Should be named `ui-settingstogglebutton` (like the file)

src/scss/skin-modern/components/_subtitlesettingspaneltogglebutton.scss
```diff
@@ -0,0 +1,12 @@
...
+
+.#{$prefix}-ui-subtitles-settings-toggle-button {
```
protyposis: `subtitle` (singular)

src/ts/subtitleutils.ts
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Helper class to handle all subtitle related events
+ *
+ * This class listens to player events as well as the `ListSelector` event if selection changed
+ */
+export class SubtitleSwitchHandler {
+  private player: bitmovin.PlayerAPI;
+  private listElement: ListSelector<ListSelectorConfig>;
+
+  constructor(player: bitmovin.PlayerAPI, element: SubtitleListBox | SubtitleSelectBox) {
```
protyposis: Can't the type be `element: ListSelector`?

src/ts/subtitleutils.ts
```diff
@@ -0,0 +1,65 @@
...
+    this.player = player;
+    this.listElement = element;
+
+    this.bindSelectionEvent();
+    this.bindPlayerEvents();
+    const callback = this.buildUpdateSubtitlesCallback();
```
protyposis: Why do you need a `build` method instead of calling it directly? (Applies to all usages of this method)

src/ts/subtitleutils.ts
```diff
@@ -0,0 +1,65 @@
...
+import {SubtitleListBox} from './components/subtitlelistbox';
+import {SubtitleSelectBox} from './components/subtitleselectbox';
+
+/**
+ * Helper class to handle all subtitle related events
+ *
+ * This class listens to player events as well as the `ListSelector` event if selection changed
+ */
+export class SubtitleSwitchHandler {
+  private player: bitmovin.PlayerAPI;
```
protyposis: Please always add a blank line between the `class` declaration and the first field/constructor/method/etc.

src/ts/subtitleutils.ts
```diff
@@ -0,0 +1,65 @@
...
+    this.player = player;
+    this.listElement = element;
+
+    this.bindSelectionEvent();
+    this.bindPlayerEvents();
+    const callback = this.buildUpdateSubtitlesCallback();
+    callback();
+  }
+
+  bindSelectionEvent(): void {
```
protyposis: Make all methods private

src/scss/skin-modern/components/_subtitlesettingspanel.scss
```diff
@@ -0,0 +1,35 @@
...
+  @extend %ui-settings-panel;
+
+  right: 3em;
+  width: 7em;
+
+  // bottom arrow from http://www.cssarrowplease.com/
+  &::after {
```
protyposis: How about we remove that arrow completely and with that get rid of the need for correct placement? Looks childish anyway. What do you think?

src/ts/components/subtitlelistbox.ts
```diff
@@ -0,0 +1,22 @@
...
+    super(config);
+    this.config = this.mergeConfig(config, {
+      cssClass: 'ui-subtitles-listbox',
+    }, this.config);
+  }
+
+  configure(player: bitmovin.PlayerAPI, uimanager: UIInstanceManager): void {
+    super.configure(player, uimanager);
+
+    new SubtitleSwitchHandler(player, this);
```
protyposis: very nice :)

src/ts/uimanager.ts
```diff
@@ -1061,11 +1072,9 @@ export namespace UIManager.Demo.Factory {
...
-            new AudioTrackSelectBox({
-              cssClasses: ['my-audio-track-selection-demo'],
-            }),
-            new SubtitleSelectBox({
-              cssClasses: ['my-subtitle-selection-demo'],
+            new SettingsToggleButton({
+              settingsPanel: subtitleSettingsPanel,
+              cssClass: 'ui-subtitles-settings-toggle-button',
             }),
             new SettingsToggleButton({ settingsPanel: settingsPanel }),
```
protyposis: You need to hide all visible settingspanels when another panel is opened. Currently we can have both visible at the same time. You can achieve this via the `UIInstanceManager.onComponentShow` event:

```ts
// Keep track of shown SettingsPanels
const shownSettingsPanels: SettingsPanel[] = [];
uiManager.onComponentShow.subscribe((sender) => {
  if (sender instanceof SettingsPanel) {
    shownSettingsPanels.push(sender);
    sender.onComponentHide.subscribeOnce(() => ArrayUtils.remove(shownSettingsPanels, sender));
  }
});

// Hide all open SettingsPanels before opening this button's panel
// (We need to iterate a copy because hiding them will automatically remove themselves from the array
// due to the subscribeOnce above)
shownSettingsPanels.slice().forEach(settingsPanel => settingsPanel.hide());

mySettingsPanel.show();
```

(this code is not tested but should give you the idea)

src/ts/components/listbox.ts
```diff
@@ -0,0 +1,115 @@
...
+        key: item.key,
+        text: item.label,
+      });
+
+      itemButton.onClick.subscribe((sender) => {
+        this.handleSelectionChange(<ListBoxItemButton>sender);
+      });
+
+      const selectedItemClass = this.prefixCss(ListBox.SELECTED_LIST_BOX_ITEM_CLASS);
+      const itemElement = itemButton.getDomElement();
```
protyposis: This is actually a hack. To do it right, the `ListSelector` must be converted to extend `Container` and the `SelectBox` needs to be adjusted to that. Currently you are managing `Button` instances outside of the component tree. This might have side effects. But lets let that be for the moment.

src/ts/subtitleutils.ts
stonko1994: because when passing the method direct to the player event callbacks we are loosing the lexical scope.

protyposis: You mean the `this` context (binding)?

src/scss/skin-modern/components/_subtitlesettingspanel.scss
stonko1994: hm yes we can do it. I'm fine with that.

src/ts/subtitleutils.ts
stonko1994: yes

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [develop]
 
 ### Added
+- New `ListBox` UI element to display multiple selectable items
```
protyposis: `New` is redundant. It's implicitly new as it's in the `Added` section. Just writing `` `ListBox` UI element`` is sufficient.

Also add an entry for the demo, e.g.

```
### Added
- Demo section in playground with separate `SettingsPanel`s for subtitle and audio track selection via `ListBox`
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
 - Icon for `Subtitles`
```
protyposis: `Subtitles` and `AudioTracks` are not componentos or other classes within the UI framework so they don't belong into `code` markup

src/scss/skin-modern/components/_audiotracklistbox.scss
```diff
@@ -0,0 +1,6 @@
+@import '../variables';
+@import '../mixins';
+
+.#{$prefix}-ui-audiotrack-listbox {
```
protyposis: This file is unnecessary. There is also no `_audiotrackselectbox.scss`.

src/scss/skin-modern/components/_audiotracksettingspanel.scss
```diff
@@ -0,0 +1,29 @@
+@import '../variables';
+@import '../mixins';
+
+.#{$prefix}-ui-audiotrack-settings-panel {
+  @extend %ui-settings-panel;
+
+  right: 5em;
```
protyposis: The settings panels should all be rendered in the same place by default.

src/scss/skin-modern/components/_subtitlesettingspanel.scss
```diff
@@ -0,0 +1,29 @@
+@import '../variables';
+@import '../mixins';
+
+.#{$prefix}-ui-subtitle-settings-panel {
+  @extend %ui-settings-panel;
+
+  right: 3em;
```
protyposis: Same here. Use default positioning.

src/ts/main.ts
```diff
@@ -131,6 +132,7 @@ const playerui = {
   HugePlaybackToggleButton,
   HugeReplayButton,
   Label,
+  ListBox,
```
protyposis: Also export the audio track and subtitle list boxes

src/ts/subtitleutils.ts
protyposis: The `build*` methods need to be removed. Write the methods directly as arrow functions to get rid of the function return value.

src/ts/demofactory.ts
```diff
@@ -90,11 +110,13 @@ export namespace DemoFactory {
             new AirPlayToggleButton(),
             new CastToggleButton(),
             new VRToggleButton(),
-            new AudioTrackSelectBox({
-              cssClasses: ['my-audio-track-selection-demo'],
+            new SettingsToggleButton({
+              settingsPanel: audioTrackSettingsPanel,
+              cssClass: 'ui-audiotrack-setting-toggle-button',
```
protyposis: The `cssClass` of the `SettingsToggleButton` is `ui-settingstogglebutton`, so this should be something along `ui-audiotracksettingstogglebutton` (the pattern is always to extend the parent's name).

src/ts/demofactory.ts
```diff
@@ -90,11 +110,13 @@ export namespace DemoFactory {
...
             }),
-            new SubtitleSelectBox({
-              cssClasses: ['my-subtitle-selection-demo'],
+            new SettingsToggleButton({
+              settingsPanel: subtitleSettingsPanel,
+              cssClass: 'ui-subtitle-setting-toggle-button',
```
protyposis: As above.

src/scss/skin-modern/components/_audiotracklistbox.scss
protyposis: Same goes for `_subtitlelistbox.scss`

src/scss/skin-modern/components/_audiotracksettingspanel.scss
```diff
@@ -0,0 +1,28 @@
+@import '../variables';
+@import '../mixins';
+
+.#{$prefix}-ui-audiotrack-settings-panel {
```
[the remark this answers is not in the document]
stonko1994: To improve styling for a `settings-panel` when there is a `list-box` in it. Same for `_subtitlesettingspanel.scss` but can be combined.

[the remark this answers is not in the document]
protyposis: And it's not possible to tweak the base style?

src/ts/components/listbox.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+export class ListBox extends ListSelector<ListSelectorConfig> {
+  private listBoxElement: DOM;
+
+  private static readonly SELECTED_LIST_BOX_ITEM_CLASS = 'ui-listboxitem-button-selected';
+
+  constructor(config: ListSelectorConfig = {}) {
+    super(config);
+
+    this.config = this.mergeConfig(config, {
```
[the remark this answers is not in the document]
stonko1994: Done in `subtitlelistbox.ts`.

src/scss/skin-modern/components/_audiotracksettingspanel.scss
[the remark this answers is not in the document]
stonko1994: Found a solution for the label but not for the margin / padding.

PR #124, opened 2018-07-06
Hide hovered select boxes when the whole UI hides

bitmovin-kenny:
The fix for hiding hovered select boxes: https://github.com/bitmovin/bitmovin-player-ui/pull/116 did not cover the case, that the whole UI is hiding.

We now listen to the `uimanager.onControlsHide` to apply the same functionality on the global hiding of the UI

### DIFF
--- a/src/ts/components/settingspanel.ts
+++ b/src/ts/components/settingspanel.ts
@@ -53,6 +53,8 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
         this.hideHoveredSelectBoxes();
       });
 
+      uimanager.onControlsHide.subscribe(this.hideHoveredSelectBoxes);
+
       this.onShow.subscribe(() => {
         // Activate timeout when shown
         this.hideTimeout.start();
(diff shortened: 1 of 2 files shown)


src/ts/components/settingspanel.ts
```diff
@@ -53,6 +53,8 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
         this.hideHoveredSelectBoxes();
       });
 
+      uimanager.onControlsHide.subscribe(this.hideHoveredSelectBoxes);
```
stonko1994: I think the event should be bound outside the if checking for the configuration if the `SettingsPanel` should hide automatically.

Because when the automatic hide for the `SettingsPanel` is disabled the event never get bound and if the whole UI hides the select box is still floating.

bitmovin-kenny:
I thought it would be a decrement to the reading flow of the setup function. 
The setup already has 50 lines and puttin in another 20 for an internal function would make it even harder to understand what is going on. @protyposis would you still prefer the internal function?

protyposis:
No it's fine. I was just expecting an explanation for why a non-preferred solution is chosen, but maybe I shouldn't have expectations ;)

Anyway FYI, even if the method would be inside `configure`, that would be one of the shorter `configure` methods in the framework. If it really gets bad, they are usually split up by purpose into separate methods like here:

https://github.com/bitmovin/bitmovin-player-ui/blob/52b8007961f9454f5224c505140cd6961c9d959e/src/ts/components/uicontainer.ts#L46-L51

ml242:
Thanks for the quick fix. Can you let me know when it gets included in a release?

bitmovin-kenny:
We will notify you as soon as a new player version is released which includes this fix

bitmovin-kenny:
@ml242 We just released player version 7.7.4 which comes with UI version (2.17.0), which includes this fix.

PR #126, opened 2018-07-11
Single image thumbnails

unki2aut:
We did always expect x/y/w/h values for a thumbnail. With single images those properties don't have to be present.

An example file looks like this:
```
WEBVTT

00:00:00.0 --> 00:00:06.142
http://example.com/images/1.jpg

00:00:06.142 --> 00:00:12.285
http://example.com/images/2.jpg
```

This PR creates the CSS properties accordingly for the single image.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [develop]
 
 ### Fixed
-- Rendering of single-image thumbnails when position and size properties are missing.
+- Rendering of single-image thumbnails
 
 ## [2.17.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- Create CSS properties for sprite and single images differently
+
 ## [2.17.0]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [develop]
 
 ### Fixed
-- Create CSS properties for sprite and single images differently
+- Rendering of single-image thumbnails when position and size properties are missing.
 
 ## [2.17.0]
 
--- a/src/ts/components/seekbarlabel.ts
+++ b/src/ts/components/seekbarlabel.ts
@@ -139,7 +139,7 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
         if (['x', 'y', 'w', 'h'].every((key) => thumbnail.hasOwnProperty(key))) {
           thumbnailElement.css(this.thumbnailCssSprite(thumbnail, width, height));
         } else {
-          thumbnailElement.css(this.thumbnailCssSingeImage(thumbnail, width, height));
+          thumbnailElement.css(this.thumbnailCssSingleImage(thumbnail, width, height));
         }
       });
     }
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- Create CSS properties for sprite and single images differently
```
```
### Fixed
- Rendering of single-image thumbnails
```

src/ts/components/seekbarlabel.ts
```diff
@@ -136,31 +136,50 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
...
-        // used because it is relative to the width and can be used to set the aspect ratio of the thumbnail.
-        // A default value for width is set in the stylesheet and can be overwritten from there or anywhere else.
-        thumbnailElement.css({
-          'display': 'inherit',
-          'background-image': `url(${thumbnail.url})`,
-          'padding-bottom': `${100 * aspectRatio}%`,
-          'background-size': `${sizeX}% ${sizeY}%`,
-          'background-position': `-${offsetX}% -${offsetY}%`,
-        });
+        if (['x', 'y', 'w', 'h'].every((key) => thumbnail.hasOwnProperty(key))) {
```
protyposis: That could have been much simpler and faster (no function calls):

```ts
if (thumbnail.x === undefined) {
  // Thumbnails without x/y/w/h values are single-image thumbnails, not sprites within an image
  // We set these values to simulate a single-sprite image whose sprite covers the whole image
  thumbnail.x = 0;
  thumbnail.y = 0;
  thumbnail.w = width;
  thumbnail.h = height;
}
```

unki2aut: I'd prefer the done approach as for single images we need no `background-position` and can also omit the calculations.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- Rendering of single-image thumbnails when position and size properties are missing.
```
protyposis: This still contains technical details that someone who is simply "using" (configuring) the UI is not aware of. Also we do not write full stops :)

src/ts/components/seekbarlabel.ts
protyposis: Then please simplify the check to `if (thumbnail.x === undefined) ...`

src/ts/components/seekbarlabel.ts
```diff
@@ -136,31 +136,50 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
...
+          thumbnailElement.css(this.thumbnailCssSprite(thumbnail, width, height));
+        } else {
+          thumbnailElement.css(this.thumbnailCssSingeImage(thumbnail, width, height));
+        }
       });
     }
   }
+
+  private thumbnailCssSprite(thumbnail: bitmovin.PlayerAPI.Thumbnail, width: number, height: number): any {
```
protyposis: Add return type

src/ts/components/seekbarlabel.ts
```diff
@@ -136,31 +136,50 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
...
+    return {
+      'display': 'inherit',
+      'background-image': `url(${thumbnail.url})`,
+      'padding-bottom': `${100 * aspectRatio}%`,
+      'background-size': `${sizeX}% ${sizeY}%`,
+      'background-position': `-${offsetX}% -${offsetY}%`,
+    };
+  }
+
+  private thumbnailCssSingeImage(thumbnail: bitmovin.PlayerAPI.Thumbnail, width: number, height: number): any {
```
protyposis: Typo in `Sigle`, add return type

src/ts/components/seekbarlabel.ts
```diff
@@ -136,31 +136,50 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
...
+    let aspectRatio = 1 / width * height;
+
+    return {
+      'display': 'inherit',
+      'background-image': `url(${thumbnail.url})`,
+      'padding-bottom': `${100 * aspectRatio}%`,
+      'background-size': `100% 100%`,
```
protyposis: You still need to set `background-position: 0 0` (in case a thumbnail sprite was displayed before)

unki2aut:
@protyposis you're right on the check, but it's possible to omit any of these properties. So probably we should rather find a solution that allows for all combinations.

PR #127, opened 2018-07-20
support player v8 events

bitmovin-kenny:
In bitmovin player Version 8 the player.EVENT got a major update.
All Events are being renamed from `ON_EVENT_NAME = onEventName` to `EventName = eventname`.
This PR updates all usages of the Events in the UI as well as in the player definition

As this is a breaking API change we need a UI Version 3, because we cannot be backwards compatible with player Version 7


src/ts/player-events.d.ts
```diff
@@ -12,87 +12,79 @@ declare namespace bitmovin {
...
+      VideoDownloadQualityChanged: EVENT;
+      VideoPlaybackQualityChanged: EVENT;
+      VolumeChanged: EVENT;
+      VRStereoChanged: EVENT;
+      Warning: EVENT;
+      PictureInPictureEnter: EVENT;
+      PictureInPictureExit: EVENT;
+      AirplayAvailable: EVENT;
+      VRViewingDirectionChange: EVENT;
+      VRViewingDirectionChanged: EVENT;
```
protyposis: There's obviously a few events missing here. Have they been removed on purpose?

bitmovin-kenny: Some Events were removed with V8:
`onShowControls`, `onHideControls` were completely removed
`onCastPlaying`, `onCastPlaybackFinished`, `onCastPaused`, `onCastTimeUpdated` have been replaced with the "normal" player events a while ago and have now been removed

PR #128, opened 2018-07-24
Track initiated play for player v8

jmsn:
In player v8 the API method `isPlaying` would only return `true` once the media element has started playback. In previous versions it would already return `true` once the play was initiated (i.e. `player.play` was called). 

For player v8 we need to track the state of a play being initiated separately so the play buttons can react to that. I am using the `onPlay` event for that matter.

### DIFF
--- a/src/ts/components/playbacktogglebutton.ts
+++ b/src/ts/components/playbacktogglebutton.ts
@@ -11,6 +11,7 @@ import WarningEvent = bitmovin.PlayerAPI.WarningEvent;
 export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
 
   private static readonly CLASS_STOPTOGGLE = 'stoptoggle';
+  protected isPlayInitiated: boolean;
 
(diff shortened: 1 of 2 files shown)


src/ts/components/playbacktogglebutton.ts
```diff
@@ -34,19 +37,30 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
...
       } else {
         this.off();
       }
     };
 
     // Call handler upon these events
-    player.addEventHandler(player.EVENT.ON_PLAY, playbackStateHandler);
-    player.addEventHandler(player.EVENT.ON_PAUSED, playbackStateHandler);
+    player.addEventHandler(player.EVENT.ON_PLAY, (e) => {
+      this.isPlayInitiated = true;
```
protyposis: You could have added this to the `playbackStateHandler`:

```ts
this.isPlayInitiated = e.type === player.EVENT.ON_PLAY;
```

PR #132, opened 2018-08-08
Prevent controls from hiding in certain player states

protyposis:
Until now it was only possible to configure if the UI controls will be hidden after a certain amount of inactivity time (`UIContainerConfig.hideDelay`) or if they are always shown.

This PR adds `UIContainerConfig.hidePlayerStateExceptions` that allows to specify a list of player states in which the UI controls will always be shown, e.g.

https://github.com/bitmovin/bitmovin-player-ui/blob/b2e1f004086d12c6137bd8c0d3edfb56109d118a/src/ts/uimanager.ts#L776

In the small screen UI we prevent the UI controls from hiding in the replay screen to avoid a required double tap on smartphones to first display the controls and then start playback. By always showing the controls, replay can be started with one tap and buttons like the fullscreen button are always easily available.

### DIFF
--- a/src/ts/components/uicontainer.ts
+++ b/src/ts/components/uicontainer.ts
@@ -68,6 +68,10 @@ export class UIContainer extends Container<UIContainerConfig> {
     let isFirstTouch = true;
     let playerState: PlayerUtils.PlayerState;
 
+    const hidingPrevented = (): boolean => {
+      return config.hidePlayerStateExceptions && config.hidePlayerStateExceptions.indexOf(playerState) > -1;
+    };
+
     let showUi = () => {
       if (!isUiShown) {
         // Let subscribers know that they should reveal themselves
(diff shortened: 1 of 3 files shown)


src/ts/components/uicontainer.ts
```diff
@@ -139,6 +148,13 @@ export class UIContainer extends Container<UIContainerConfig> {
     player.addEventHandler(player.EVENT.ON_CAST_STARTED, () => {
       showUi(); // Show UI when a Cast session has started (UI will then stay permanently on during the session)
     });
+    this.playerStateChange.subscribe((_, state) => {
+      playerState = state;
+      if (hidingPrevented()) {
+        this.uiHideTimeout.clear();
+        showUi();
+      }
```
stonko1994: There is one issue when the user press `replay` (`playerState` change from `finished` to `playing`) the UI will not hide automatically.

Add something like:
```
      } else {
        if (isUiShown) {
          this.uiHideTimeout.start();
        }
      }
```

PR #133, opened 2018-08-08
Adjust to player getFigure removal

bitmovin-kenny:
`player.getFigure` has been removed with V8 along with the actual figure
`player.getContainer` now returns the element which embeds the player, use that one

### DIFF
--- a/src/ts/player.d.ts
+++ b/src/ts/player.d.ts
@@ -179,7 +179,7 @@ declare namespace bitmovin {
      */
     getDuration(): number;
     /**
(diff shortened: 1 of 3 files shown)


src/ts/player.d.ts
```diff
@@ -181,7 +181,7 @@ declare namespace bitmovin {
     /**
      * Returns the figure element that the player is embedded in, if the player is set up, or null otherwise.
```
protyposis: Please update the text

PR #136, opened 2018-08-14
Settings panel paging / navigation

stonko1994:
### Description

Issue: #119


src/ts/components/settingspanel.ts
```diff
@@ -72,28 +103,125 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
...
     }
+
+    const width = clone.scrollWidth + widthOffset;
+    const height = clone.scrollHeight + heightOffset;
+
+    // remove clone from the DOM
+    clone.remove();
+
+    // set 'real' width / height
+    domElement.css('width', width + 'px');
```
protyposis: Do with one `css` call

src/ts/components/settingspanelnavigatorbutton.ts
```diff
@@ -0,0 +1,45 @@
...
+   */
+  container: SettingsPanel;
+  /**
+   * Page where the button should navigate to
+   * If empty it will navigate to the root page (not intended to use as navigate back behavior)
+   */
+  targetPage?: SettingsPanelPage;
+}
+
+export class SettingsPanelNavigatorButton extends Button<SettingsPanelNavigatorConfig> {
```
protyposis: Please call them `SettingsPanelPageNavigatorButton` and `SettingsPanelPageNavigatorButtonConfig` (yes its along name but that's the pattern)

src/ts/components/settingspanel.ts
```diff
@@ -72,28 +103,125 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
...
+    const htmlElement = domElement.get(0);
+    // ensure container has real width / height (for first animation)
+    if (htmlElement.style.width === '' || htmlElement.style.height === '') {
+      domElement.css('width', domElement.css('width'));
+      domElement.css('height', domElement.css('height'));
+    }
+
+    const targetPageHtmlElement = targetPage.getDomElement().get(0);
+    const clone = targetPageHtmlElement.cloneNode(true) as HTMLElement;
+    // append to parent so we get the 'real' size
```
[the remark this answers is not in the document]
stonko1994: i will update the comment regarding the clone.

src/ts/components/settingspanel.ts
```diff
@@ -72,28 +103,132 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
...
+  }
+
+  popToRootSettingsPanelPage(): void {
+    this.resetNavigation();
+  }
+
+  popSettingsPanelPage() {
+    // pop one navigation item from stack
+    const currentPage = this.navigationStack.pop(); // remove current page
+    const targetPage = this.navigationStack.slice(-1)[0]; // pick target page without removing it
```
bitmovin-kenny: i would prefer `this.navigationStack[this.navigationStack.length - 1]` as it does the same but expresses the intent better IMO

src/ts/components/settingspanel.ts
```diff
@@ -72,28 +103,132 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
...
+        page.getDomElement().addClass(SettingsPanel.ACTIVE_PAGE_CLASS);
+      } else {
+        page.getDomElement().removeClass(SettingsPanel.ACTIVE_PAGE_CLASS);
       }
+    });
+  }
+
+  setActivePageIndex(index: number): void {
+    const targetPage = this.getPages()[index];
+    if (targetPage !== undefined) {
```
bitmovin-kenny: you can just use `if (targetPage)`, this will also avoid `null` passing your check

src/ts/components/settingspanel.ts
```diff
@@ -72,28 +103,132 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
...
+  popToRootSettingsPanelPage(): void {
+    this.resetNavigation();
+  }
+
+  popSettingsPanelPage() {
+    // pop one navigation item from stack
+    const currentPage = this.navigationStack.pop(); // remove current page
+    const targetPage = this.navigationStack.slice(-1)[0]; // pick target page without removing it
+
+    if (targetPage !== undefined) {
```
bitmovin-kenny: just omitt the `!== undefined`

src/ts/components/settingspanel.ts
```diff
@@ -72,28 +103,132 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
...
+    if (!(this.config as SettingsPanelConfig).pageTransitionAnimation)
+      return;
+    // workaround to enable css transition for elements with auto width / height property
+    // css transition does not work with auto properties by definition so we need to calculate 'real'
+    // width / height values to have a nice looking animation
+
+    const domElement = this.getDomElement();
+    const htmlElement = domElement.get(0);
+    // ensure container has real width / height (for first animation)
+    if (htmlElement.style.width === '' || htmlElement.style.height === '') {
```
bitmovin-kenny: i'm not entirely sure if this is a problem here, but as far as i remember using ele.style.width/height is discuraged. 

see https://stackoverflow.com/questions/294250/how-do-i-retrieve-an-html-elements-actual-width-and-height for alternatives (`getBoundingClientRect` seems to be the way to go)

stonko1994: It is intended in this case.

src/ts/components/settingspanel.ts
```diff
@@ -72,28 +103,132 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
...
+      // add margins
+      widthOffset += getNumberOfCss(computedStyles.marginLeft) + getNumberOfCss(computedStyles.marginRight);
+      heightOffset += getNumberOfCss(computedStyles.marginTop) + getNumberOfCss(computedStyles.marginBottom);
     }
+
+    const width = clone.scrollWidth + widthOffset;
+    const height = clone.scrollHeight + heightOffset;
+
+    // remove clone from the DOM
+    clone.remove();
```
bitmovin-kenny: This line does not work in IE11 :(

bitmovin-kenny: `clone.parentElement.removeChild(clone);` should does the job

src/ts/components/settingspanel.ts
```diff
@@ -20,15 +21,29 @@ export interface SettingsPanelConfig extends ContainerConfig {
...
+  pageTransitionAnimation?: boolean;
 }
 
 /**
  * A panel containing a list of {@link SettingsPanelItem items} that represent labelled settings.
  */
 export class SettingsPanel extends Container<SettingsPanelConfig> {
 
-  private static readonly CLASS_LAST = 'last';
+  private static readonly ACTIVE_PAGE_CLASS = 'active';
```
protyposis: Statically defined class string are always prefixed with `CLASS_` thoughout this project, not suffixed.

src/ts/components/settingspanel.ts
```diff
@@ -72,28 +103,132 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
...
-      }
-      if (lastShownItem) {
-        lastShownItem.getDomElement().addClass(this.prefixCss(SettingsPanel.CLASS_LAST));
+    this.updateActivePageClass();
+  }
+
+  private updateActivePageClass(): void {
+    this.getPages().forEach((page: SettingsPanelPage, index) => {
+      if (index === this.activePageIndex) {
+        page.getDomElement().addClass(SettingsPanel.ACTIVE_PAGE_CLASS);
```
protyposis: You need to prefix with `this.prefixCss`, also in any other place where this class is used.

src/ts/components/settingspanel.ts
```diff
@@ -152,11 +294,21 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
 
   /**
    * Gets the event that is fired when one or more {@link SettingsPanelItem items} have changed state.
+   * @deprecated
```
protyposis: Why is this deprecated? Can't it be removed? A deprecated element in a major release does not make a lot of sense.

src/ts/components/settingspanel.ts
```diff
@@ -152,11 +294,21 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
...
   get onSettingsStateChanged(): Event<SettingsPanel, NoArgs> {
     return this.settingsPanelEvents.onSettingsStateChanged.getEvent();
   }
+
+  addComponent(component: Component<ComponentConfig>) {
+    if (component instanceof SettingsPanelPage) {
+      super.addComponent(component);
+    } else {
+      // backwards compatibility
+      this.getRootPage().addComponent(component);
```
protyposis: It must not be anything other than a `SettingsPanelPage`. Remove the backwards compatibility, not necessary for v3.

PR #143, opened 2018-08-21
Player resized event not working

stonko1994:
## Description

Found out that it was not possible to listen on the `PlayerResized` event.

## Fix

Update wording of Enum value for the event to match name in the player


bitmovin-kenny:
But why is the build failing?

stonko1994:
@bitmovin-kenny don't know. Just restarted it and now it worked.

PR #144, opened 2018-08-21
Fix an issue that many components are not updating the content after an ui variant switch

stonko1994:
## Description

Intentionally I was investigating a issue that the video-quality select box value was not updating correctly but I was not able to reproduce it in UI v3 with player v8. 

I found another issue that after an UI variant switch some components were not able to update their state again.

## Fix

Partly revert of [fd817f](https://github.com/bitmovin/bitmovin-player-ui/commit/fd817fc9eb495237e103aba9c68ad8da462cc7aa)

### DIFF
--- a/src/ts/uimanager.ts
+++ b/src/ts/uimanager.ts
@@ -497,7 +497,7 @@ export class UIManager {
      * undesirable at this time. */
     this.uiContainerElement.append(dom);
 
-    // Some components initialize their state on ON_READY. When the UI is loaded after the player is already ready,
+    // Some components initialize their state on Ready. When the UI is loaded after the player is already ready,
     // they will never receive the event so we fire it from here in such cases.
     if (player) {
       player.fireEventInUI(player.exports.Event.Ready, {});
--- a/src/ts/uimanager.ts
+++ b/src/ts/uimanager.ts
@@ -489,12 +489,19 @@ export class UIManager {
 
   private addUi(ui: InternalUIInstanceManager): void {
     let dom = ui.getUI().getDomElement();
+    let player = ui.getWrappedPlayer();
 
     ui.configureControls();
     /* Append the UI DOM after configuration to avoid CSS transitions at initialization
      * Example: Components are hidden during configuration and these hides may trigger CSS transitions that are
      * undesirable at this time. */
     this.uiContainerElement.append(dom);
+
+    // Some components initialize their state on ON_READY. When the UI is loaded after the player is already ready,
+    // they will never receive the event so we fire it from here in such cases.
+    if (player) {
+      player.fireEventInUI(player.exports.Event.Ready, {});
+    }
 
     // Fire onConfigured after UI DOM elements are successfully added. When fired immediately, the DOM elements
     // might not be fully configured and e.g. do not have a size.


src/ts/uimanager.ts
```diff
@@ -489,13 +489,20 @@ export class UIManager {
...
     let dom = ui.getUI().getDomElement();
+    let player = ui.getWrappedPlayer();
 
     ui.configureControls();
     /* Append the UI DOM after configuration to avoid CSS transitions at initialization
      * Example: Components are hidden during configuration and these hides may trigger CSS transitions that are
      * undesirable at this time. */
     this.uiContainerElement.append(dom);
 
+    // Some components initialize their state on ON_READY. When the UI is loaded after the player is already ready,
```
Reevn: The `ON_READY` event is called `Ready` starting with v8

src/ts/uimanager.ts
```diff
@@ -489,13 +489,20 @@ export class UIManager {
...
+    // they will never receive the event so we fire it from here in such cases.
+    if (player) {
+      player.fireEventInUI(player.exports.Event.Ready, {});
```
bitmovin-kenny: The semantic of the Ready Event has changed, Ready is not fired when the player has enough data to actually start playback.
It should not be a concern of this PR, i just at some point wanted to point out, that all reactions to Ready might potentially be wrong. The event sequence can now be:
`Loaded`, `Play`, `Ready`, `Playing` where before `Ready` was always the first event to be fired, when setup was complete

stonko1994: Is this a common use case? eg. when `autoplay` is enabled?

Reevn: It all depends on the client and how long it takes to download enough data to be able to start playback.

protyposis: `ON_READY` was used in v2 to initialize the components. When a UI instance was created after player setup (i.e. after `ON_READY`), the `UIManager` fired an `ON_READY` into the UI components to trigger their initialization. 

Since `Event.Ready` is something totally different, the same logic doesn't apply in v3 any more and needs to be rewritten. Components now need to (re)initialize themselves on `Event.SourceLoaded` instead of `Event.Ready`. The `UIManager` must fire `Event.SourceLoaded` instead of `Event.Ready` when the UI is created and `player.getSource() !== null` (which is semantically similar to `player.isReady()`). Or event better, insteaf of firing a "fake" event, the components can just check `player.getSource() !== null` themselves (this could even be implemented into the player wrapper as an `isSourceLoaded(): boolean` helper method).

The `Event.Ready` is totally irrelevant for UI v3.

stonko1994: I am confused now.

@bitmovin-kenny you wrote `Ready is not fired when the player has enough data to actually start playback.`. 
But @Reevn pointed out the in the definition it says `Is fired when the player has enough data to start playback`.


@protyposis how can there be a `Event.SourceLoaded` event and the `player.getSource()` return `null` afterwards. And can i merge this or should i rewrite the 'on-Ready' occurrences.

Reevn: I think @bitmovin-kenny just accidentally wrote 'not' instead of 'now'

protyposis: > how can there be a `Event.SourceLoaded` event and the `player.getSource()` return `null` afterwards. And can i merge this or should i rewrite the 'on-Ready' occurrences

There can't be. But for the case the UI is loaded *after* `SourceLoaded`, you need to check `player.getSource() != null` to know that a source is loaded and the `SourceLoaded` event was already fired. In that case, you need to `player.fireEventInUI(player.exports.Event.SourceLoaded, {});` to thell the UI that a source is loaded and the components to initialize themselves:

```ts
if (player.getSource()) {
  player.fireEventInUI(player.exports.Event.SourceLoaded, {});
}
```

instead of https://github.com/bitmovin/bitmovin-player-ui/blob/15b6fc46fd877543e6a32d97357812ee066eb89b/src/ts/uimanager.ts#L500-L504

bitmovin-kenny: Sorry for the confusion, i really meant "now" not "not"

dweinber: I think we should merge this PR and create a separate one for the `Ready` --> `SourceLoaded` change @stonko1994.

PR #145, opened 2018-08-22
Separate issuer in play/pause during seeking

bitmovin-kenny:
Analytics could not differentiate between user initiated play/pause calls and play/pause called which were issued due to seeking while the player was playing.

Now the play/pause calls which happen during seeking have a different issuer `ui-seek`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This major release is adjusted to the changed API of Bitmovin Player v8.
 
 ### Changed
 - Player type definitions updated to player v8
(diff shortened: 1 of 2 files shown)


Tigraine:
Are we sure this doesn't break things? What if a customer looks at the issuer?

bitmovin-kenny:
The issuer was introduced to differentiate between internal actions of the player and actions triggered by the user via the API.
The play/pause with seeking is not directly triggerd by the user, but is a UI specific handling so I guess it will be ok.

dweinber:
And also this is for UI v3.0.0, so it's fine to change that.

Tigraine:
Then I approve this message ☑️

PR #146, opened 2018-08-23
Recommendation overlay not shown

stonko1994:
## Description

I recognized that the recommendation-overlay was not shown when playback finished.

## Fix

* update the UI to use the new AdsAPI
* update ad samples

## Discussion

@Reevn two open questions:
* Is it possible to set `skipMessage` and `adMessage`?
* The `skipAfter` is currently not working with the IMA module?

### DIFF
--- a/src/ts/components/recommendationoverlay.ts
+++ b/src/ts/components/recommendationoverlay.ts
@@ -67,7 +67,8 @@ export class RecommendationOverlay extends Container<ContainerConfig> {
     player.on(player.exports.Event.PlaybackFinished, () => {
       // Dismiss ON_PLAYBACK_FINISHED events at the end of ads
       // TODO remove this workaround once issue #1278 is solved
-      if (player.isAd()) {
+      if (player.ads.isLinearAdActive()) {
+        console.log('was ad');
         return;
       }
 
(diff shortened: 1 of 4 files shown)


src/ts/components/recommendationoverlay.ts
```diff
@@ -67,7 +67,7 @@ export class RecommendationOverlay extends Container<ContainerConfig> {
     player.on(player.exports.Event.PlaybackFinished, () => {
       // Dismiss ON_PLAYBACK_FINISHED events at the end of ads
       // TODO remove this workaround once issue #1278 is solved
```
stonko1994: How do i find issue #1278?

protyposis: It doesn't exist any more. In v7, the player fired `ON_PLAYBACK_FINISHED` *before* a postroll ad would play, hence this workaround. This is not the case any longer with the new ads module in v8 so you can just remove this `if` block.

Reevn:
@stonko1994 Regarding your questions:
- `skipMessage` and `adMessage` is currently not available in the new Advertising API in v8
- Yes, `skipAfter` is not available for the IMA Advertising Module

PR #148, opened 2018-08-24
Update error message handling

stonko1994:
## Description

Issue: The `errorEvent` has changed in player version v8. There is no longer a readable message in it so we had to add it.

## Fix

* Add default error descriptions for error codes
* Pass those description to the callback if customizing messages is used.


src/scss/skin-modern/components/_errormessageoverlay.scss
```diff
@@ -18,11 +18,12 @@
...
     text-align: center;
     user-select: text;
+    white-space: pre-line; // enable linebreak in text
 
-    // Vertically center the label (not exactly, but works for single line errors)
+    // Vertically center the label
     & {
-      margin-top: -1em;
       top: 50%;
+      transform: translateY(-50%);
```
protyposis: Does this work in IE9 & IE11?

src/ts/components/errormessageoverlay.ts
```diff
@@ -4,9 +4,10 @@ import {UIInstanceManager} from '../uimanager';
 import ErrorEvent = bitmovin.PlayerAPI.ErrorEvent;
 import {TvNoiseCanvas} from './tvnoisecanvas';
 import PlayerEvent = bitmovin.PlayerAPI.PlayerEvent;
+import {ErrorEventUtil} from '../erroreventutil';
 
 export interface ErrorMessageTranslator {
-  (error: ErrorEvent): string;
+  (description: string, error: ErrorEvent): string;
```
protyposis: The event is more important so it should be the first parameter. Also the description is optional because there's not a description for every code. Actually I think this is not a nice design as the `ErrorMessageTranslator` should output the "description" and not receive it as input. Also the name `description` seems a little off, `errorMessage` would be more fitting.

src/ts/components/errormessageoverlay.ts
```diff
@@ -99,22 +101,25 @@ export class ErrorMessageOverlay extends Container<ErrorMessageOverlayConfig> {
...
-      // Process message translations
+      let name = event.name;
+      let message = description + '\n(' + name + ')'; // default error message style
+
+      // Custom error message handling
       if (config.messages) {
         if (typeof config.messages === 'function') {
-          // Translation function for all errors
-          message = config.messages(event);
+          // Custom function for all errors
```
protyposis: Why did you change the wording of the comments around here? The old ones are still more fitting.

src/ts/erroreventutil.ts
```diff
@@ -0,0 +1,61 @@
...
+    1401: 'The manifest download timed out',
+    1403: 'The progressive stream download timed out',
+    1402: 'The segment download timed out',
+    1301: 'General error if Flash renderer has an error',
+    1302: 'Flash doesn\'t have sufficient resources',
+    1303: 'The transmuxer could not be initialized',
+    1300: 'Video decoder or demuxer had an error with the content',
+    1112: 'Flash container API not available',
+    1110: 'Native Flash is not authorized by a valid Adobe token',
+    1111: 'Flash doesn\'t have sufficient resources',
```
protyposis: Same text as 1302?

src/ts/components/errormessageoverlay.ts
```diff
@@ -99,22 +101,25 @@ export class ErrorMessageOverlay extends Container<ErrorMessageOverlayConfig> {
     let config = <ErrorMessageOverlayConfig>this.getConfig();
 
     player.on(player.exports.Event.Error, (event: ErrorEvent) => {
-      let message = event.message;
+      let description = ErrorEventUtil.defaultErrorMessagesMapping[event.code] as string;
 
-      // Process message translations
+      let name = event.name;
+      let message = description + '\n(' + name + ')'; // default error message style
```
protyposis: The message should be, by priority:
 1. The output of an externally supplied translator
 2. The output if the internal default translator (`defaultErrorMessagesMapping`)
 3. `${error.code} ${error.name}`

The `ErrorMessageTranslator` shall not receive a `message`, only output one.

protyposis:
Please also extend the `UIConfig` to allow error translations:

```ts
interface UIConfig {
  ...
}
```

src/scss/skin-modern/components/_errormessageoverlay.scss
stonko1994: afaik yes but i will test it in a sec.

src/ts/components/errormessageoverlay.ts
stonko1994: Yes but we where thinking about using the callback (`ErrorMessageTranslator`) to 'format' the description without the need to define an own.

src/ts/components/errormessageoverlay.ts
stonko1994: translation was confusing me because the meaning off translating to a different language.

src/ts/erroreventutil.ts
stonko1994: Texts are extracted from the [docs](http://cdn.bitmovin.com/player/web/staging/8/docs/enums/errors.playererrorcode.html)

src/ts/components/errormessageoverlay.ts
stonko1994: Hm maybe possible when accessing the `defaultErrorMessagesMapping` within the callback by yourself.

src/ts/components/errormessageoverlay.ts
protyposis: "Translation" doesn't have to refer to different spoken languages. You can also translate an error code/event to an error message.

protyposis: Also the basic idea was indeed to allow translations into different languages.

src/ts/components/errormessageoverlay.ts
protyposis: Exactly. Implement is as an `ErrorMessageTranslator`, within it you can do whatever you want, i.e. processing placeholders. Then use it as fallback as described in the comment below.

src/scss/skin-modern/components/_errormessageoverlay.scss
stonko1994: It is working find in IE11. Unfortunately i'm currently not able to test it in IE9 because the player v8 is currently not working there.

protyposis:
The pattern for all configuration is the same as with all other comoonents: all configuration is offered by the components themselves. Some config options are added to the UIConfig for convenience, and they always take precendence over the config at the component.

src/scss/skin-modern/components/_errormessageoverlay.scss
protyposis: You can test the CSS transform by itself.

src/scss/skin-modern/components/_errormessageoverlay.scss
```diff
@@ -18,11 +18,14 @@
...
     text-align: center;
     user-select: text;
+    white-space: pre-line; // enable linebreak in text
 
-    // Vertically center the label (not exactly, but works for single line errors)
+    // Vertically center the label
     & {
-      margin-top: -1em;
+      // sass-lint:disable no-vendor-prefixes
+      -ms-transform: translateY(-50%); // required for IE9
```
protyposis: Not necessary. This is automatically added by the `autoprefixer` during the build.

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Added
+- Default error description for existing error codes
+
+### Changed
+- The `ErrorMessageTranslator` receives a `description` alongside the `ErrorEvent` object
```
protyposis: Remove

src/ts/components/errormessageoverlay.ts
```diff
@@ -99,16 +102,18 @@ export class ErrorMessageOverlay extends Container<ErrorMessageOverlayConfig> {
     let config = <ErrorMessageOverlayConfig>this.getConfig();
 
     player.on(player.exports.Event.Error, (event: ErrorEvent) => {
-      let message = event.message;
+      let message = ErrorEventUtil.defaultErrorMessageTranslator(event);
 
+      // errorMessages configured in `UIConfig` take precedence `ErrorMessageOverlayConfig`
+      let customErrorMessages = uimanager.getConfig().errorMessages || config.messages;
```
protyposis: Call it `errorMessages`. They are only "custom" when provided via the config, not by default.

src/ts/erroreventutil.ts
```diff
@@ -0,0 +1,68 @@
+import {ErrorMessageMap, ErrorMessageTranslator} from './components/errormessageoverlay';
+import ErrorEvent = bitmovin.PlayerAPI.ErrorEvent;
+
+export namespace ErrorEventUtil {
+  export const defaultErrorMessagesMapping: ErrorMessageMap = {
```
protyposis: Call it `defaultErrorMessages`. The type specifies that it is a mapping.

src/ts/erroreventutil.ts
```diff
@@ -0,0 +1,68 @@
...
+    1208: 'Progressive stream type not supported or the stream has an error',
+    1203: 'The stream type is not supported',
+    1000: 'Error is unknown',
+    2100: 'Player technology not compatible with VR playback',
+  };
+
+  export const defaultErrorMessageTranslator: ErrorMessageTranslator = (error: ErrorEvent) => {
+    const errorMessage = ErrorEventUtil.defaultErrorMessagesMapping[error.code];
+    const errorIdentifier = error.name;
+    return `${errorMessage}\n(${errorIdentifier})`; // default error message style
```
protyposis: Add a default code/name fallback in case no default message is defined.

src/scss/skin-modern/components/_errormessageoverlay.scss
stonko1994: I checked this but `-ms` was not added for the `transform` property.

stonko1994: @protyposis our config says `Last 2 versions` and IE9 is not included there so it will not be added in this case.

protyposis: Then we might have to update the config. 2 versions seem a bit low anyway given the fast release cycle of Chrome & FF. Can you check what the currently recommended config is?

dweinber: I think it's fair to drop IE9 support in UI v3 / Player v8 more than 2.5 years after Microsoft stopped supporting it: https://www.microsoft.com/en-us/windowsforbusiness/end-of-ie-support

src/ts/erroreventutil.ts
```diff
@@ -0,0 +1,71 @@
...
+    1208: 'Progressive stream type not supported or the stream has an error',
+    1203: 'The stream type is not supported',
+    1000: 'Error is unknown',
+    2100: 'Player technology not compatible with VR playback',
+  };
+
+  export const defaultErrorMessageTranslator: ErrorMessageTranslator = (error: ErrorEvent) => {
+    const errorMessage = ErrorEventUtil.defaultErrorMessages[error.code] || unknownErrorMessage;
+    const errorIdentifier = error.name || unknownErrorName ;
+    return `${errorMessage}\n(${errorIdentifier})`; // default error message style
```
protyposis:
```ts
const errorMessage = ErrorEventUtil.defaultErrorMessages[error.code];

if (errorMessage) {
  // Use the error message text if there is one
  return errorMessage;
} else {
  // Fallback to error code/name if no message is defined
  return `${error.code} ${error.name}`;
```

src/scss/skin-modern/components/_errormessageoverlay.scss
protyposis: Even better 👍

protyposis:
Please also add an export of `ErrorEventUtil` to the main export file so it can be used by external error message translator implementations.

protyposis:
Another thing: all other `Utils` are in plural. Also it should not be unnecessarily constrained to the `ErrorEvent` so better call it `ErrorUtils`.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Added
+- Default error description for existing error codes
```
```md
### Added
- Default `ErrorMessageTranslator` with english error messages for the `ErrorMessageOverlay`
- `UIConfig.errorMessages` to allow customization of error messages via custom `ErrorMessageTranslator`\`ErrorMessageMap`
```

src/ts/uimanager.ts
```diff
@@ -114,6 +114,11 @@ export interface UIConfig {
    * Default: false
    */
   playbackSpeedSelectionEnabled?: boolean;
+  /**
+   * Provide customized errorMessages
+   * For an example have a look at {@link ErrorMessageOverlay}
```
protyposis: `{@link ErrorMessageOverlayConfig.messages}`

unki2aut:
@stonko1994 please be aware that because of the introduction of general error codes the codes you used will change with version `v8.0.0-rc2` of the player.

Issue #149, opened 2018-08-24
Extract UIManager.Factory into separate file (v3)


protyposis:
Move it into `uifactory.ts`,

Might also be a good idea to move `UIConfig`+`UIRecommendationConfig`+`TimelineMarker` into `uiconfig.ts`.

PR #156, opened 2018-09-03
Extract factory and config interfaces to separete files

protyposis:
https://github.com/bitmovin/bitmovin-player-ui/issues/149


src/ts/uiconfig.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Specifies the container in the DOM into which the UI will be added. Can be a CSS selector string or a
+   * HTMLElement object. By default, the player figure will be used ({@link PlayerAPI#getFigure}).
+   */
+  container?: string | HTMLElement;
+  metadata?: {
+    title?: string;
+    description?: string;
+    markers?: TimelineMarker[];
+  };
+  // TODO move recommendations into metadata in next major release
```
stonko1994: v3 would be a major release. Should we do this now (separate PR)?

src/ts/uimanager.ts
```diff
@@ -573,451 +465,12 @@ export class UIManager {
...
-  }
-
-  export function buildLegacyTestUI(player: PlayerAPI, config: UIConfig = {}): UIManager {
-    return new UIManager(player, legacyTestUI(), config);
-  }
+/**
+ * Export the {@link UIFactory} as `UIManager.Factory` for backwards compatibility with the UI module in player v8 RCs.
+ * @deprecated remove with UI v4
+ */
+export namespace UIManager {
```
stonko1994: The player v8 RCs do have their suitable UI version bundled. 
Do we really need this backwards compatibility for RCs since there was no stable release of the player or the UI?

src/ts/uiconfig.ts
protyposis: No I left this open on purpose... it's not important

src/ts/uimanager.ts
protyposis: For now we need it, else it doesn't work with the player. We can adjust the UI module with the next player RC and then remove this, but since an RC must have a stable API I went with this approach. Or course one can argue that this is not related to the player API and therefore a valid change.

protyposis:
> Should we unify the code styling regarding spaces after curly brackets (`{ ... }`)? maybe at least for import statements?

We are going to do this as soon as we have a common style for all open source projects.

PR #161, opened 2018-09-04
Remove legacy skin

stonko1994:


src/ts/uifactory.ts
```diff
@@ -41,7 +41,6 @@ import { MetadataLabel, MetadataLabelContent } from './components/metadatalabel'
 import { PlayerUtils } from './playerutils';
 import { Label } from './components/label';
 import { CastUIContainer } from './components/castuicontainer';
-import { VolumeControlButton } from './components/volumecontrolbutton';
```
stonko1994: @protyposis can we remove the `VolumeControlButton` at all?

protyposis: Let's keep it, maybe somebody can make use of it.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Removed
 - `SubtitleSettingsPanel` in favour of `SubtitleSettingsPanelPage` to use with the new navigation feature of `SettingsPanel`
+- All variants of `skin-legacy`
```
protyposis: `` Player v6 legacy skin (`skin-legacy`) and its UI variants``

src/scss/skin-modern/_skin.scss
```diff
@@ -1,94 +1,93 @@
...
   user-select: none;
+}
 
-  &.#{$prefix}-player-state-idle {
-    .#{$prefix}-ui-controlbar,
-    .#{$prefix}-ui-titlebar,
-    .#{$prefix}-ui-hugeplaybacktogglebutton {
-      display: none;
-    }
+&.#{$prefix}-player-state-idle {
```
protyposis: I think you need to nest all into `.#{$prefix}-ui-skin-modern {`, else the `&` extender does not really work as there is no parent?

stonko1994: FYI: It actually behaves like the `&` isn't there. 

It looks a bit weird so I will change it anyway.

Issue #163, opened 2018-09-05
Timeline markers are not updating after switching source


stonko1994:
When you load a source with `timeline-markers` (`source.markers`) first and afterwards loading a source **without** timeline-markers the old ones are still there.

Similar could happen to `recommendations` `title` & `description`.

Affected Versions: 2.18.x / 3.x

EDIT:
This happened because the `uiConfig` was not copied but merged with the config passed to the player.

protyposis:
Fixed in `3.0.0`

PR #165, opened 2018-09-05
Timeline markers update

stonko1994:
Issue: #163 

This PR also includes a fix where the playback position marker was jumping from 0 to 100 when switching to a live-stream after first watching a VoD.


src/ts/uimanager.ts
```diff
@@ -127,7 +128,8 @@ export class UIManager {
     }
 
     this.player = player;
-    this.config = {
+    this.uiconfig = config;
+    this.mergedConfig = {
```
protyposis: Why were these changes necessary? It seems to me as if the only thing necessary was to fix the markers update in `updateConfig`?

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Type definitions are now generated correctly and referenced from `package.json`
+- Remained timeline markers from previous source
```
protyposis: `Update timeline markers when loading a new source`

src/ts/uimanager.ts
```diff
@@ -439,15 +438,15 @@ export class UIManager {
    * Returns the list of all added markers in undefined order.
    */
   getTimelineMarkers(): TimelineMarker[] {
-    return this.config.metadata.markers;
+    return this.uiconfig.metadata.markers;
```
protyposis: Should be `mergedConfig`?

src/ts/uimanager.ts
```diff
@@ -456,8 +455,8 @@ export class UIManager {
    * be found and thus not removed.
    */
   removeTimelineMarker(timelineMarker: TimelineMarker): boolean {
-    if (ArrayUtils.remove(this.config.metadata.markers, timelineMarker) === timelineMarker) {
-      this.config.events.onUpdated.dispatch(this);
+    if (ArrayUtils.remove(this.uiconfig.metadata.markers, timelineMarker) === timelineMarker) {
```
protyposis: Again, `mergedConfig`? I don't think the unprocessed `uiconfig` should evner be used, that's why I originally didnt even store it in a field. Only the processed config should be used.

src/ts/uimanager.ts
stonko1994: To clarify what the config is actually and to facilitate the handling when the config is updated.

protyposis: Still the `uiconfig` field is never used except in the constructor which has access to the `config` parameter anyway, hence the field is unnecessary.

PR #169, opened 2018-09-07
Fix misplaced UI

stonko1994:
Issue: We found out that the entirety UI was misplaced when no bootstrap is on the side. So we added the default `box-sizing` from bootstrap to all elements within the `ui-container`


stonko1994:
@protyposis you were right with the assumption that the `box-sizing` attribute from bootstrap is 'fixing' this issue.

PR #171, opened 2018-09-11
Seek bar marker positioning

stonko1994:


src/ts/components/seekbar.ts
```diff
@@ -765,18 +769,19 @@ export class SeekBar extends Component<SeekBarConfig> {
     if (this.config.vertical) {
       px = this.seekBar.height() - px - this.seekBarPlaybackPositionMarker.height();
     }
+    const propertyValueString = 'calc(' + SeekBar.TRANSLATE_OFFSET + '% + ' + px + 'px)';
```
protyposis: `calc` won't work in old Android WebViews so we shouldn't use it. Is this really necessary? Why can't we calculate in JS like before?

src/ts/components/seekbar.ts
```diff
@@ -65,6 +65,10 @@ export class SeekBar extends Component<SeekBarConfig> {
    * The CSS class that is added to the DOM element while the seek bar is in 'seeking' state.
    */
   private static readonly CLASS_SEEKING = 'seeking';
+  /**
+   * Offset for the translateX/translateY value applied when scrubbing in %
+   */
+  private static TRANSLATE_OFFSET = -50;
```
protyposis: This value is only used once and the description is not really explaining anything as there's no added value in creating this field... better inline it directly.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Changed
+- Positioning of `SeekBar` markers improved
```
protyposis: `SeekBar`/`VolumeSlider`

src/ts/components/seekbar.ts
stonko1994: It's possible in JS but i thought the support for `calc` is good enough. 

I saw that calc is supported since > 4.4.4 but 4.4 and 4.4.4 has < 1% and our `autoprefixer` setting is >1% so i assumed that it is ok to use calc.

protyposis: Sounds reasonable, ok then let's use `calc` and see what happens :) https://caniuse.com/#search=calc says that 4.4 and 4.4.4 partially support `calc` and are only missing the ability to multiply and divide... since we're only adding here, this might even work.

protyposis: Actually, if you revert the `transform` to `left` in `_seekbar.scss` this becomes unnecessary and can be removed :)

PR #178, opened 2018-09-20
Improving hit-boxes

stonko1994:
## Description

Our buttons where styled with margins that means that there are dead space between them. This lead to sometimes hard hittable buttons especially on an mobile device. This pull request improve this by working with paddings.

**Please have a look also on the UI and test it (click through everything)**

## Discussion
I hope this doesn't change to much for a minor release?

## TODO
* Issue in live-stream UI

### DIFF
--- a/src/scss/skin-modern/_skin.scss
+++ b/src/scss/skin-modern/_skin.scss
@@ -55,6 +55,7 @@
 
   // align -ui-playbacktimelabel with our play button in the default ui variant
   // this does not apply for all buttons / elements but it looks weird without
+  // sass-lint:disable force-pseudo-nesting nesting-depth
   &:not(.#{$prefix}-ui-skin-cast-receiver):not(.#{$prefix}-ui-skin-smallscreen) {
     .#{$prefix}-ui-controlbar {
       .#{$prefix}-ui-container > {
--- a/src/scss/skin-modern/_skin.scss
+++ b/src/scss/skin-modern/_skin.scss
@@ -53,6 +53,20 @@
   text-align: left;
   user-select: none;
 
+  // align -ui-playbacktimelabel with our play button in the default ui variant
+  // this does not apply for all buttons / elements but it looks weird without
+  &:not(.#{$prefix}-ui-skin-cast-receiver):not(.#{$prefix}-ui-skin-smallscreen) {
+    .#{$prefix}-ui-controlbar {
+      .#{$prefix}-ui-container > {
+        .#{$prefix}-container-wrapper > * {
+          &.#{$prefix}-ui-playbacktimelabel {
+            margin-left: 5px;
+          }
+        }
+      }
+    }
+  }
+
   // Include padding and border in the element's total width and height
   *,
   *::before,
--- a/src/scss/skin-modern/components/_airplaytogglebutton.scss
+++ b/src/scss/skin-modern/components/_airplaytogglebutton.scss
@@ -4,17 +4,13 @@
 .#{$prefix}-ui-airplaytogglebutton {
   @extend %ui-button;
 
-  &::before {
-    background-image: svg('assets/skin-modern/images/airplay.svg');
-  }
+  background-image: svg('assets/skin-modern/images/airplay.svg');
 
   &:hover {
     @include svg-icon-shadow;
   }
 
   &.#{$prefix}-on {
-    &::before {
-      background-image: svg('assets/skin-modern/images/airplayX.svg');
-    }
+    background-image: svg('assets/skin-modern/images/airplayX.svg');
   }
 }
--- a/src/scss/skin-modern/components/_airplaytogglebutton.scss
+++ b/src/scss/skin-modern/components/_airplaytogglebutton.scss
@@ -4,13 +4,17 @@
 .#{$prefix}-ui-airplaytogglebutton {
   @extend %ui-button;
 
-  background-image: svg('assets/skin-modern/images/airplay.svg');
+  &::before {
+    background-image: svg('assets/skin-modern/images/airplay.svg');
+  }
 
   &:hover {
     @include svg-icon-shadow;
   }
 
   &.#{$prefix}-on {
-    background-image: svg('assets/skin-modern/images/airplayX.svg');
+    &::before {
+      background-image: svg('assets/skin-modern/images/airplayX.svg');
+    }
   }
 }
--- a/src/scss/skin-modern/components/_button.scss
+++ b/src/scss/skin-modern/components/_button.scss
@@ -4,6 +4,7 @@
   @extend %ui-component;
 
   background-color: transparent;
+  background-origin: content-box;
   background-position: center;
   background-repeat: no-repeat;
   border: 0;
--- a/src/scss/skin-modern/components/_button.scss
+++ b/src/scss/skin-modern/components/_button.scss
@@ -24,7 +24,6 @@
   &::before {
     content: '';
     vertical-align: middle;
-    color: black;
     position: relative;
     display: inline-block;
     background-repeat: no-repeat;
--- a/src/scss/skin-modern/components/_button.scss
+++ b/src/scss/skin-modern/components/_button.scss
@@ -11,7 +11,8 @@
   font-size: 1em;
   height: 1.5em;
   min-width: 1.5em;
-  padding: 0;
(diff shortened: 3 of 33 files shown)


src/scss/skin-modern/components/_button.scss
```diff
@@ -7,18 +7,31 @@
   background-position: center;
   background-repeat: no-repeat;
   border: 0;
+  // sass-lint:disable no-important
+  box-sizing: content-box !important;
```
stonko1994: The 'global' box-sizing selector 
```scss
  *,
  *::before,
  *::after {
    box-sizing: border-box;
  }
```

in skin.scss has precedence so over this selector so important is needed here. A possible solution would be to move the box-sizing (in skin.scss) outside of the `.#{$prefix}-ui-uicontainer` but this would affect the styling on the entire page where the UI is included. Bootstrap does it too but i don't think it is a perfect solution.

What do you think @protyposis?

src/scss/skin-modern/components/_controlbar.scss
```diff
@@ -8,38 +8,46 @@
...
   }
 
-  .#{$prefix}-controlbar-top {
-    .#{$prefix}-ui-label {
-      font-size: .9em;
-    }
+  .#{$prefix}-ui-container > {
+    .#{$prefix}-container-wrapper > * {
+      &:first-child {
+        padding-left: 2rem;
```
stonko1994: I just found out that this is an issue if the first issue is `hidden`. (Happens in livestreams without `timeShift`).

A solution would be something like this:
```scss
&:not(.#{$prefix}-hidden) {
  padding-left: 2rem;
}
&:not(.#{$prefix}-hidden) ~ :not(.#{$prefix}-hidden) {
  padding-left: $default-button-badding;
}
```

But i feels so hacky that i don't want this. Again do you have any other idea @protyposis. I was thinking about detecting the first visible item in container JS and add another class but this also feels not perfect.

src/scss/skin-modern/_skin.scss
```diff
@@ -53,6 +53,21 @@
   text-align: left;
   user-select: none;
 
+  // align -ui-playbacktimelabel with our play button in the default ui variant
+  // this does not apply for all buttons / elements but it looks weird without
+  // sass-lint:disable force-pseudo-nesting nesting-depth
+  &:not(.#{$prefix}-ui-skin-cast-receiver):not(.#{$prefix}-ui-skin-smallscreen) {
```
protyposis: This really doesn't look pretty 😨

src/scss/skin-modern/components/_airplaytogglebutton.scss
```diff
@@ -4,13 +4,17 @@
 .#{$prefix}-ui-airplaytogglebutton {
   @extend %ui-button;
 
-  background-image: svg('assets/skin-modern/images/airplay.svg');
+  &::before {
```
protyposis: As discussed offline, it would be nice if the images could remain as is and we insert a dummy element in `::before` for correct sizing.

src/scss/skin-modern/components/_button.scss
protyposis: We need to get rid of the `!important`, but let's postpone this for later. Maybe we get an idea in the meantime on how to solve this.

src/scss/skin-modern/components/_label.scss
```diff
@@ -6,6 +6,7 @@
   @include hidden;
 
   cursor: default;
+  padding: .25em;
```
protyposis: Do you think it would be possible to scale the padding via a global skin variable (in `_variables.scss`)? E.g. `$padding: .25em !default;` or `$padding-scale: 1 !default;`.

src/scss/skin-modern/components/_listbox.scss
```diff
@@ -9,10 +9,12 @@
 
     display: block;
     font-size: .8em;
-    height: 100%;
     min-width: 10em;
     padding: .5em;
-    width: 100%;
+
+    &::before {
+      display: none;
```
protyposis: Why is this necessary?

src/ts/components/playbacktimelabel.ts
```diff
@@ -28,7 +28,7 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
     super(config);
 
     this.config = this.mergeConfig(config, <PlaybackTimeLabelConfig>{
-      cssClass: 'ui-playbacktimelabel',
+      cssClasses: ['ui-playbacktimelabel'],
```
protyposis: This should be `cssClass`

src/ts/components/titlebar.ts
```diff
@@ -27,8 +28,16 @@ export class TitleBar extends Container<TitleBarConfig> {
       cssClass: 'ui-titlebar',
       hidden: true,
       components: [
-        new MetadataLabel({ content: MetadataLabelContent.Title }),
-        new MetadataLabel({ content: MetadataLabelContent.Description }),
+        new Container({
```
protyposis: The `TitleBar` is already a `Container`, why is a `Container` within a `Container` necessary? This looks like a hack.

src/ts/components/titlebar.ts
```diff
@@ -86,4 +95,20 @@ export class TitleBar extends Container<TitleBarConfig> {
     // init
     checkMetadataTextAndUpdateVisibility();
   }
+
+  /**
+   * Collect all Leaves Components within the Container structure (Needed for keepHiddenWithoutMetadata)
+   * TODO: move to container if needed elsewhere
+   */
+  private static getLeavesComponents(parentComponent: Container<ContainerConfig>): Component<ComponentConfig>[] {
```
protyposis: `getLeafComponents`

src/ts/components/titlebar.ts
```diff
@@ -86,4 +95,20 @@ export class TitleBar extends Container<TitleBarConfig> {
     // init
     checkMetadataTextAndUpdateVisibility();
   }
+
+  /**
+   * Collect all Leaves Components within the Container structure (Needed for keepHiddenWithoutMetadata)
```
protyposis: `Collect all leaves from the component tree...`

src/ts/uifactory.ts
```diff
@@ -238,14 +237,18 @@ export namespace UIFactory {
...
           components: [
-            new MetadataLabel({ content: MetadataLabelContent.Title }),
-            new CastToggleButton(),
-            new VRToggleButton(),
-            new PictureInPictureToggleButton(),
-            new AirPlayToggleButton(),
-            new VolumeToggleButton(),
-            new SettingsToggleButton({ settingsPanel: settingsPanel }),
-            new FullscreenToggleButton(),
+            new Container({
```
protyposis: Again `Container` within `Container` - looks like a hack. There should be a cleaner solution. What is the actual issue why this is needed?

src/scss/skin-modern/components/_label.scss
stonko1994: Hm, it would increase complexity to positioning the `seekbar` / `volumeslider`.

src/scss/skin-modern/components/_listbox.scss
stonko1994: Because the `::before` is blocking content. For buttons without images this would mean that the text is no longer centered.

src/ts/components/titlebar.ts
stonko1994: This allowed me to apply css similar to the `controlbar` also for the `titlebar`. 
This `css` expect the first level of `.#{$prefix}-ui-container` to represent a 'row' in the UI. Maybe it's no good approach.

src/scss/skin-modern/components/_controlbar.scss
```diff
@@ -8,38 +8,46 @@
...
+      }
 
-    > .#{$prefix}-container-wrapper > * {
-      margin: 0 .5em;
+      &:last-child {
+        padding-right: 2rem;
```
protyposis: Is it a good idea to mix `rem` with `em`?

src/scss/skin-modern/components/_button.scss
protyposis: Solved in https://github.com/bitmovin/bitmovin-player-ui/pull/181

PR #179, opened 2018-09-27
Fix horizontal seek bar and playback position marker alignment

stonko1994:
## Idea
The idea is to calculate all dimensions within the seekbar based on one value (bar-height).
For the position-maker multiply with `3` then there is the same space above and bottom (position center with `margin: auto`). When you would like to adjust the size more exact add / subtract an even number to it to always keep the same. (Same applies for `seek-bar-markers`) This requires to replace the background images with css styling.

bildschirmfoto 2018-09-27 um 13 58 01

@protyposis what is your opinion on this change? Shouldn't this be at least similar in all browsers on all devices?

### DIFF
--- a/src/scss/skin-modern/_variables.scss
+++ b/src/scss/skin-modern/_variables.scss
@@ -20,3 +20,5 @@ $subtitle-text-border-color: #000 !default;
 
 $animation-duration: .3s !default;
 $animation-duration-short: $animation-duration / 2 !default;
+
+$seekbar-height: 5px;
--- a/src/scss/skin-modern/_variables.scss
+++ b/src/scss/skin-modern/_variables.scss
@@ -20,5 +20,3 @@ $subtitle-text-border-color: #000 !default;
 
 $animation-duration: .3s !default;
 $animation-duration-short: $animation-duration / 2 !default;
-
-$seekbar-height: 0.3125em;
--- a/src/scss/skin-modern/_variables.scss
+++ b/src/scss/skin-modern/_variables.scss
@@ -21,4 +21,4 @@ $subtitle-text-border-color: #000 !default;
 $animation-duration: .3s !default;
 $animation-duration-short: $animation-duration / 2 !default;
 
-$seekbar-height: 5px;
+$seekbar-height: 0.3125em;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
+### Changed
+- `Seekbar`/`VolumeSlider` position markers changed from SVG to pure CSS to improve vertical alignment with bar
+
 ### Fixed
 - Uncaught `PlayerAPINotAvailableError` in `SeekBar` position updater when player is destroyed
 - Unresponsive UI when a user canceled connection establishment to a Cast receiver
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -79,7 +79,7 @@
     </div>
 </div>
 
-<script src="https://bitmovin-a.akamaihd.net/bitmovin-player/staging/8/bitmovinplayer.js"></script>
+<script src="https://cdn.bitmovin.com/player/web/8/bitmovinplayer.js"></script>
 <script src="js/bitmovinplayer-ui.js"></script>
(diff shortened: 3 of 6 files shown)


protyposis:
Great idea! Looks very promising 🙂 Will do some more tests tomorrow.

protyposis:
On highres displays this seems to work perfectly.

On a lowres display (Lenovo T420s), the volume marker seems a litle bit off to the bottom
image 1, but it isn't: 
image So this is fine.

On a MacBook Air with Safari 10.1 and MacOs 10.12.4, the seekbar marker is rendered weirdly: 
screen shot 2018-09-28 at 09 22 52.

There's one issue though which I'll bring up in the review that follows.

src/scss/skin-modern/_variables.scss
```diff
@@ -20,3 +20,5 @@ $subtitle-text-border-color: #000 !default;
 
 $animation-duration: .3s !default;
 $animation-duration-short: $animation-duration / 2 !default;
+
+$seekbar-height: 5px;
```
protyposis: Working with pixel values is a problem because the volume and seekbars don't scale with the UI, e.g. when setting `font-size: 2em` on the container to scale the UI to 200%.

stonko1994: True but when we would change this to `em`, i guess then we would have the old issue again because we wont get the pixel-perfect design for the seek-bar. 

For example: 
Currently the size of the position marker is `$seekbar-height * 3 - 4px`. It's `4px` because it's preferred to be an even number (why? described in the PR description). 

`4px` -> `.25em` with `16px` default size. When changing the size to `$seekbar-height * 3 - .25em` everything looks fine.

But when changing the default size to `20px` then `.25em` will result in -> `5px`. And it looks weird again.

protyposis: So are you saying that this purely CSS-based approach will not yield any advantages over the current SVG-based approach if we don't use absolute pixel sizes? (And in turn we could use pixel sizes with the SVG approach to achieve the same result?)

stonko1994: True. What if we calculate the `$seekbar-height` based on `em` and convert it to `px`?

```scss
@function em2px($target, $context: $font-size) {
  @if $target == 0 { @return 0 }
  @return ($target / 1em) * ($context / 1em) * 16 + 0px;
}

$seekbar-height: em2px(.3125em);
```

This would support the UI scale with changing the `$font-size`.

protyposis: I don't think it's possible to know the font-size at compile time. This would have to be implemented as a native CSS calculation?

stonko1994: That's also true this will only support our `$font-size` variable.

protyposis: Move this into `_seekbar.scss`, it's not a global setting

stonko1994: i need it also in the `_volumeslider.scss` so either declaring it two times or moving somewhere else. Is there a way to 'inherit' it?

stonko1994:
Changed back to `em` but please react to [this](
https://github.com/bitmovin/bitmovin-player-ui/pull/179#discussion_r222214451) comment

src/scss/skin-modern/_variables.scss
protyposis: Declare it at the top of `_seekbar.scss` (below the imports) and add an `@import '_seekbar';` to `_volumeslider.scss`

src/scss/skin-modern/_variables.scss
```diff
@@ -20,3 +20,5 @@ $subtitle-text-border-color: #000 !default;
 
 $animation-duration: .3s !default;
 $animation-duration-short: $animation-duration / 2 !default;
+
+$seekbar-height: 0.3125em;
```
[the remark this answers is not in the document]
stonko1994: Those values are used to keep a version which is most similar to the above elaborated idea.
The default `font-size` is 16px so `0.3125em` will calculate to 5px. All `em` values here will calculate to the `px` values used before.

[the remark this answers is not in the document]
protyposis: Ok then let's keep it like this and see if we get an improvement. I was just concerned about unnecessary changes.

PR #180, opened 2018-09-27
Replace margins with paddings to improve hit boxes

stonko1994:
## Description
This is extracted from #178 as **part 1**.

We deiced do split #178 into two parts: first will be to just replace margins with paddings to have a slight improvement for the hit boxes. Part two would be to use available space until the edge of the UI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Avoid unnecessary animation when `BufferingOverlay` is hidden
 - Avoid unnecessary DOM modification when the text of a `Label` does not change
 
+### Changed
+- Margin to padding to improve hit-boxes
+
 ## [3.0.1]
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Avoid unnecessary DOM modification when the text of a `Label` does not change
 
 ### Changed
-- Margin to padding to improve hit-boxes
+- Improved `Button` hit-boxes by changing margins to paddings
 
 ## [3.0.1]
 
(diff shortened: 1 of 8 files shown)


src/scss/skin-modern/components/_button.scss
```diff
@@ -6,12 +6,13 @@
...
   background-position: center;
   background-repeat: no-repeat;
+  background-size: 1.5em;
   border: 0;
   cursor: pointer;
   font-size: 1em;
   height: 1.5em;
   min-width: 1.5em;
-  padding: 0;
+  padding: 1em;
```
protyposis: Why did you take this approach now, instead of 

```
background-origin: content-box;
padding: .25em;
box-sizing: content-box;
```
like in #178?

stonko1994: Because it's not needed when setting the background-size. (And it's not working without more changes)

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Avoid unnecessary animation when `BufferingOverlay` is hidden
 - Avoid unnecessary DOM modification when the text of a `Label` does not change
 
+### Changed
+- Margin to padding to improve hit-boxes
```
protyposis: ``Improved `Button` hit-boxes by changing margins to paddings``

src/scss/skin-modern/components/_button.scss
protyposis: This might not work with irregular paddings, e.g. a larger padding on the bottom of a button to fill unused space... the background image will always be centered, `background-origin: content-box` seems not to work, and the background needs to be explicitely positioned with `background-position: x y`.

stonko1994: That was the case for `box-sizing: content-box !important;`. Then `background-origin` will work. So let's merge #181 first, then i can change it.

protyposis: Merged.

stonko1994:
@protyposis since this is the same on current `develop` branch, can we cover this in a separated PR?

protyposis:
Sure. Wasn't aware of that :/

PR #181, opened 2018-09-28
Cleanup box-sizing

protyposis:
#169 introduced a generic `box-sizing: border-box;` for all elements within the `UIContainer` to fix styling issues that appeared in an environment without Bootstrap or similar frameworks with certain style baseline resets.

Unfortunately this introduces the issue that UI components could not override this setting without `!important` or other complicated selector hacks.

This PR removes the generic `box-sizing` reset and only sets it in the appropriate places where necessary.

### DIFF
--- a/src/scss/skin-modern/_skin.scss
+++ b/src/scss/skin-modern/_skin.scss
@@ -53,13 +53,6 @@
   text-align: left;
   user-select: none;
 
-  // Include padding and border in the element's total width and height
-  *,
-  *::before,
-  *::after {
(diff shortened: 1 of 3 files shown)


protyposis:
I was thinking about that approach as well but then decided for the most non-generic change. There are many `Container` instances used in the UI and only `ControlBar` and `TitleBar` require the `box-sizing`, so I thought if anyone creates a structure that requires different `box-sizing` he could just add the property himself.

But yeah I'm not really sure what the best approach is.

PR #182, opened 2018-10-01
Remove outline when buttons have focus

stonko1994:
## Description
As discussed some time ago this will remove the blue outline from buttons when they are pressed.
bildschirmfoto 2018-10-01 um 13 24 42


protyposis:
I'm not sure if we want to remove them. Wouldn't the UI be completely unusable from the keyboard?

stonko1994:
I wasn't aware that we support this. Since this happened / works only in chrome i treated this as unintended behaviour. But anyway yes it would become completely unusable by keyboard.

dweinber:
Controlling via keyboard is required for accessibility.

protyposis:
Yap. Let's keep it as is.

PR #186, opened 2018-10-02
Add output naming customization via CLI parameters

protyposis:
Adds the following parameters to `gulp`:

 * `--outputnames.globalNamespace`: The name of the *global* on the `window` object to which the UI is attached if loaded without a module loader. Default: `bitmovin.playerui`
 * `--outputnames.filename`: The name of the generated UI CSS and JS files. Default: `bitmovinplayer-ui`
 * `--outputnames.cssPrefix`: The prefix of the UI's CSS selectors. Default: `bmpui` (as defined in `_variables.scss`)

Example: `gulp build --outputnames.filename=example` generates `example.css` and `example.js` files instead of `bitmovinplayer-ui.css` and `bitmovinplayer-ui.js` files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
+### Added
+- Customization of output naming via CLI parameters (see `outputnames` in `gulpfile.js`)
+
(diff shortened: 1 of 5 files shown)


dweinber:
Is this going to be merged anytime soon @protyposis?

PR #190, opened 2018-10-16
IE not turning off subtitles

stonko1994:
## Description
We have an issue that in IE it is not possible to turn off subtitles after turning it on once.

### Fix
The issue was that `null` will not be rendered in IE for the value of a select box option. The fix is to use always strings for the option value like other browsers do it by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [support-v2]
 
+### Changed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [support-v2]
 
+### Changed
+- Use strings as value in select box options for subtitles
```
```
### Fixed
- Handling of `null` keys in `SelectBox` (fixes subtitle deselection in IE11)
```

PR #195, opened 2018-10-22
Adding CSS selectors for default elements

stonko1994:
## Description
Some default elements such as select boxes to style the subtitles does not have specific css selectors.

## Changes
Adding css selectors where they were missing.


stonko1994:
I don't really like the current scheme but i would go for the consistency so i will change it. Maybe we can add some notes if we update to something like v4 to rework this?

PR #94, opened 2018-04-18
Improve volume control

protyposis:
* Add `data-bmpui-volume-level-tens` attribute to `VolumeToggleButton` for more granular volume icon styling and remove `bmpui-low` marker class
 * Merge mute and zero volume state (unmute always changes to > 0 volume, zero volume changes to mute state)
 * Rename `off`/`on` states of `VolumeToggleButton` to `unmuted`/`muted`
 * Remove volume (buffer) level animation from `VolumeSlider`

PR #196, opened 2018-10-25
[v2] Simplify volume handling

protyposis:
Once upon a time, the `VolumeToggleButton` displayed the mute icon when the volume was turned down to `0`. `player.isMuted()` still returned `false` because it was only the volume that was turned down, but no `player.mute()` call was done. When the user then pressed the `VolumeToggleButton`, it changed the mute state to `isMuted() === true`. In the mute state the muted icon is shown, but because it was already shown for the zero volume, nothing changed in the UI and to the user it seemed as if pressing the button did not work.

Then came #94 with the aspiration to mitigate this UX issue. Since then, it considered both `muted === true` and `volume === 0` to be muted states, so no matter if the player was muted or just the volume was turned down, a click on the button while showing the muted icon unmuted the player and raised the volume to 10% to give the user a feedback that pressing the button actually worked and the muted icon would change to the low-volume icon.

Today it turned out that this badly influences the player API, as `player.mute()` did not always work correctly when the UI was used: 

```ts
player.setVolume(0); // [player.isMuted(), player.getVolume()] => [false, 0]

// Mute the player. It is expected that isMuted() returns true afterwards
player.mute(); // [player.isMuted(), player.getVolume()] => [false, 10]
// Actually, it turned the volume to 10 instead of muting because the UI unexpectedly interfered

// A second mute call was needed to get the player into the muted state
player.mute(); // [player.isMuted(), player.getVolume()] => [true, 10]
```

The UI must not have any influence on the mechanics of the API though, so this PR removes the "UX optimization" of unifying of the mute and zero-volume states.  It now simply displays the low-volume icon for zero volume and the mute icon when the player is muted. A press on the toggle button when the volume is zero thus switches the low-volume icon to the muted icon, and vice versa, so the user now always gets a visual feedback to button presses as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [support-v2]
 
+### Changed
+- Handle audio volume and mute state separately again (pre v2.14.0 behavior)
+
 ### Fixed
 - Avoid unnecessary animation when `BufferingOverlay` is hidden
 - Avoid unnecessary DOM modification when the text of a `Label` does not change
--- a/src/scss/skin-modern/components/_volumetogglebutton.scss
+++ b/src/scss/skin-modern/components/_volumetogglebutton.scss
@@ -13,10 +13,7 @@
   }
 
   &.#{$prefix}-unmuted {
-    &[data-#{$prefix}-volume-level-tens='0'] {
-      background-image: svg('assets/skin-modern/images/music-off.svg');
-    }
-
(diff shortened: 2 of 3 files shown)


protyposis:
Yup that would be possible. Good idea.

protyposis:
While it's true that we could easily merge the muted and zero-volume states and handle everything correctly without interfering external API calls by differentiating them via the issuer, there's still the problem what happens when a muted zero-volume player is unmuted. Before, we set the volume to an arbitrary `0.1`/10% - which I always found a bit odd...

protyposis:
Replaced by #197.

PR #197, opened 2018-10-30
[v2] Improve volume handling

protyposis:
Once upon a time, the `VolumeToggleButton` displayed the mute icon when the volume was turned down to `0`. `player.isMuted()` still returned `false` because it was only the volume that was turned down, but no `player.mute()` call was done. When the user then pressed the `VolumeToggleButton`, it changed the mute state to `isMuted() === true`. In the mute state the muted icon is shown, but because it was already shown for the zero volume, nothing changed in the UI and to the user it seemed as if pressing the button did not work.

Then came #94 with the aspiration to mitigate this UX issue. Since then, it considered both `muted === true` and `volume === 0` to be muted states, so no matter if the player was muted or just the volume was turned down, a click on the button while showing the muted icon unmuted the player and raised the volume to 10% to give the user a feedback that pressing the button actually worked and the muted icon would change to the low-volume icon.

It turned out that this badly influences the player API, as `player.mute()` did not always work correctly when the UI was used: 

```ts
player.setVolume(0); // [player.isMuted(), player.getVolume()] => [false, 0]

// Mute the player. It is expected that isMuted() returns true afterwards
player.mute(); // [player.isMuted(), player.getVolume()] => [false, 10]
// Actually, it turned the volume to 10 instead of muting because the UI unexpectedly interfered

// A second mute call was needed to get the player into the muted state
player.mute(); // [player.isMuted(), player.getVolume()] => [true, 10]
```

The UI must not have any influence on the mechanics of the API though, so this PR rewrites the volume and mute state handling by moving all logic into a single `VolumeController` instance. Unmuting now sets the volume to the last "valid" value instead of an arbitrary 10%. 

Replaces #196.

### DIFF
--- a/src/ts/components/volumeslider.ts
+++ b/src/ts/components/volumeslider.ts
@@ -1,6 +1,6 @@
 import {SeekBar, SeekBarConfig} from './seekbar';
 import {UIInstanceManager} from '../uimanager';
-import { VolumeController } from '../volumecontroller';
+import { VolumeController, VolumeTransition } from '../volumecontroller';
 
 /**
  * Configuration interface for the {@link VolumeSlider} component.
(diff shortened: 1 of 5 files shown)


src/ts/components/volumeslider.ts
```diff
@@ -42,26 +43,28 @@ export class VolumeSlider extends SeekBar {
...
-    player.addEventHandler(player.EVENT.ON_UNMUTED, volumeChangeHandler);
+    let volumeTransition: VolumeTransition;
 
+    this.onSeek.subscribe(() => {
+       volumeTransition = volumeController.startTransition();
+    });
     this.onSeekPreview.subscribeRateLimited((sender, args) => {
-      if (args.scrubbing) {
-        player.setVolume(args.position, VolumeSlider.issuerName);
+      if (args.scrubbing && volumeTransition) {
```
stonko1994: The `volumeTransition` will always be resolve to true after the first time. There is no 'release'. So is it necessary to add it here?

src/ts/components/volumeslider.ts
```diff
@@ -1,5 +1,6 @@
 import {SeekBar, SeekBarConfig} from './seekbar';
 import {UIInstanceManager} from '../uimanager';
+import { VolumeController, VolumeTransition } from '../volumecontroller';
```
stonko1994: I guess the `VolumeController` import is not needed here. Also not imported in `volumetogglebutton.ts`

src/ts/components/volumeslider.ts
protyposis: It's just a safety measure in case the events for some reason do not arrive in the expected order.

protyposis:
Changelog got messed up and fixed in https://github.com/bitmovin/bitmovin-player-ui/commit/f5d56e4e722f3e4968a8fb42cb67111df6411523

PR #198, opened 2018-10-30
[v2] Set playbackSpeedSelectionEnabled to true per default

stonko1994:
## Description
We disabled the playback speed selection per default as we implemented the `playbackSpeedSelectionEnabled` flag.

## Changes
We decided to enable the speed selection per default again.

## TODO

- [ ] forward port to v3

### DIFF
--- a/src/ts/uimanager.ts
+++ b/src/ts/uimanager.ts
@@ -111,7 +111,7 @@ export interface UIConfig {
   autoUiVariantResolve?: boolean;
   /**
    * Specifies if the `PlaybackSpeedSelectBox` should be displayed within the `SettingsPanel`
-   * Default: false
+   * Default: true
    */
   playbackSpeedSelectionEnabled?: boolean;
 }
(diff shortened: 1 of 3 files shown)


src/ts/uimanager.ts
```diff
@@ -233,6 +233,11 @@ export class UIManager {
       this.uiVariants = <UIVariant[]>playerUiOrUiVariants;
     }
 
+    // Switch on speed selector by default
+    if (config.playbackSpeedSelectionEnabled === undefined) {
+      config.playbackSpeedSelectionEnabled = true;
+    }
+
     this.player = player;
     this.config = {
```
protyposis: Would be nicer to write

```ts
this.config = {
   ...{
    // Default values
    playbackSpeedSelectionEnabled: true,
  },
  ...config,
  etc...
```

instead of the `undefined` check above.

protyposis: You can also move `autoUiVariantResolve` up here.

PR #201, opened 2018-11-13
Ads UI support for player v8

stonko1994:
## Description
Up to now we do not have support for Ads in the UI as it's not part in player v8.0.

As this is getting added in player v8.1 we need to add UI support for ads as well.

## Fix
Add Ads UI support.

_Builds will fail until upgrading to player 8.1_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - CSS selector to default elements
 - `VolumeController` to control and manage volume and mute state by multiple `Component`s in a single place
+- Ads UI support for player v8
+- `UIConditionContext#adRequiresUi` to distinguish if the current ad requires an external UI
 
 ### Changed
 - Set `UIConfig.playbackSpeedSelectionEnabled` to true per default
+
+### Removed
+- Deprecated `UIConditionContext#adClientType` as it's no longer supported from the v8 player
 
 ### Fixed
 - Handling of `null` keys in `SelectBox` (fixes subtitle deselection in IE11)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Set `UIConfig.playbackSpeedSelectionEnabled` to true per default
 
 ### Removed
-- Deprecated `UIConditionContext#adClientType` as it's no longer supported from the v8 player
+- Deprecated `UIConditionContext#adClientType` as it's no longer supported by player v8
 
 ### Fixed
 - Handling of `null` keys in `SelectBox` (fixes subtitle deselection in IE11)
--- a/src/scss/skin-modern/_skin-ads.scss
+++ b/src/scss/skin-modern/_skin-ads.scss
@@ -28,16 +28,6 @@
         }
       }
 
-      /* Let button look like it is not click-able during the skippable-in countdown */
-      &.countdown-active {
-        cursor: default;
-        .#{$prefix}-label {
-          &:hover {
-            text-decoration: none;
-          }
-        }
-      }
-
       // Add the dot between ad message and skip button
       &::before {
         color: $color-highlight;
--- a/src/scss/skin-modern/_skin-ads.scss
+++ b/src/scss/skin-modern/_skin-ads.scss
@@ -29,7 +29,7 @@
       }
 
       /* Let button look like it is not click-able during the skippable-in countdown */
-      &.skippable-in-countdown {
+      &.countdown-active {
         cursor: default;
         .#{$prefix}-label {
           &:hover {
--- a/src/scss/skin-modern/_skin-ads.scss
+++ b/src/scss/skin-modern/_skin-ads.scss
@@ -28,6 +28,16 @@
         }
       }
 
+      /* Let button look like it is not click-able during the skippable-in countdown */
+      &.skippable-in-countdown {
+        cursor: default;
+        .#{$prefix}-label {
+          &:hover {
+            text-decoration: none;
+          }
+        }
+      }
+
       // Add the dot between ad message and skip button
       &::before {
         color: $color-highlight;
--- a/src/ts/components/adclickoverlay.ts
+++ b/src/ts/components/adclickoverlay.ts
@@ -15,7 +15,7 @@ export class AdClickOverlay extends ClickOverlay {
     player.on(player.exports.PlayerEvent.AdStarted, (event: AdEvent) => {
       let ad = event.ad;
       this.setUrl(ad.clickThroughUrl);
-      clickThroughCallback = ad.adClickedCallback;
+      clickThroughCallback = ad.clickThroughUrlOpened;
     });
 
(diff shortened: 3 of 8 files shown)


src/ts/components/adclickoverlay.ts
```diff
@@ -10,39 +10,30 @@ export class AdClickOverlay extends ClickOverlay {
   configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
     super.configure(player, uimanager);
 
-    let clickThroughUrl = <string>null;
-    let clickThroughEnabled = !player.getConfig().advertising
-      || !player.getConfig().advertising.hasOwnProperty('clickThroughEnabled')
-      || (player.getConfig().advertising as any).clickThroughEnabled;
+    let clickThroughCallback: () => void = null;
 
     player.on(player.exports.PlayerEvent.AdStarted, (event: AdEvent) => {
```
offnertho: as discussed offline pls add a comment why this works.

src/ts/components/adclickoverlay.ts
```diff
@@ -10,39 +10,30 @@ export class AdClickOverlay extends ClickOverlay {
...
     };
+
     player.on(player.exports.PlayerEvent.AdFinished, adFinishedHandler);
     player.on(player.exports.PlayerEvent.AdSkipped, adFinishedHandler);
     player.on(player.exports.PlayerEvent.AdError, adFinishedHandler);
 
     this.onClick.subscribe(() => {
       // Pause the ad when overlay is clicked
-      player.pause('ui-content-click');
+      player.pause('ui-ad-click-overlay');
```
offnertho: do we have to pause the content here? 

Also we should not add the `ad` name, because they might be detected more easily by ad blockers.

src/ts/components/adskipbutton.ts
```diff
@@ -32,29 +34,32 @@ export class AdSkipButton extends Button<AdSkipButtonConfig> {
...
         this.show();
       } else {
         this.hide();
       }
 
       // Update the skip message on the button
       if (player.getCurrentTime() < skipOffset) {
-        this.setText(
-          StringUtils.replaceAdMessagePlaceholders(skipMessage.countdown, skipOffset, player));
+        this.setText(StringUtils.replaceAdMessagePlaceholders(skipMessage.countdown, skipOffset, player));
```
offnertho: pls specify the exact placeholders in the related interface in https://github.com/bitmovin/player-specification/issues/38#issuecomment-438584897

src/ts/components/adskipbutton.ts
```diff
@@ -32,29 +34,32 @@ export class AdSkipButton extends Button<AdSkipButtonConfig> {
...
+        this.getDomElement().addClass(AdSkipButton.SKIPPABLE_IN_COUNTDOWN_CLASS);
       } else {
         this.setText(skipMessage.skip);
+        this.getDomElement().removeClass(AdSkipButton.SKIPPABLE_IN_COUNTDOWN_CLASS);
```
offnertho: how can someone customize where this skip message gets dipslayed.

src/ts/components/adclickoverlay.ts
protyposis: Yes when the clickthough Url is opnened the ad needs to be paused.

src/ts/components/adskipbutton.ts
protyposis: Similar to every other component in the UI framework. Via the container tree and CSS.

src/ts/components/adclickoverlay.ts
protyposis: If you're referring to last week's discussion, ad blockers are looking for "ad" in DOM classes and IDs, and maybe filenames, but most certainly not values of random variables in JS code. If that were the case, our whole UI would have alwas been blocked (`AdClickOverlay`, `AdSkipButton`, etc).

CHANGELOG.md
```diff
@@ -9,10 +9,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
 - CSS selector to default elements
 - `VolumeController` to control and manage volume and mute state by multiple `Component`s in a single place
+- Ads UI support for player v8
+- `UIConditionContext#adRequiresUi` to distinguish if the current ad requires an external UI
 
 ### Changed
 - Set `UIConfig.playbackSpeedSelectionEnabled` to true per default
 
+### Removed
+- Deprecated `UIConditionContext#adClientType` as it's no longer supported from the v8 player
```
protyposis: "supported by"

protyposis: "player v8" instead of "the v8 player" (sounds better, and consistency with entry above)

src/scss/skin-modern/_skin-ads.scss
```diff
@@ -28,6 +28,16 @@
         }
       }
 
+      /* Let button look like it is not click-able during the skippable-in countdown */
+      &.skippable-in-countdown {
```
protyposis: What does this name mean? Why is it not part of the adskipbutton style?

src/ts/components/adclickoverlay.ts
protyposis: What exactly is unclear?

stonko1994: It was about how the `AdClickoverlay` (and other ad components) gets the `AdStarted` event because they get initialized when the first `AdStarted` event gets triggered.

stonko1994:
@offnertho about moving the skip button to the right: I think this should not be part of this PR

src/ts/components/adclickoverlay.ts
protyposis: I see. This description does not belong in here. This component simply listens to the event and does not need to be aware nor care why it receives it.

src/scss/skin-modern/_skin-ads.scss
stonko1994: I will look for a better name and rename it.
We don't have separate files for ad related component styling.

stonko1994: What about the name `countdown-active` ?

protyposis: Yap the file placement makes perfect sense.

Regarding the classname, I don't think this should be specifically taylored to the "countdown" usecase. It can just be generically called "disabled".

The beautiful solution would be to extend `Component` with an `enabled` state (similar to `hidden`) with `enable`, `disable`, `isEnabled` methods an `onEnable`, `onDisable` events. This would be quite simple to implement and make a nice feature addition that had to be added anyway at some point, and I think this is the right time now. In `_button.scss` you would add a `.disabled` style which is basically what you did here. In `AdSkipButton` you'd simply `enable`/`disable` the button accordingly.

stonko1994: Added in a separate PR -> #202

offnertho:
screenshot 2018-11-16 at 15 29 47

It seems that the time calculation is still a bit off

protyposis:
To get this building and us being able to create a UI release for the player release, you can temporarily add local types and use them, e.g.

```ts
// TODO remove once an player package with the updates types is available
interface LocalLinearAd extends LinearAd {
  uiConfig: ...
}
```

PR #202, opened 2018-11-15
Add enable and disable functionality to components

stonko1994:
## Description
As we want to disable the `AdSkipButton` during the countdown we needed some mechanism to enable / disable components. (Ads UI PR #201)

## Added
This will add `disable()` / `enable()` methods and disable click events on buttons.


src/scss/skin-modern/_skin-ads.scss
```diff
@@ -28,16 +28,6 @@
         }
       }
 
-      /* Let button look like it is not click-able during the skippable-in countdown */
```
protyposis: This idicates that the branching base was wrong (branch from #202 instead of develop).

PR #203, opened 2018-11-29
[v2] Add events to update quality select boxes

stonko1994:
## Description
Add `ON_VIDEO_QUALITY_ADDED` `ON_VIDEO_QUALITY_REMOVED` `ON_AUDIO_QUALITY_ADDED` `ON_AUDIO_QUALITY_REMOVED` events to listen to availability changes.

## TODO

- [ ] port to v3

PR #204, opened 2018-12-03
Add events to update quality select boxes

stonko1994:
## Description
Forward port of #203


src/ts/components/audioqualityselectbox.ts
```diff
@@ -52,6 +52,12 @@ export class AudioQualitySelectBox extends SelectBox {
     player.on(player.exports.PlayerEvent.PeriodSwitched, updateAudioQualities);
     // Update quality selection when quality is changed (from outside)
     player.on(player.exports.PlayerEvent.AudioQualityChanged, selectCurrentAudioQuality);
+    if (player.exports.PlayerEvent.hasOwnProperty('AudioQualityAdded')) {
```
protyposis: Please write these checks similar to their usage below (`if ((player.exports.PlayerEvent as any).AudioQualityAdded) { ...`) to avoid strings which are easily missed in searches and refactorings/renamings.

PR #205, opened 2018-12-12
Update test page

protyposis:
Updates all functionality in the test page to v8 API:

 * Enumerate module namespace methods
 * Sort API methods and events a-z
 * Update ads test actions
 * Update subtitle test actions
 * Update resize test action
 * Add source reload test action
 * Decrease ad schedulting delay from 5 to 2 seconds
 * Add native video element event logging

### DIFF
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -380,22 +380,22 @@
         position: 'pre',
       });
     },
-    'Schedule skippable VAST ad in 5 secs': function() {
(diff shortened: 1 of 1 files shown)


protyposis:
> for the buffer api we need 8.1.0-rc1 (my bad forgot that this is not yet released, may update to rc1 or add checks or revert changes)

This is not an issue because buffer API calls are only enumerated if the buffer API exists.

PR #207, opened 2018-12-14
Improve ad variant switching

protyposis:
#201 added ads UI support for UI v3 but didn't consider the new ad playback event sequence from player v8. This wasn't and issue with player 8.0 with the IMA ads module (because it uses an external player for playback), but is an issue for player 8.1 and the new Bitmovin Advertising module.

 * Enter ads UI on `PlayerEvent.AdStarted`
 * Exit ads UI on `PlayerEvent.AdBreakFinished`
 * Refresh active UI variants when exiting ads UI (when switching from ads to main content)

### DIFF
--- a/src/ts/uimanager.ts
+++ b/src/ts/uimanager.ts
@@ -216,29 +216,33 @@ export class UIManager {
 
     // Dynamically select a UI variant that matches the current UI condition.
     let resolveUiVariant = (event: PlayerEventBase) => {
-      // Make sure that the ON_AD_STARTED event data is persisted through ad playback in case other events happen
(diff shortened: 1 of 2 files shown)


src/ts/uimanager.ts
```diff
@@ -247,7 +260,7 @@ export class UIManager {
       let adRequiresUi = false;
       if (isAd) {
         let ad = adStartedEvent.ad;
-        // for now only linear ads can request an UI
+        // for now only linear ads can request a UI
```
stonko1994: Shouldn't it be `an`

protyposis: https://www.quora.com/Is-it-an-user-or-a-user

PR #223, opened 2019-02-14
Do not keep updating the skipMessage if the ad is not skippable

Reevn:
### Description
The text of the skip button was constantly re-set to the same message, even though the button is hidden because the active ad is not skippable.

### Fix
`updateSkipMessageHandler` now returns early if the active ad is not skippable.

Additionally, in player version < `v8.3.0`, `Ad.skippableAfter` was always set to `-1` if an ad was not skippable. Starting with `v8.3.0`, `Ad.skippableAfter` will not be set if an ad is not skippable. I adjusted the relevant check to support both cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - `UIContainerConfig.userInteractionEventSource` to allow tracking of user interaction events (which toggle the visibility of certain components like the `ControlBar`) on a custom element
 
 ### Changed
-- Hidden `adSkipButton` is not constantly updated anymore if a non skippable ad is playing
+- Avoid unnecessary updating of hidden `AdSkipButton`
 
 ## [3.3.1]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - `UIContainerConfig.userInteractionEventSource` to allow tracking of user interaction events (which toggle the visibility of certain components like the `ControlBar`) on a custom element
 
+### Changed
+- Hidden `SkipMessage` is not constantly updated anymore if an ad is playing and is not skippable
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - `UIContainerConfig.userInteractionEventSource` to allow tracking of user interaction events (which toggle the visibility of certain components like the `ControlBar`) on a custom element
 
+### Changed
+- Hidden `SkipMessage` is not constantly updated anymore if an ad is playing and is not skippable
```
```suggestion
- Hidden `adSkipButton` is not constantly updated anymore if a non skippable ad is playing
```

src/ts/components/adskipbutton.ts
```diff
@@ -44,13 +44,14 @@ export class AdSkipButton extends Button<AdSkipButtonConfig> {
...
-      if (skipOffset >= 0) {
-        this.show();
-      } else {
+      // non skippable ads will return -1 for skippableAfter for player version < v8.3.0
+      if (!skipOffset || skipOffset < 0) {
         this.hide();
+        return;
       }
 
+      this.show();
```
protyposis: Wouldn't it be cleaner to prevent the handler attachment inside the `adStartHandler` below? Now this handler is still unnecessarily executed on every `TimeChanged` event.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - `UIContainerConfig.userInteractionEventSource` to allow tracking of user interaction events (which toggle the visibility of certain components like the `ControlBar`) on a custom element
 
+### Changed
+- Hidden `adSkipButton` is not constantly updated anymore if a non skippable ad is playing
```
protyposis: `` Avoid unnecessary updating of hidden `AdSkipButton` `` (component names always start with an uppercase letter)

Issue #208, opened 2018-12-19
Update to gulp4


michael-riha:
I did a checkout today and ran into this issue https://github.com/gulpjs/gulp/issues/2162

`npm 6.5.0`
`node v11.5.0` but the same with which I had before  `node v10.0.2`

protyposis:
Update to Gulp 4 is not planned. Did you try any of the workarounds in the linked issue?

michael-riha:
yes but they are messy on a global scope of NPM.

protyposis:
Simply using Gulp 4 seems to solve the issue. No need to adjust the `gulpfile`.

michael-riha:
shouldn't it be documented somewhere?
repo package.json still fetches 3.9.1 and breaks.

PR #226, opened 2019-02-21
Upgrade to Gulp 4

dweinber:
Newer node versions started to fail running gulp3. This PR upgrades the build chain to gulp4 based on https://www.liquidlight.co.uk/blog/how-do-i-update-to-gulp-4/

Issue: https://github.com/bitmovin/bitmovin-player-ui/issues/208

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -115,11 +115,7 @@ gulp.task('lint-sass', function() {
 });
 
 // Runs all linters
-gulp.task('lint', function(callback) {
-  // this fails at first error so we can't run all linters sequentially with runSequence and then print the errors
-  runSequence('lint-ts', 'lint-sass', callback);
-  // TODO check in Gulp 4.0 if all linters can be run sequentially before aborting due to an error
-});
+gulp.task('lint', gulp.series('lint-ts', 'lint-sass'));
 
 // Copies html files to the target directory
 gulp.task('html', function() {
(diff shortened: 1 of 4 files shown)


gulpfile.js
```diff
@@ -199,28 +193,22 @@ gulp.task('sass', function() {
...
-    callback);
-});
+gulp.task('build', gulp.series('clean', gulp.parallel('html', 'browserify', 'sass')));
 
-gulp.task('build-prod', function(callback) {
-  production = true;
-  runSequence('lint', 'build', callback);
-});
+gulp.task('build-prod', gulp.series(function(callback) {
+  production = true
```
protyposis: missing semicolon

gulpfile.js
```diff
@@ -199,28 +193,22 @@ gulp.task('sass', function() {
...
+  callback();
```
protyposis: I think the callback is not necessary (no async code)

dweinber: I thought the same and I didn't have it in the first place. 

But if I change it to:
```ts
gulp.task('build-prod', gulp.series(function() {
  production = true;
}, 'lint', 'build'));
```
I get an error:
```
$ gulp build-prod
Using gulpfile ~/Projects/bitmovin-player-ui/gulpfile.js
Starting 'build-prod'...
Starting '<anonymous>'...
The following tasks did not complete: build-prod, <anonymous>
Did you forget to signal async completion?
```

Therefore, I added the callback and it works fine then. Happy to change it if we can find out what the problem is, but currently I don't really have an idea.

dweinber:
The `<anonymous>` tasks are all the anonymous functions used in `gulp.serial` or `gulp.parallel` calls.

gulpfile.js
```diff
@@ -116,11 +114,7 @@ gulp.task('lint-sass', function() {
 });
 
 // Runs all linters
-gulp.task('lint', function(callback) {
-  // this fails at first error so we can't run all linters sequentially with runSequence and then print the errors
-  runSequence('lint-ts', 'lint-sass', callback);
-  // TODO check in Gulp 4.0 if all linters can be run sequentially before aborting due to an error
-});
+gulp.task('lint', gulp.series('lint-ts', 'lint-sass'));
```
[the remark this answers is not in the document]
dweinber: I'll have another look.

Now that I think about this - why do we have to run them sequentially as stated in the comment? Can't we run them in parallel?

gulpfile.js
dweinber: Missed that part, thanks!

gulpfile.js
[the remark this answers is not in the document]
dweinber: This would output both, sass and ts lint errors/warning.

PR #228, opened 2019-03-11
Fix stopping time shift offset updater of SeekBar

jmsn:
### Issue

If the player is paused on a live stream that has a time shift window, a `Timeout` is started which periodically updates the displayed time shift offset of the seek bar. In case the player is destroyed in this state, the updater would not get stopped and would cause uncaught `PlayerAPINotAvailableError`s.

### Fix
Added clearing the updater when `release` is called on the `SeekBar`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- Stopping time shift offset updater of `SeekbBar` when player is destroyed
```
protyposis: Typo

PR #233, opened 2019-03-22
Improve rateLimiting handling of RateLimitedEventListener

stonko1994:
## Description
Currently, the `RateLimitedEventListener` is checking if the delta of the `lastFiredTime` to the current timestamp is greater than the rate limit and if it will delegate the event. This lead to all events during the delta period will not be received.

## Problem
This leads to an issue e.g that the time in the `seekbarLabel` could be inaccurate if the user moves the mouse fast enough over the `seekbar`. (Moving during the delta period) Those events won't be delegated to the label and the time won't update. This again leads to the impression that the seeking is inaccurate 'cause the `currentTime` after seeking is different to the time seen in the `seekbarLabel`.

## Fix
We came up with the solution that the last seen event need to be triggered anyway after the `rateLimit` period.

## How to test
The easiest way to test this is to modify the `rateMs` value within the `seekbarlabel.ts#86` to something like 5000 and move the mouse over the seekbar and stop somewhen within the rate limit. The time in the `seekbarLabel` should then update after the period.
_The longer the stream is the easier this is to reproduce_

## Discussion
Should we add a flag if this feature (`fire-last-seen-event-after-rate-limit`) should be enabled or not?
_Currently there is no such flag_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
+### Changed
+- Dispatch last event of a rate-limited event sequence to `Event.subscribeRateLimited` listeners
+
 ### Fixed
 - Inaccurate time within `SeekBarLabel` on seek preview
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- Inaccurate time within `SeekBarLabel` on seek preview
+
 ## [3.4.1]
 
 ### Added
--- a/src/ts/eventdispatcher.ts
+++ b/src/ts/eventdispatcher.ts
@@ -255,7 +255,7 @@ class RateLimitedEventListenerWrapper<Sender, Args> extends EventListenerWrapper
   }
 
   private shouldFireEvent(): boolean {
-    return !this.rateLimitTimout.isRunning();
+    return !this.rateLimitTimout.isActive();
   }
 
   private fireSuper(sender: Sender, args: Args) {
--- a/src/ts/eventdispatcher.ts
+++ b/src/ts/eventdispatcher.ts
@@ -244,6 +244,7 @@ class RateLimitedEventListenerWrapper<Sender, Args> extends EventListenerWrapper
       if (this.shouldFireEvent()) {
         this.fireSuper(sender, args);
         startRateLimiting();
+        return;
       }
 
       this.lastSeenEvent = {
--- a/src/ts/eventdispatcher.ts
+++ b/src/ts/eventdispatcher.ts
@@ -197,9 +197,6 @@ class EventListenerWrapper<Sender, Args> {
  */
 class RateLimitedEventListenerWrapper<Sender, Args> extends EventListenerWrapper<Sender, Args> {
 
-  // Timout in MS where the last seen event is delayed
-  private static NEXT_EVENT_TIMEOUT_MS = 50;
-
   private readonly rateMs: number;
   private readonly rateLimitingEventListener: EventListener<Sender, Args>;
 
@@ -207,7 +204,6 @@ class RateLimitedEventListenerWrapper<Sender, Args> extends EventListenerWrapper
   private lastSeenEvent: any;
 
   private rateLimitTimout: Timeout;
-  private waitAndFireLastTimout: Timeout;
 
   // Boolean to track if rateLimit timeout is currently active
   private isRateLimiting: boolean = false;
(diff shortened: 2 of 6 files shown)


src/ts/eventdispatcher.ts
```diff
@@ -196,28 +197,76 @@ class EventListenerWrapper<Sender, Args> {
...
-  private rateLimitingEventListener: EventListener<Sender, Args>;
+  // Timout in MS where the last seen event is delayed
+  private static NEXT_EVENT_TIMEOUT_MS = 50;
 
-  private lastFireTime: number;
+  private readonly rateMs: number;
+  private readonly rateLimitingEventListener: EventListener<Sender, Args>;
+
+  // save last seen event attributes
+  private lastSeenEvent: any;
```
protyposis: Please add the type (it is known)

src/ts/eventdispatcher.ts
```diff
@@ -196,28 +197,76 @@ class EventListenerWrapper<Sender, Args> {
...
 
   constructor(listener: EventListener<Sender, Args>, rateMs: number) {
     super(listener); // sets the event listener sink
 
     this.rateMs = rateMs;
-    this.lastFireTime = 0;
 
-    // Wrap the event listener with an event listener that does the rate-limiting
+    // delegating the received event
+    const fireEvent = (sender: Sender, args: Args) => {
```
protyposis: Why don't you call `fireSuper` directly? This function seems unnecessary.

src/ts/eventdispatcher.ts
```diff
@@ -196,28 +197,76 @@ class EventListenerWrapper<Sender, Args> {
...
+  private readonly rateLimitingEventListener: EventListener<Sender, Args>;
+
+  // save last seen event attributes
+  private lastSeenEvent: any;
+
+  private rateLimitTimout: Timeout;
+  private waitAndFireLastTimout: Timeout;
+
+  // Boolean to track if rateLimit timeout is currently active
+  private isRateLimiting: boolean = false;
```
protyposis: The name is very misleading, to me it signals that rate limiting can be active or disabled. It is always active though.

src/ts/eventdispatcher.ts
```diff
@@ -196,28 +197,76 @@ class EventListenerWrapper<Sender, Args> {
...
+    const startRateLimiting = () => {
+      if (this.lastSeenEvent) {
+        this.isRateLimiting = true;
+        this.rateLimitTimout.start();
+      }
+    };
+
+    // We need to wait a short time-windows to allow a current event to get delegated instead of the last saved one.
+    // Otherwise the lastSaveEvent will always be triggered after the rateLimiting.
+    this.waitAndFireLastTimout = new Timeout(RateLimitedEventListenerWrapper.NEXT_EVENT_TIMEOUT_MS, () => {
```
protyposis: I don't understand `NEXT_EVENT_TIMEOUT_MS`. Why isn't it simply `rateMs`? Yes there's the description above but I don't understand this either.

stonko1994: Let me try that again:

_Let's assume we have a rate limit of 5 seconds._
1. The first event comes in and will be delegated immediately.
2. The rate limit timer with 5 seconds will be started.
    1. Each following event within the 5 seconds will be saved as last-seen event.

**Now we have 2 use cases:**
1. Incoming events stops during the rate limit timer. ->
    1. The rate limit timer expires after 5 seconds
    2. The last-seen event will be triggered

2. Incoming events never stops ->
    1. The rate limit timer expires after 5 seconds
    2. We need to check if a 'real' event is coming in (within the timeout `NEXT_EVENT_TIMEOUT_MS `)
Without this check, it could happen that a 'real' event will come at the same time we want to fire the last-seen event and that lead to one extra event within the rate limit timout.

        1. Yes, a 'real' event came within the interval (maybe at 5.00001) 
            1. cancel firing last seen event and delegate 'real' event
            2. restart rate limit
        2. No, no 'real' event came -> send last seen event

protyposis: IMO case 2 is not a problem and can be ignored. The event arriving a 5.00001 missed the 5 second interval and thus goes into the next one, which is pefectly fine. We can't make the interval longer just to make sure, because there can always be an event which is just an epsilon too late, even after `NEXT_EVENT_TIMEOUT_MS` - what happens then?

protyposis:
The timeout itself is also a danger that we somehome need to solve. Assume that the UI is destroyed while a timeout is active. In the best case, it will prevent the UI from being completely destroyed until the timeout has passed. In the worst case, it will access an API afterwards, e.g. the player API of a destroyed player, and trigger an `ApiNotAvailableError`. There is no "global" way to cancel all timers since every `EventDispatcher` lives on its own, but we could add a `reset` method to be called by components instantiating the dispatchers, and within the UI we know where rate limiting is used and call that method at shutdown.

src/ts/timeout.ts
```diff
@@ -13,6 +13,7 @@ export class Timeout {
   // To work around the issue we use type "any". The type does not matter anyway because we're not working with
   // this value except providing it to clearTimeout.
   private timeoutOrIntervalId: any;
+  private running: boolean;
```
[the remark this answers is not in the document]
stonko1994: `start()` just calls `reset` and it's set to true there so I decided to not set it there too.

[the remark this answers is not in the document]
bitmovin-kenny: Ok "makes sense", didn't catch that.

src/ts/eventdispatcher.ts
```diff
@@ -189,35 +195,68 @@ class EventListenerWrapper<Sender, Args> {
...
+    // rate limiting is finished. This prevents missing the last update due to the rate limit.
     this.rateLimitingEventListener = (sender: Sender, args: Args) => {
-      if (Date.now() - this.lastFireTime > this.rateMs) {
-        // Only if enough time since the previous call has passed, call the
-        // actual event listener and record the current time
+      // only fire events if the rateLimiting is not running
+      if (this.shouldFireEvent()) {
         this.fireSuper(sender, args);
-        this.lastFireTime = Date.now();
+        startRateLimiting();
```
protyposis: I think you need to `return;` here, else this fired event will also be set as `lastSeenEvent` and fired again when the timeout times out.

src/ts/eventdispatcher.ts
```diff
@@ -227,4 +266,9 @@ class RateLimitedEventListenerWrapper<Sender, Args> extends EventListenerWrapper
     // Fire the internal rate-limiting listener instead of the external event listener
     this.rateLimitingEventListener(sender, args);
   }
-}
\ No newline at end of file
+
+  clear(): void {
```
protyposis: This must also be called when unsubscribing an event handler. Then we have to explicitely unsubscribe all event handlers subscribed with `subscribeRateLimited` to avoid leaks.

stonko1994:
@protyposis I added that one here: [6e77eaa](https://github.com/bitmovin/bitmovin-player-ui/pull/233/commits/6e77eaa83547e6169d5127ae1d70c1340409a0f9). Or what do you mean exactly?
The `subscribeRateLimited` listener is held in the same listener array as the normal ones.

protyposis:
You need to unsubscribe the following handlers in `Component.release`:

https://github.com/bitmovin/bitmovin-player-ui/blob/b610a4fab615ebc375243f7a21bd34077e52f847/src/ts/components/seekbar.ts#L267
https://github.com/bitmovin/bitmovin-player-ui/blob/a226476aecf6b76221866c862270c96f82d27433/src/ts/components/volumeslider.ts#L62

The one that you don't necessarily need to unsubscribe is this one
https://github.com/bitmovin/bitmovin-player-ui/blob/d2f900bfc6c71a24393e2101bd2ebf0267a2fb97/src/ts/components/seekbarlabel.ts#L58

because it will be unsubscribed here
https://github.com/bitmovin/bitmovin-player-ui/blob/d2f900bfc6c71a24393e2101bd2ebf0267a2fb97/src/ts/uimanager.ts#L599-L603
but for the sake of completeness and consistency you could still do it.

I have not tested it but I believe that if you do not unsubscribe them, they potentially live the rate limiting time (e.g. 100ms) beyond the destruction of the UI, thus leaking parts of the UI for that time and probably leading to invalid player API calls.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- Inaccurate time within `SeekBarLabel` on seek preview
+
```
protyposis: Please also add a 

```
### Changed
- Dispatch last event of a rate-limited event sequence to `Event.subscribeRateLimited` listeners
```

PR #235, opened 2019-04-12
Chromecast UI doesn't disappear when playing

stonko1994:
## Problem
When starting casting during playback the UI on the cast receiver sometimes doesn't disappear. This is due to the issue that the `player.isPlaying()` call returns `false` within the `seeked` event.

## Fix
The fix is to listen to the `playing` event and start the hide timer there instead of the seeked event. This is more reliable and we can remove the `seeked` handling.

## How to Test
- Check out this branch
- Build a custom receiver app which uses the `UIFactory.modernCastReceiverUI()` UI
- Start casting during playback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- `CastReceiverUI` does not hide when start casting during playback
+
 ## [3.4.2]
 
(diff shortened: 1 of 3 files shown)


protyposis (changes requested):
I don't think this is working correctly. When a seek is done during playback, the `Seek` event will now display the UI permanently and there's no `Seeked` listener that hides it again. There is no `Play`/`Playing` event after a seek because the player is already in the playing state in this case.

stonko1994:
@protyposis I thought the same but as the UI pauses the player before calling `seek` we always have a `Playing` event after seeking so it is actually working as expected.

stonko1994:
I will close this PR for now as this is due to a wrong player state.

PR #237, opened 2019-04-18
Change the AirPlay icon into active

stonko1994:
## Description
Currently there was no event listener in place to check if AirPlay is currently active.

## Fix
Add an event listener for `AirPlayChanged` event and turn on the active state of the icon.

### DIFF
--- a/src/ts/components/airplaytogglebutton.ts
+++ b/src/ts/components/airplaytogglebutton.ts
@@ -43,7 +43,7 @@ export class AirPlayToggleButton extends ToggleButton<ToggleButtonConfig> {
       }
     };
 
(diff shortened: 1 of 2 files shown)


src/ts/components/airplaytogglebutton.ts
```diff
@@ -43,9 +43,19 @@ export class AirPlayToggleButton extends ToggleButton<ToggleButtonConfig> {
       }
     };
 
+    let airPlayChangeHandler = () => {
```
protyposis: `const airPlayChangedHandler` (note the added "d")

stonko1994:
@offnertho That's a player issue and already tracked internally.

PR #239, opened 2019-04-30
First attempt to add unit tests in the UI

stonko1994:
## Description
This PR will add a simple test setup for test using jest

## Test
Simple test which tests if `UIContainer.release` does not crash even if `uiHideTimeout` and / or `userInteractionEvents` is undefined.

@protyposis Some questions:
- What do you think about the setup (playerMock, ...)
- I'm aware that this test never calls `configure` for example but it's not needed to test this fix.
- What about expecting private properties (with any cast)


spec/components/uicontainer.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+  describe('release', () => {
+    it('checks if userInteractionEvents and uiHideTimeout is undefined', () => {
+      const uiContainer = new UIContainer({
+        hideDelay: -1, // With an hideDelay of -1 the uiHideTimeout and userInteractionEvents never will be initialized
+        components: [],
+      });
+
+      uiContainer.release();
+      // Ensure that we don't try to call something on undefined properties
+      expect((uiContainer as any).uiHideTimeout).toBeUndefined();
```
protyposis: How does this check ensure that no calls are made?

spec/helper/MockHelper.ts
```diff
@@ -0,0 +1,318 @@
...
+
+    return new PlayerMockClass();
+  }
+}
+
+export interface TestingPlayerAPI extends PlayerAPI {
+  eventEmitter: EventEmitter;
+}
+
+interface EventEmitter {
```
protyposis: What's the purpose of this interface if there's only one class that implements it?

spec/components/uicontainer.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+let playerMock: TestingPlayerAPI;
+
+describe('UIContainer', () => {
+  beforeEach(() => {
+    playerMock = MockHelper.getPlayerMock();
+  });
+
+  describe('release', () => {
+    it('checks if userInteractionEvents and uiHideTimeout is undefined', () => {
```
protyposis: I think this test needs to be split in two. The first one testing if `release` does not crash, and the second one if `userInteractionEvents` and `uiHideTimeout` are undefined, which only makes sense if  `configure` is called (without `configure` they will always be undefined).

protyposis: Although, `release` only crashed if those properties were undefined, right? Anyway without calling `configure` the test does not cover the actual issue we had. So maybe two tests which check if `release` works with and without `configure`.

spec/components/uicontainer.spec.ts
stonko1994: It actually doesn't but it would crash if so ... I wasn't sure how to test as they are private properties and I don't like tests called `should not crash ...` and without actual expectation.

spec/helper/MockHelper.ts
stonko1994: Good question ... actually nothing :D

spec/components/uicontainer.spec.ts
stonko1994: > Although, release only crashed if those properties were undefined, right?

Yes

stonko1994:
> Can you include the tests into the Travis build?

Yes

PR #240, opened 2019-05-10
Support start time for live

stonko1994:
## Description
The upcoming player version v8.8.0 will support the `SourceConfigOption` `startTime` also for livestreams.

## Problem
No initial `TimeShifted` event will be fired when a `startTime` is configured.

## Solution
We need to listen to the `Playing` event to update the live-edge status indicator.


src/ts/components/playbacktimelabel.ts
```diff
@@ -116,6 +116,7 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
     player.on(player.exports.PlayerEvent.TimeShift, updateLiveTimeshiftState);
     player.on(player.exports.PlayerEvent.TimeShifted, updateLiveTimeshiftState);
     player.on(player.exports.PlayerEvent.Play, updateLiveTimeshiftState);
```
offnertho: do we still need the `play`?

stonko1994: No

Issue #234, opened 2019-04-05
Subtitle or Audio select on demofactory not close UI


vfernandes0408:
This bug is related a demofactory.ts.
When i use this ts and set another subtitle different default has the problem that ui not hidden.

stonko1994:
Thanks for reporting we will look into it.

vfernandes0408:
Any news about this issue ?

stonko1994:
We were able to reproduce the issue and we have an internal issue open but there's no ETA for a fix at this time.

PR #242, opened 2019-05-21
Fix issue that UI stays active after selecting something in a listBox

stonko1994:
## Description
Issue #234 

The UI stays active after a user selects something within a `ListBox`.

**Sry about the size of the PR. The most changes are new tests.**

## Problem
The `ListBox` recreates its elements way too often. Together with our `hoverStackCount` which prevents hiding the UI when a user hovers over an element this lead to the issue that the `hoverStackCount` doesn't unregister the selected item which gets recreated immediately after selecting something. The outcome is that the whole UI stays active forever.

## Solution
Properly update the `ListBox` and its items instead of simply recreating it all the time.

## Manual tests
Select the `buildDemoWithSeparateAudioSubtitlesButtons` and change the current `Subtitle` or `AudioTrack` within the `Subtitle-` or `AudioTrackListBox`

## Tests
- Added tests for:
  - `ListSelector`
  - `AudioUtils`
  - `SubtitleUtils`


src/ts/components/listbox.ts
```diff
@@ -13,8 +16,10 @@ import {DOM} from '../dom';
  *   </div
  * </code>
  */
+// TODO: change to extends container in v4
```
protyposis: Make it a little clearer (e.g. that the `ListSelector` should extend `Container`)

src/ts/components/listbox.ts
```diff
@@ -24,65 +29,94 @@ export class ListBox extends ListSelector<ListSelectorConfig> {
...
-  }
+  };
 
-  private handleSelectionChange(sender: ListBoxItemButton) {
-    this.onItemSelectedEvent(sender.key);
-  }
+  private refreshSelectedItem = () => {
+    // This gets called twice because the first time is triggered when the user clicks on the ListBoxItemButton. And the
+    // second call comes from the player event when the actual item is selected (Subtitle / AudioTrack in this case).
+    // As this is a generic component we can't prohibit this behaviour. We need to tread this component as it acts
```
protyposis: Typo: `tread` -> `treat`

src/ts/components/listbox.ts
```diff
@@ -24,65 +29,94 @@ export class ListBox extends ListSelector<ListSelectorConfig> {
...
+    // independent form PlayerEvents and on the other hand we need to react to PlayerEvents as it could be triggered
```
protyposis: typo: `form` -> `from`

src/ts/components/listselector.ts
```diff
@@ -108,7 +116,9 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
       listItem.label = this.config.translator(listItem);
     }
 
-    this.removeItem(key); // Try to remove key first to get overwrite behavior and avoid duplicate keys
+    // Try to remove key first to get overwrite behavior and avoid duplicate keys
+    this.removeItem(key); // FIXME: This will trigger an ItemRemoved and an ItemAdded event which should not be the case
```
protyposis: I think this is a matter of taste. One could also argue that there should not be an `ItemAdded` event when an overwrite happens, because it's been there before (count stays the same, thus nothing is "added").

src/ts/components/listselector.ts
```diff
@@ -159,6 +169,31 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
...
+   * Returns the items for the given key or undefined if no item with the given key exists.
+   * @param key the key of the item to return
+   * @returns {ListItem} the item with the requested key. Undefined if no item with the given key exists.
+   */
+  getItemForKey(key: string): ListItem {
+    return this.items.find((item) => item.key === key);
+  }
+
+  /**
+   * Synchronize the current items of this selector with the given once. This will remove and add items selectively.
```
protyposis: typo: `once` -> `ones`

src/ts/components/listselector.ts
```diff
@@ -159,6 +169,31 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
...
+   * For each removed item the ItemRemovedEvent and for each added item the ItemAddedEvent will be triggered. Favour
+   * this method over using clearItems and adding all items again afterwards.
+   * @param newItems
+   */
+  synchronizeItems(newItems: ListItem[]): void {
+    newItems
+      .filter((item) => !this.hasItem(item.key))
+      .map((item) => this.addItem(item.key, item.label));
```
protyposis: `forEach` (you're not doing anything with the mapping result)

src/ts/components/listselector.ts
stonko1994: Should we keep the FIXME or do we say that this isn't defined so the events are fine? Maybe change the FIXME to a normal comment to just point that out?

protyposis: If this really is purely a matter of taste and does not have any negative impact that should be fixed sooner or later, then change to a normal comment.

If it has some kind of major drawback then keep the fixme.

PR #243, opened 2019-05-21
Feature/fix browserutils

pcanterini:
### Description
When importing this npm package in a server sider rendered app the following error occurs:
`ReferenceError: navigator is not defined`
https://d.pr/i/hPQEY9

Affected versions: `3.4.5` and possibly older versions

I wasn't sure if I should have created this PR against develop instead but just tried to follow the contrib guidelines.

This is obviously a suggestion and _one_ way on how to solve this issue but feel free to close the PR and address it in a different way you guys see fit.

Thanks!

### Problem
`navigator` is not available on the server side even though all we are trying to do is import the package, not instantiate it on the BE.

### Solution
Check if `window` is available before using `navigator` and set a default otherwise.

### Manual tests
Import the package in a node.js app (or server side rendered file). The file should be imported with no errors as long as not instantiated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.4.6]
+
+### Fixed
+- Allow npm package to be imported in server side app without `navigator` error.
+
 ## [3.4.5]
 
 ### Fixed
--- a/src/ts/browserutils.ts
+++ b/src/ts/browserutils.ts
@@ -1,4 +1,6 @@
 export namespace BrowserUtils {
+  // set default value for navigator when using package in nodejs or ssr app
+  const navigator = typeof window === 'undefined' ? {userAgent: ''} : window.navigator;
 
   // isMobile only needs to be evaluated once (it cannot change during a browser session)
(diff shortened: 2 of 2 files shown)


protyposis:
Thanks for submitting this PR. We have been aware of this issue and have worked around it internally when used in a Node environment by setting
```ts
(global as any).navigator = {
  userAgent: 'node.js',
};
```
somewhere before the `BrowserUtils` are imported. This hasn't been fixed until now since the UI is basically designed as a client-side framework and even though your approach would mitigate the immediate problem, I am not entirely sure it is the right one that helps most users.

I am thinking about server-side rendering, where user agent detection is required and done with the agent string from the client. In such case it might make sense to have a parameterized function that initializes the `BrowserUtils`.

Let's see if there are other opinions.

stonko1994:
Thanks for submitting but please create the PR against `develop` and change the version in the changelog to `## [develop]`.

---

Another approach could be to detect the browser lazy when it's needed.

```ts
export class BrowserUtils {
  …
  static get isMobile(): boolean {
    return navigator && navigator.userAgent && /Mobi/.test(navigator.userAgent);
  }
  …
}
```

_If we want to do this check only once we could cache the result._

pcanterini:
Feel free to close my PR and handle it the best way you guys see fit. I’ll look again at this in the morning if needed. Thanks for all the feedback!

stonko1994:
Closing in favour of #244

PR #244, opened 2019-05-22
Fix issues if the UI is used in a non browser environment

stonko1994:
## Description
Our `BrowserUtils` check against the `userAgent` without checking if we are in a browser environment and if the `navigator` object even exists. This lead to issues e.g. when rendering server side.

## Solution
Check the user agent when it's needed and add a check if the window object is defined.

## Tests
Simple test for `isMobile` added


pcanterini:
This is great guys, I really appreciate you looking into it for us ❤️

PR #245, opened 2019-05-22
Add support for es6 inheritance of the player which is passed to the UI

stonko1994:
## Description
When an instance of an inherited player is passed to the UI our `PlayerWrapper` does not wrap all properties of the superclass correctly.

## Solution
Search for `PropertyDescriptor`s also of the prototype of the passed player instance.

### DIFF
--- a/src/ts/uimanager.ts
+++ b/src/ts/uimanager.ts
@@ -763,23 +763,16 @@ class PlayerWrapper {
     for (let property of properties) {
       // Get an eventually existing property descriptor to differentiate between plain properties and properties with
       // getters/setters.
-      const getPropertyDescriptor = (object: PlayerAPI): PropertyDescriptor => {
-        const propertyDescriptor = Object.getOwnPropertyDescriptor(object, property);
-        if (propertyDescriptor) {
-          return propertyDescriptor;
-        } else {
-          const childPrototype = Object.getPrototypeOf(object);
-          if (childPrototype) {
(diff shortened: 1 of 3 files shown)


src/ts/uimanager.ts
```diff
@@ -763,8 +763,23 @@ class PlayerWrapper {
     for (let property of properties) {
       // Get an eventually existing property descriptor to differentiate between plain properties and properties with
       // getters/setters.
-      let propertyDescriptor: PropertyDescriptor = Object.getOwnPropertyDescriptor(player, property) ||
-        Object.getOwnPropertyDescriptor(Object.getPrototypeOf(player), property);
+      const getPropertyDescriptor = (object: PlayerAPI): PropertyDescriptor => {
```
protyposis: `object` is a keyword in TypeScript so better to avoid that (wasn't used when the code was written, but is now).

Also just fyi this would be the iterative non-recursive way:
```ts
const propertyDescriptor = ((target: object) => {
    while (target) {
        const propertyDescriptor = Object.getOwnPropertyDescriptor(target, property);
        if (propertyDescriptor) {
            return propertyDescriptor;
        }
        target = Object.getPrototypeOf(target);
    }
})(player);
```

stonko1994: I prefer the non-recursive way as it's much cleaner.

stonko1994:
@protyposis sry for another review but I was able to add a simple test. Can you please also have a look at it.

PR #249, opened 2019-05-29
Fix generic typing of config objects

### LINKED ISSUE #74: Subclassing button requires casting of config in subclasses
When you extend the button class
```
export class MyButton extends Button<MyButtonConfig>
```
the type of config is ButtonConfig inside your subclass, which requires you to cast it. 
I would expect config to be of type MyButtonConfig. 

That can be achieved by changing the button declaration 
```
//from   
export class Button<Config extends ButtonConfig> extends Component<ButtonConfig> {   
//to    
export class Button<Config extends ButtonConfig> extends Component<Config> {   
```

wdyt?

protyposis:
Removes all casts around `component.getConfig()` calls which used to return a base type instead of the config type of the actual component. Closes #74.

This has been bugging me since writing the very first version of this framework. Now it is finally fixed (except for a few still required type hints in `mergeConfig`) 🎈 🎉🎈🍾🎈🍸

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Improved documentation about time / number formats which can be used in `AdMessageLabel` placeholders
+- TypeScript update to 3.4.5
+- Improved generic type inheritance of `Component` `Config`s ([#74](https://github.com/bitmovin/bitmovin-player-ui/issues/74))
 
 ## [3.5.0]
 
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - `ListBox` no longer recreates itself after the list was updated
+- TypeScript update to 3.4.5
```
stonko1994: This should be under `## [develop]` not under `## [3.5.0]`. Feels like a wrong base branch 🤔

protyposis: Base branch was right but I didn't look at the header :)

PR #258, opened 2019-09-17
TTML multi region

unki2aut:
Allow for different regions and use a `default-subtitle` style for other cases.

### DIFF
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -6,14 +6,20 @@
 
   @include hidden;
 
-  bottom: 2em;
+  bottom: 0;
   font-size: 1.2em;
-  left: 3em;
+  left: 0;
   pointer-events: none;
   position: absolute;
-  right: 3em;
+  right: 0;
   text-align: center;
   transition: bottom $animation-duration-short ease-out;
+  top: 0;
+
+  .#{$prefix}-container-wrapper {
+    width: 100%;
+    height: 100%;
+  }
 
   .#{$prefix}-ui-subtitle-label {
     @include text-border($subtitle-text-border-color);
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -16,7 +16,7 @@
   transition: bottom $animation-duration-short ease-out;
   top: 0;
 
-  &.#{$prefix}-default-subtitle-position {
+  &.#{$prefix}-subtitle-position-default {
     bottom: 2em;
     top: initial;
     left: 3em;
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -16,24 +16,25 @@
   transition: bottom $animation-duration-short ease-out;
   top: 0;
 
-  &.#{$prefix}-subtitle-position-default {
+  .#{$prefix}-subtitle-position-default {
     bottom: 2em;
     top: initial;
     left: 3em;
     right: 3em;
+    position: absolute;
+
+    .#{$prefix}-subtitle-position-default .#{$prefix}-ui-subtitle-label {
+      &:nth-child(1n-1)::after {
+        content: '\A';
+        white-space: pre-line;
+      }
+    }
   }
 
   .#{$prefix}-ui-subtitle-label {
     @include text-border($subtitle-text-border-color);
 
     color: $subtitle-text-color;
-
-    // Break labels into separate lines
-    // sass-lint:disable force-pseudo-nesting
-    &:nth-child(1n-1)::after {
-      content: '\A';
-      white-space: pre-line;
-    }
   }
 
   // Move the subtitle up above the controlbar when it appears to avoid them overlapping
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -23,11 +23,6 @@
     right: 3em;
   }
 
-  .#{$prefix}-container-wrapper {
-    width: 100%;
-    height: 100%;
-  }
-
   .#{$prefix}-ui-subtitle-label {
     @include text-border($subtitle-text-border-color);
 
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -16,6 +16,13 @@
   transition: bottom $animation-duration-short ease-out;
   top: 0;
 
+  &.#{$prefix}-default-subtitle {
+    bottom: 2em;
+    top: initial;
+    left: 3em;
+    right: 3em;
+  }
+
   .#{$prefix}-container-wrapper {
     width: 100%;
     height: 100%;
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -21,20 +21,23 @@
     top: initial;
     left: 3em;
     right: 3em;
-    position: absolute;
+  }
 
-    .#{$prefix}-subtitle-position-default .#{$prefix}-ui-subtitle-label {
-      &:nth-child(1n-1)::after {
-        content: '\A';
-        white-space: pre-line;
-      }
-    }
+  .#{$prefix}-subtitle-region-container {
+    position: absolute;
   }
 
   .#{$prefix}-ui-subtitle-label {
     @include text-border($subtitle-text-border-color);
 
     color: $subtitle-text-color;
+
+    // Break labels into separate lines
+    // sass-lint:disable force-pseudo-nesting
+    &:nth-child(1n-1)::after {
+      content: '\A';
+      white-space: pre-line;
+    }
   }
 
   // Move the subtitle up above the controlbar when it appears to avoid them overlapping
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -16,7 +16,7 @@
   transition: bottom $animation-duration-short ease-out;
   top: 0;
 
-  &.#{$prefix}-default-subtitle {
+  &.#{$prefix}-default-subtitle-position {
     bottom: 2em;
     top: initial;
     left: 3em;
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -16,15 +16,15 @@
   transition: bottom $animation-duration-short ease-out;
   top: 0;
 
-  .#{$prefix}-subtitle-position-default {
-    bottom: 2em;
-    top: initial;
-    left: 3em;
-    right: 3em;
-  }
-
   .#{$prefix}-subtitle-region-container {
     position: absolute;
+
+    &.#{$prefix}-subtitle-position-default {
+      bottom: 2em;
+      top: initial;
+      left: 3em;
+      right: 3em;
+    }
   }
 
   .#{$prefix}-ui-subtitle-label {
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -3,9 +3,9 @@ import {UIInstanceManager} from '../uimanager';
 import {Label, LabelConfig} from './label';
 import {ComponentConfig, Component} from './component';
 import {ControlBar} from './controlbar';
-import {EventDispatcher} from '../eventdispatcher';
+import { EventDispatcher } from '../eventdispatcher';
 import {DOM} from '../dom';
-import {PlayerAPI, SubtitleCueEvent} from 'bitmovin-player';
+import { PlayerAPI, SubtitleCueEvent } from 'bitmovin-player';
 
 /**
  * Overlays the player to display subtitles.
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -98,8 +98,11 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
       this.updateComponents();
     };
 
+    player.on(player.exports.PlayerEvent.AudioChanged, subtitleClearHandler);
     player.on(player.exports.PlayerEvent.SubtitleEnabled, subtitleClearHandler);
     player.on(player.exports.PlayerEvent.SubtitleDisabled, subtitleClearHandler);
+    player.on(player.exports.PlayerEvent.Seek, subtitleClearHandler);
+    player.on(player.exports.PlayerEvent.TimeShift, subtitleClearHandler);
     player.on(player.exports.PlayerEvent.PlaybackFinished, subtitleClearHandler);
     player.on(player.exports.PlayerEvent.SourceUnloaded, subtitleClearHandler);
 
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -326,6 +326,10 @@ class ActiveSubtitleManager {
       text: event.html || ActiveSubtitleManager.generateImageTagText(event.image) || event.text,
     });
 
+    if (event.regionStyle) {
+      label.getDomElement().attr('style', event.regionStyle + ';position:absolute');
+    }
+
     // Create array for id if it does not exist
     this.activeSubtitleCueMap[id] = this.activeSubtitleCueMap[id] || [];
 
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -47,7 +47,7 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
     let subtitleManager = new ActiveSubtitleManager();
     this.subtitleManager = subtitleManager;
 
-    let subtitleContainerManager = new SubtitleContainerManager(this);
+    let subtitleContainerManager = new SubtitleRegionContainerManager(this);
 
     player.on(player.exports.PlayerEvent.CueEnter, (event: SubtitleCueEvent) => {
       // Sanitize cue data (must be done before the cue ID is generated in subtitleManager.cueEnter)
@@ -419,7 +419,7 @@ class ActiveSubtitleManager {
   }
 }
 
-class SubtitleContainerManager {
+class SubtitleRegionContainerManager {
   private subtitleRegionContainers: { [region: string]: SubtitleRegionContainer } = {};
   private subtitleOverlay: SubtitleOverlay;
 
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -460,6 +460,7 @@ export class SubtitleRegionContainerManager {
     }
 
     this.subtitleRegionContainers = {};
+    this.labelsInRegionCount = {};
   }
 
   /**
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -481,9 +481,12 @@ class SubtitleRegionContainerManager {
 }
 
 class SubtitleRegionContainer extends Container<ContainerConfig> {
-  constructor(config: ContainerConfig) {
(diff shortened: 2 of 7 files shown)


src/ts/components/subtitleoverlay.ts
```diff
@@ -92,11 +96,8 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
       this.updateComponents();
     };
 
-    player.on(player.exports.PlayerEvent.AudioChanged, subtitleClearHandler);
```
protyposis: Looking at your commits it looks like the player is firing `PlayerEvent.CueExit` with these removed events? The UI is defined to be backwards compatible with all player releases of a major version - so this works correctly back to 8.0.0.

unki2aut: Right, the intention to remove these callbacks was that the cues get cleaned up by the `PlayerEvent.CueExit` rather than doing it on these actions. This lead to a cue that spans over a longer time range would be cleared although it should still be visible.

protyposis: > This lead to a cue that spans over a longer time range would be cleared although it should still be visible.

You mean when seeking within the queue's timespan? Then this case could specifically be handled. The reason why the callbacks were there was that the cues did not get cleaned up.

src/ts/components/subtitleoverlay.ts
```diff
@@ -326,6 +327,10 @@ class ActiveSubtitleManager {
       text: event.html || ActiveSubtitleManager.generateImageTagText(event.image) || event.text,
     });
 
+    if (event.regionStyle) {
```
stonko1994: As far as I know, there can be multiple cues / regions at once. How are they handled by the player? Are they resulting in different cue events?

src/ts/components/subtitleoverlay.ts
```diff
@@ -36,7 +36,7 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
     this.previewSubtitle = new SubtitleLabel({ text: 'example subtitle' });
 
     this.config = this.mergeConfig(config, {
-      cssClass: 'ui-subtitle-overlay',
+      cssClasses: ['ui-subtitle-overlay', 'default-subtitle'],
```
stonko1994: Can we add positioning into the class name to make visible this is for subtitles which don't come with positioning attributes? `default-subtitle-positioning` or something.

src/ts/components/subtitleoverlay.ts
unki2aut: I thought so. I'll see if I can come up with an improved cleanup strategy.

src/ts/components/subtitleoverlay.ts
unki2aut: Yes, they result in different cues with the same time range but different `region` attribute.

stonko1994: Perfect 👍

src/ts/components/subtitleoverlay.ts
protyposis: `subtitle-position-(default|left|top|...)`?

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -6,14 +6,27 @@
...
+  left: 0;
   pointer-events: none;
   position: absolute;
-  right: 3em;
+  right: 0;
   text-align: center;
   transition: bottom $animation-duration-short ease-out;
+  top: 0;
+
+  &.#{$prefix}-default-subtitle-position {
```
stonko1994: Please rename to `subtitle-position-default` as discussed in the review https://github.com/bitmovin/bitmovin-player-ui/pull/258#discussion_r325537968.

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -6,14 +6,27 @@
...
+    bottom: 2em;
+    top: initial;
+    left: 3em;
+    right: 3em;
+  }
+
+  .#{$prefix}-container-wrapper {
```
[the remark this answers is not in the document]
hanneslinder: You are right, we can actually remove those styles.

src/ts/components/subtitleoverlay.ts
```diff
@@ -54,6 +56,10 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
         event.position.column = event.position.column || 0;
       }
 
+      if (!event.region) {
+        this.getDomElement().addClass(this.prefixCss(SubtitleOverlay.CLASS_DEFAULT_SUBTITILE_POSITION));
```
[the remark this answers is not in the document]
hanneslinder: I added both the add and removal of the subtitle class now here to make it clearer what is happening

src/ts/components/subtitleoverlay.ts
```diff
@@ -54,6 +56,12 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
         event.position.column = event.position.column || 0;
       }
 
+      if (event.region) {
```
protyposis: Why don't you check for `regionStyle` here? Seems odd given that you only use the `regionStyle` propety on the label. Also, does it make sense to have a `region` specified, remove the default positioning, and then not have a `regionStyle` for the label?

src/ts/components/subtitleoverlay.ts
```diff
@@ -326,6 +331,10 @@ class ActiveSubtitleManager {
       text: event.html || ActiveSubtitleManager.generateImageTagText(event.image) || event.text,
     });
 
+    if (event.regionStyle) {
+      label.getDomElement().attr('style', event.regionStyle + ';position:absolute');
```
protyposis: Are you sure that the right way to handle TTML cues is to set the region style on the label itself (in contrast to e.g. putting the labels into a container with the region style)? I'm thinking about multiple labels at the same time in the same region here... with this implementation they would overlap.

src/ts/components/subtitleoverlay.ts
protyposis: This is still an open point to resolve.

src/ts/components/subtitleoverlay.ts
hanneslinder: You are right, checking for `regionStyle` actually makes more sense.

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +418,82 @@ class ActiveSubtitleManager {
...
+  public removeLabel(labelToRemove: SubtitleLabel): void {
+    for (let region in this.subtitleRegionContainers) {
+      this.subtitleRegionContainers[region].removeLabel(labelToRemove);
+    }
+  }
+
+  /**
+   * Removes all subtitle containers.
+   */
+  public clear(): void {
```
stonko1994: Shouldn't that also be called on `cueExit`? What if there is no need for this container anymore? But I'm not sure if the `cueExitEvent` does have the needed information.

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +418,82 @@ class ActiveSubtitleManager {
...
+  constructor(config: ContainerConfig) {
+    super(config);
+  }
+
+  public removeLabel(labelToRemove: SubtitleLabel): void {
+    this.removeComponent(labelToRemove);
+    this.updateComponents();
+  }
+
+  public updateComponents(): void {
```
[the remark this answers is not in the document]
protyposis: I consider this a hack due to non-optimal architecture, so I'd keep it like this for now and not make it public everywhere for everyone.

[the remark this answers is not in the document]
protyposis: Add an `addLabel` method and call `updateComponents` in there - then you can get rid if this hack, right?

src/ts/components/subtitleoverlay.ts
```diff
@@ -46,6 +47,8 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
     let subtitleManager = new ActiveSubtitleManager();
     this.subtitleManager = subtitleManager;
 
+    let subtitleContainerManager = new SubtitleRegionContainerManager(this);
```
protyposis: `const`

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +418,82 @@ class ActiveSubtitleManager {
...
 }
+
+class SubtitleRegionContainerManager {
+  private subtitleRegionContainers: { [region: string]: SubtitleRegionContainer } = {};
+  private subtitleOverlay: SubtitleOverlay;
+
+  /**
+   * @param subtitleOverlay Reference to the subtitle overlay for adding and removing the containers.
+   */
+  constructor(subtitleOverlay: SubtitleOverlay) {
```
protyposis: Change the signature to `private subtitleOverlay` to get rid of the explicit private field definition

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +418,82 @@ class ActiveSubtitleManager {
...
+    this.subtitleOverlay = subtitleOverlay;
+  }
+
+  /**
+   * Removes a subtitle label from a container.
+   * @param labelToRemove
```
protyposis: Remove the `@param` if there is not helpful description

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +418,82 @@ class ActiveSubtitleManager {
...
+   */
+  public removeLabel(labelToRemove: SubtitleLabel): void {
+    for (let region in this.subtitleRegionContainers) {
```
protyposis: `const region`

protyposis: or `const region of this.subtitleRegionContainers` and `region.removeLabel(labelToRemove)`

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +418,82 @@ class ActiveSubtitleManager {
...
+      this.subtitleRegionContainers[region].removeLabel(labelToRemove);
+    }
+  }
+
+  /**
+   * Removes all subtitle containers.
+   */
+  public clear(): void {
+    for (let region in this.subtitleRegionContainers) {
```
protyposis: Here as well:

```ts
for (const region of this.subtitleRegionContainers) {
  this.subtitleOverlay.removeComponent(region);
}
```

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +418,82 @@ class ActiveSubtitleManager {
...
+    }
+
+    this.subtitleRegionContainers = {};
+  }
+
+  /**
+   * Creates and wraps a subtitle label into a container div based on the subtitle region.
+   * If the subtitle has positioning information it is added to the container.
+   * @param label The subtitle label to wrap
+   * @param event Subtitle cue event that contains the positoning information
```
protyposis: Typo

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +418,82 @@ class ActiveSubtitleManager {
...
+        regionContainer.getDomElement().attr('style', event.regionStyle + ';position:absolute');
+      } else {
+        // getDomElement needs to be called at least once to ensure the component exists
+        regionContainer.getDomElement();
+      }
+    }
+
+    this.subtitleRegionContainers[subtitleRegion].addComponent(label);
+
+    for (let region in this.subtitleRegionContainers) {
```
protyposis: `const region`

protyposis: It might be better to call the variable `regionName` (everywhere) to make it obvious that this is the name of the region and not the actual region (container).

protyposis: Also I think you can again do the `for ... of`

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +418,82 @@ class ActiveSubtitleManager {
...
+      this.subtitleOverlay.addComponent(this.subtitleRegionContainers[region]);
+      this.subtitleRegionContainers[region].updateComponents();
```
protyposis: Why do you do this for every region on every `wrap` call? That feels like much more than just wrapping.

src/ts/components/subtitleoverlay.ts
hanneslinder: Could there be a case where multiple subtitles are in one region and only one of the subtitles fires the `cueExit` event?

protyposis: Sounds like a default case to me.

src/ts/components/subtitleoverlay.ts
hanneslinder: `this.subtitleRegionContainers` is currently an object so I cannot use `for... of. I could change it to be a map if you want.

src/ts/components/subtitleoverlay.ts
hanneslinder: You are right, this was not necessary in every `cueEnter`.

src/ts/components/subtitleoverlay.ts
hanneslinder: So it's probably better to not recreate the containers on every cueEnter / cueExit and reuse the existing ones instead.

src/ts/components/subtitleoverlay.ts
protyposis: Right - my bad. Then it's fine of course :)

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +418,81 @@ class ActiveSubtitleManager {
...
+      }
+    }
+
+    this.subtitleRegionContainers[regionName].addLabel(label);
+  }
+}
+
+class SubtitleRegionContainer extends Container<ContainerConfig> {
+  constructor(config: ContainerConfig) {
+    const mergedConfig = {...config, cssClasses: [config.cssClass, 'subtitle-region-container']};
```
stonko1994: We do have a method for merging configurations: `mergeConfig(...)` (See other components for an example)

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -6,14 +6,26 @@
...
+  left: 0;
   pointer-events: none;
   position: absolute;
-  right: 3em;
+  right: 0;
   text-align: center;
   transition: bottom $animation-duration-short ease-out;
+  top: 0;
+
+  .#{$prefix}-subtitle-position-default {
```
stonko1994: Shouldn't that be nested via `&.#{$prefix}-subtitle-position-default {` into the `.#{$prefix}-subtitle-region-container` 🤔 Does it make sense to have this class on something else than a container 🤔

src/ts/components/subtitleoverlay.ts
```diff
@@ -3,9 +3,9 @@ import {UIInstanceManager} from '../uimanager';
 import {Label, LabelConfig} from './label';
 import {ComponentConfig, Component} from './component';
 import {ControlBar} from './controlbar';
-import { EventDispatcher } from '../eventdispatcher';
+import {EventDispatcher} from '../eventdispatcher';
 import {DOM} from '../dom';
-import { PlayerAPI, SubtitleCueEvent } from 'bitmovin-player';
+import {PlayerAPI, SubtitleCueEvent} from 'bitmovin-player';
```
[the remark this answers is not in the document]
stonko1994: still open ^

ioreper:
What's left to address this PR and can it be done in time for Monday's release?

src/ts/components/subtitleoverlay.ts
```diff
@@ -59,34 +63,36 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
       this.preprocessLabelEventCallback.dispatch(event, labelToAdd);
 
       if (this.previewSubtitleActive) {
-        this.removeComponent(this.previewSubtitle);
+        this.subtitleContainerManager.removeLabel(this.previewSubtitle);
       }
-      this.addComponent(labelToAdd);
-      this.updateComponents();
 
+      this.subtitleContainerManager.addLabel(labelToAdd, event);
```
protyposis: Passing the `event` here should be unnecessary... the `labelToAdd = subtitleManager.cueEnter(event);` call above should attach the information directly to the `SubtitleLabel`,

src/ts/components/subtitleoverlay.ts
```diff
@@ -59,34 +63,36 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+
```
protyposis: This empty line seems unnecessary

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +422,97 @@ class ActiveSubtitleManager {
...
+  }
+
+  /**
+   * Creates and wraps a subtitle label into a container div based on the subtitle region.
+   * If the subtitle has positioning information it is added to the container.
+   * @param label The subtitle label to wrap
+   * @param event Subtitle cue event that contains the positoning information
+   */
+  addLabel(label: SubtitleLabel, event?: SubtitleCueEvent): void {
+    const regionName = event && event.region || 'default';
```
protyposis: As noted above, it shouldn't be necessary to provide the `event` here and also in `removeLabel`.

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +422,97 @@ class ActiveSubtitleManager {
...
+    this.labelsInRegionCount = {};
+  }
+
+  /**
+   * Creates and wraps a subtitle label into a container div based on the subtitle region.
+   * If the subtitle has positioning information it is added to the container.
+   * @param label The subtitle label to wrap
+   * @param event Subtitle cue event that contains the positoning information
+   */
+  addLabel(label: SubtitleLabel, event?: SubtitleCueEvent): void {
```
protyposis: Please order the methods - feels odd that `addLabel` comes after `removeLabel` and even after `clear`. Suggested order: `addLabel`, `removeLabel`, `clear`.

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +422,97 @@ class ActiveSubtitleManager {
     this.activeSubtitleCueCount = 0;
   }
 }
+
+export class SubtitleRegionContainerManager {
+  private subtitleRegionContainers: { [regionName: string]: SubtitleRegionContainer } = {};
+  private labelsInRegionCount: { [regionName: string]: number } = {};
```
protyposis: Why doesn't the `SubtitleRegionContainer` itself keep track of the count?

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +422,97 @@ class ActiveSubtitleManager {
...
+   */
+  removeLabel(labelToRemove: SubtitleLabel, event?: SubtitleCueEvent): void {
+    const region = event && event.region || 'default';
+    for (const regionName in this.subtitleRegionContainers) {
+      this.subtitleRegionContainers[regionName].removeLabel(labelToRemove);
+    }
+    this.labelsInRegionCount[region]--;
+
+    // Remove container if no more labels are displayed
+    if (this.labelsInRegionCount[region] === 0) {
```
protyposis: If the `SubtitleRegionContainer` kept count of its labels, this could look like `subtitleRegionContainer.getLabelCount() === 0` or even `subtitleRegionContainer.isEmpty()`

src/ts/components/subtitleoverlay.ts
```diff
@@ -416,3 +436,98 @@ class ActiveSubtitleManager {
...
+  }
+
+  removeLabel(labelToRemove: SubtitleLabel): void {
+    this.labelCount--;
+    this.removeComponent(labelToRemove);
+    this.updateComponents();
+  }
+
+  public isEmpty(): boolean {
+    return this.labelCount === 0;
```
protyposis: You could omit the `labelCount` and check `this.getComponents().length` instead.

hanneslinder: I agree, but unfortunately that breaks the unit tests as I have to mock the container class.

hanneslinder:
> Do we have a sample stream which we can publish via adding it to our `index.html` as source?

I don't have a publicly available asset right now I'm afraid. Any Ideas where I could get one with TTML subtitles?

src/ts/components/subtitleoverlay.ts
protyposis: Ok, I think we can live with that :)

stonko1994:
> Any Ideas where I could get one with TTML subtitles?

Unfortunately no.

PR #265, opened 2019-09-23
Show UI in not playing states

stonko1994:
## Description
In order to improve UX we decided to show the UI per default in the following states:
- Prepared
- Paused
- Finished

As the UI is visible per default now we decided to lower the default `hideDelay` value.

**This works around multiple issues:**
- First tap not starting playback on iOS 13
- UI hides if user is scrubbing longer than the `hideDelay`

### DIFF
--- a/src/ts/components/uicontainer.ts
+++ b/src/ts/components/uicontainer.ts
@@ -243,8 +243,6 @@ export class UIContainer extends Container<UIContainerConfig> {
     uimanager.getConfig().events.onUpdated.subscribe(() => {
       updateState(PlayerUtils.getState(player));
     });
-    // Init in current player state
-    updateState(PlayerUtils.getState(player));
 
     // Fullscreen marker class
     player.on(player.exports.PlayerEvent.ViewModeChanged, () => {
--- a/src/ts/components/uicontainer.ts
+++ b/src/ts/components/uicontainer.ts
@@ -178,7 +178,9 @@ export class UIContainer extends Container<UIContainerConfig> {
     });
     uimanager.onSeeked.subscribe(() => {
       isSeeking = false;
-      this.uiHideTimeout.start(); // Re-enable UI hide timeout after a seek
+      if (!hidingPrevented()) {
+        this.uiHideTimeout.start(); // Re-enable UI hide timeout after a seek
+      }
     });
     player.on(player.exports.PlayerEvent.CastStarted, () => {
       showUi(); // Show UI when a Cast session has started (UI will then stay permanently on during the session)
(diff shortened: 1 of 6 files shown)


src/ts/components/uicontainer.ts
```diff
@@ -53,7 +53,12 @@ export class UIContainer extends Container<UIContainerConfig> {
 
     this.config = this.mergeConfig(config, <UIContainerConfig>{
       cssClass: 'ui-uicontainer',
-      hideDelay: 5000,
+      hideDelay: 2000,
+      hidePlayerStateExceptions: [
```
protyposis: This will impact customers' custom UIs so I'd rather go the safe way and do this in our default UI structure only (and schedule the general change for v4).

src/ts/components/uicontainer.ts
```diff
@@ -236,8 +243,6 @@ export class UIContainer extends Container<UIContainerConfig> {
     uimanager.getConfig().events.onUpdated.subscribe(() => {
       updateState(PlayerUtils.getState(player));
     });
-    // Init in current player state
-    updateState(PlayerUtils.getState(player));
```
protyposis: Why did you remove the initialization?

src/ts/components/uicontainer.ts
stonko1994: > ... in our default UI structure ...

Do you mean only within `UIFactory`?

I thought about the same but decided against as we want to change the default behaviour in all UI variants. But thinking again it would make more sense to not change components behaviour.

src/ts/components/uicontainer.ts
stonko1994: The problem here is that the `updateState` now actually triggers a `showUI` before other components had a chance to subscribe to the `onControlsShow` callback. This happens if we are in a `hidingPreventedState` (Prepared). But within the `onShow` we have a flag (`isUiShown`) which will change into `true` but nothing is actual visible.

Call stack:
- `updateState`
- `stateChanged.dispatch`
- Calls `stateChanged.subscribe` callback in `uicontainer`
- `hidingPrevented` returns `true`
- calls `showUI`
- switches the `isUiShown` flag
- a second call would not trigger the `showUI` again

---

I think it's safe to remove the initialization here because the `uimanager` will fire `onUpdated` after initialization (if player is ready). Or the player will fire an appropriated event when it's ready.

protyposis: Makes sense 👍

src/ts/components/uicontainer.ts
protyposis: Yes that's what I meant :)

stonko1994: But thinking about it a third time maybe it's better to have it here 🤔 
My argument for this is that the value in `UIFactory` would override this one here. So if a customer changed it already within the `UIContainer` he would need to remove it again from the `UIFactory` (all variants). 

On the other side if a customer doesn't want this change he just needs to change it here once 🤔 

Maybe both variants have advantages and disadvantages. Not sure what's the better place to do this 'breaking' change.

PR #271, opened 2019-10-03
Feature/add localization support to ui

Cagince:
This is a draft PR for the Localization Support this was meant to support/visualize the Specification issue that I have created.
Please take a look at the implementation and give feedback so we can close the specification issue and turn this PR into an actual PR... There might be some cases I haven't considered. if you find any you can add it either as a comment or update the list below.

#### TODO:
- [x] allow customer to pass localization Config.
- [x] support static text 
- [x] support dynamic texts `'some value {value}'`
- [x] replace all existing texts (all replaced all that I could find. should be looked again on the PR process...)
- [x] support configuration of fallback languages.
- [x] set language to `default` language of `user` **if language exists in the translations**
- [x] provide a vocabulary template for the customer to edit. 
- [x] Add examples in index.html
- [x] Custom label customization


src/ts/localization/de.json
```diff
@@ -0,0 +1,30 @@
+{
+  "Video Quality": "Videoqualität",
+  "Audio Quality": "Audioqualität",
+  "Audio Track": "Audiospur",
+  "Speed": "Geschwindigkeit",
+  "Play/Pause": "Abspielen/Pause",
+  "open": "öffnen",
+  "Volume/Mute": "Lautstärke/Stummschaltunge",
```
lukaskroepfl:
```suggestion
  "Volume/Mute": "Lautstärke/Stummschaltung",
```

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,117 @@
...
+  private extractVariablesFromTranslationString(translation: string, values: {[key: string]: any} = {}) {
+    const translationVariables = Array.from(translation.match(/{([A-Z]|[a-z])+}/g) || []); // looks for {anyValue}
+
+    return translationVariables
+      .map(match => [match, match.slice(1, -1)])  // extract key for next step
+      .map(([match, key]) => ({ match, value: values[key]}));
+  }
+
+
+  public t(key: string, config: BitmovinPlayerUiTranslationConfig = {}): string {
```
lukaskroepfl: what's the upside of `t` instead of `translated`? as we get the translated word for the path

Cagince:
@stonko1994 thank you for your input. I understand the argument that: ` calling label.setText('Video Quality'), it would get translated because there is a translation for the input string available`.
imho. the concern seems a little unnecessary on this regard because: 
- the *text* that the user passes is a key `Video Quality`(which will also be pointed in the documentation) if the user doesn't want it to be translated he/she easily `does not` add it to `translations` object in the other languages. 
- Also in other case if we put in the config.translations: `{...'Video Quality': 'Hello World'}` `Hello world` will be rendered when we set Text to `Video Quality` my counter argument in this point is: if the customer wants to show 'Video Quality' when they `setText('Video Quality')`, they shouldn't set the value to something else, or if they don't want translation they just don't pass the config...
- The customer also have the ability to set their own keys via config like: `{ 'myOwnVideoQualityText': 'top notch'}`, if they want to translate they can add the 'key' to target language vocabulary or leave it only in english if they don't want to...

> What happens if label.setText is called, and the onTextChanged event fires? Does it convey the input text, or the translated text? Or does setText not translate texts on its own? (it probably should not; i.e. we don't want an unnecessary translation lookup on the playback time label when the time updates)

this I understand the concern in my implementation the translation fn. is quite simple: 
```javascript
 const translate = (key) => vocabulary[key] || key; 
``` 
meaning we only initialize vocabulary once we change the config. after that we read from the *vocabulary* object and return key if nothing can be found. we might also add an additional parameter to config to stop translation from happening but I am not sure if that'd reduce the comp. power to add an additional check but I think it's more expensive to make `if (confg.doNotTranslate) return` each time a text should be translated.
- other solution to this problem would be to wrap every 'string' with `t()` when they are declared example: `new SettingsItem(translator.t('Subtitles'), {....})` and leave  the `label` and `button` components as is. so that we really `force` the translations on the texts that should be translated and any other thing we don't wrap would not call a lookup eg: `timestamp` 
I am not sure about the second discussion but I think you and @protyposis  can give me feedback on if that is really necessary and how that'd effect the customer and us to develop & to maintain. I just don't think we should introduce a complex structure for such simple task.

Cagince:
that is also the reason that I need your feedback on the issue because I might be oversimplifying the feature and might miss some pitfalls or use-cases. it'd be nice to have a conversation about what might go wrong with the implementation and what we can and cannot sacrifice

protyposis:
Here's 3 points I discussed with @stonko1994 and would like to address:

The current architecture makes it a little difficult to integrate localization, which is the reason why it's been delayed that much. We should at least consider going the route of a new major release (v4) with a proper integration of localization (to be discussed in the internal specification issue). Assuming that is not an option:

_No singleton_. I know a singleton is a lot easier to implement in the current architecture, but we really should be supporting multiple UI instances with different configurations, and that includes localization. We've been going this way in all our products since a long time and the UI should not suddenly get to be an exception.

_Translation function_. Let's take the label in its current state in this PR as an example:
https://github.com/bitmovin/bitmovin-player-ui/blob/792e0f7123c7c7214a122c92b9a5046eed3665d8/src/ts/components/admessagelabel.ts#L12-L19
Here the translation takes place in the constructor, which is before the UI is _actually_ initialized (before `configure`). This happens in quite a lot of places and I assume this is one of the reasons for choosing a singleton-approach. I propose to work around this problem in the following way, which also allows us to integrate localization with backwards compatibility: Currently `i18n.t` is called and returns the translated string. Instead, we could use a what's nowadays fancily called "thunk" approach and return a function instead of the actual translated string, and this function is invoked when the DOM is rendered. This way we could leave the initialization in the constructors as they are, but the actual translation will be executed later. The type of every text-property in the components could be extended from type `string` to type `string | TranslationThunk` (not the final name). If it's a basic string, no translation is applied at all. A similar approach has been suggested somewhere, what's new is the "thunk".

_String Keys_. We have discussed that already, using string keys or using "pure strings" for translation lookups, with the example of changing casing in a string from "Video Quality" to "Video quality". @Cagince prefers "pure strings" with the argument that it's the developer's responsibility to take care of correct mappings, and if we were to change the casing he also needs to take care of changing the mapping. While this is a valid approach, I think this introduces a pathway for probable oversights because this requires extended knowledge of the framework and a lot of discipline - it would be much better to offload this to a fitting architecture or the language itself (TypeScript types). Of course we could lowercase everything for the lookup to fix this specific issue, but that is not the only one. We came to the conclusion that the key-based approach would be much cleaner. Assuming we have multiple translations, a changed letter case does not have to be changed in 11 places, but only in one. Changing a key would be much simpler because it will be unique and not a grammatically correct expression, so a simple search and replace will likely work without replacing in unexpected places like comments. But even better, we can create a typed map which can give us type checking (e.g. is the used key in the label really existing) and IDE's will be able to provide auto-completion and refactoring functionality. Also I think it would me cleaner to also have the "default" strings in the same place like any other translation (e.g. a specific folder).

src/ts/audiotrackutils.ts
```diff
@@ -46,7 +47,7 @@ export class AudioTrackSwitchHandler {
   private addAudioTrack = (event: AudioTrackEvent) => {
     const audioTrack = event.track;
     if (!this.listElement.hasItem(audioTrack.id)) {
-      this.listElement.addItem(audioTrack.id, audioTrack.label);
+      this.listElement.addItem(audioTrack.id, getLocalizedText(i18n.t(audioTrack.label)));
```
protyposis: This is unnecessarily complicated, why two function calls? This should simply be `i18n.something(audioTrack.label)` instead of `getLocalizedText(i18n.t(audioTrack.label))` (applies in other places too).

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,31 @@
+import I18nApi, {  LocalizationConfig, BitmovinPlayerUiTranslations } from './i18nApi';
+import vocabularyDe from './de.json';
+import vocabularyEn from './en.json';
+
+
+type LocalizableCallback = () => string;
+type LocalizableText = string | LocalizableCallback;
+const getLocalizedText = (text: LocalizableText) => typeof text === 'function' ? text() : text;
+
+const defaultBitmovinPlayerUiTranslations: BitmovinPlayerUiTranslations = {
```
protyposis: Simply call it `defaultTranslations`

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,31 @@
+import I18nApi, {  LocalizationConfig, BitmovinPlayerUiTranslations } from './i18nApi';
+import vocabularyDe from './de.json';
+import vocabularyEn from './en.json';
+
+
+type LocalizableCallback = () => string;
+type LocalizableText = string | LocalizableCallback;
+const getLocalizedText = (text: LocalizableText) => typeof text === 'function' ? text() : text;
```
protyposis: This should be a hidden internal and not exposed to users.

src/ts/localization/i18nApi.ts
```diff
@@ -0,0 +1,161 @@
+import { defaultBitmovinPlayerUiTranslations } from './i18n';
+
+interface StaticVocabulary {
```
protyposis: This should simply be called `Vocabulary`, because this is the only kind of vocabulary we have.

src/ts/localization/i18nApi.ts
```diff
@@ -0,0 +1,161 @@
...
+export type Vocabulary<V> =  V & Partial<StaticVocabulary>;
+type StringKeyMap = {[key: string]: string};
+
+export interface BitmovinPlayerUiTranslations {
+  [key: string]: Vocabulary<StringKeyMap>;
+}
+
+export interface LocalizationConfig {
+  language: string;
+  fallbackLanguages?: string[];
```
protyposis: Why is `language` a single language, and `fallbackLanguages` an array? Why not the other way round?

I think it should be the other way round because I'd like to have a fallback that is supposedly the full set of vocabulary, e.g. english. Lets assume I have base line English `en` and partial British English `en-uk`. Then I have a default Portuguese translation `pt` and a partial Brazilian Portuguese `pt-br` where also a subset of terms is overwritten.

Now I create a configuration for British English:

```ts
{
  language: 'en-uk',
  fallbackLanguages: ['en'],
}
```

Now I want to switch the language to Brazilian: how would I switch to this?

```ts
{
  language: 'pt-br',
  fallbackLanguages: ['pt', 'en'],
}
```

Is it the responsibility of the user to create a merged vocabulary `{ ...pt, ...pt-br }`? If so, what do we have the fallback array at all?

I think this should be discussed.

src/ts/localization/localizationTemplate.json
```diff
@@ -0,0 +1,30 @@
+{
```
protyposis: Since we have JSON files for English and German, the template should not be necessary?

src/ts/subtitleutils.ts
```diff
@@ -84,7 +85,7 @@ export class SubtitleSwitchHandler {
 
     const offListItem: ListItem = {
       key: SubtitleSwitchHandler.SUBTITLES_OFF_KEY,
-      label: 'off',
+      label: getLocalizedText(i18n.t('labels.off')),
```
protyposis: I think the call of `getLocalizedText` should be the responsibility of the "renderer", i.e. `Component.toDomElement`.

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,31 @@
...
+
+
+const i18n = new I18nApi(defaultLocalizationConfig);
+
+export {
+  i18n,
+  defaultBitmovinPlayerUiTranslations,
+  defaultLocalizationConfig,
+  LocalizableText,
+  getLocalizedText,
```
protyposis: I don't like the exposure of a pure function, as this is not a pattern used in the UI framework (there is no other place doing this. Either move it to a class (optionally static), or into a namespace.

src/ts/localization/de.json
```diff
@@ -0,0 +1,37 @@
+{
```
stonko1994: Can we move the `*.json` files into an additional subfolder called `languages` or something?

src/ts/localization/de.json
```diff
@@ -0,0 +1,37 @@
...
+  "settings.video.quality": "Videoqualität",
+  "settings.audio.quality": "Audioqualität",
+  "settings.audio.track": "Audiospur",
+  "settings.speed": "Geschwindigkeit",
+  "settings.playPause": "Abspielen/Pause",
+  "settings.open": "öffnen",
```
stonko1994: could be moved to `common`

src/ts/localization/de.json
```diff
@@ -0,0 +1,37 @@
...
+  "settings.audio.mute": "Lautstärke/Stummschaltung",
```
stonko1994: Why is this under `settings`

src/ts/localization/de.json
```diff
@@ -0,0 +1,37 @@
...
+  "labels.pictureInPicture": "Bild im Bild",
+  "labels.appleAirplay": "Apple AirPlay",
+  "labels.googleCast": "Google Cast",
+  "labels.vr": "VR",
```
stonko1994: Why are those under `labels`? These are actual buttons.

src/ts/localization/de.json
```diff
@@ -0,0 +1,37 @@
...
+  "settings.subtitles.font.size": "Größe",
+  "settings.subtitles.font.family": "Schriftart",
+  "settings.subtitles.font.color": "Farbe",
+  "settings.subtitles.font.opacity": "Deckkraft",
+  "settings.subtitles.characterEdge": "Ränder",
+  "settings.subtitles.background.color": "Hintergrundfarbe",
+  "settings.subtitles.background.opacity": "Hintergrunddeckkraft",
+  "settings.window.color": "Hintergrundfarbe",
+  "settings.window.opacity": "Hintergrunddeckkraft",
+  "labels.back": "Zurück",
```
stonko1994: Could be moved to `common`

src/ts/localization/de.json
```diff
@@ -0,0 +1,37 @@
...
+  "labels.reset": "Zurücksetzen",
```
stonko1994: Could be moved to `common`

src/ts/localization/de.json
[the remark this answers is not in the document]
Cagince: it's a good idea and also taught about it. I think mario disagreed on this part. also as a personal opinion : it would make things much more complicated if we want to *merge* the languages together, and also getting a translation would require a lot of *checks* if the property even exists ...etc. at this point I don't really see the benefit of doing so other than it being *better* for the eye

src/ts/localization/i18nApi.ts
```diff
@@ -0,0 +1,161 @@
...
+      if (Object.keys(translations).findIndex((k) => k === userLanguage) !== -1) {
+        this.language = userLanguage;
+        return;
+      }
+    }
+
+    this.language = language;
+  }
+
+  private initializeFallbackLanguages(translations: BitmovinPlayerUiTranslations, fallbackLanguages?: string[]) {
```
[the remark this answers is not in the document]
Cagince: agreed

src/ts/localization/i18nApi.ts
```diff
@@ -0,0 +1,161 @@
...
+
+export type Vocabulary<V> =  V & Partial<StaticVocabulary>;
+type StringKeyMap = {[key: string]: string};
+
+export interface BitmovinPlayerUiTranslations {
+  [key: string]: Vocabulary<StringKeyMap>;
+}
+
+export interface LocalizationConfig {
+  language: string;
```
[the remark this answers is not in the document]
Cagince: might be a good idea. We should have a discussion about it tough because we currently have only 'en' and 'de' and as far as I know we are not planning to provide additional languages.
another possibility would be to have `language: 'de' | 'en' | string` as typedef. and user can add their language or can see what we currently have?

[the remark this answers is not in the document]
stonko1994: True we don't have much 🤔 An enum might be an overkill.

src/ts/uimanager.ts
```diff
@@ -302,6 +306,14 @@ export class UIManager {
     resolveUiVariant(null);
   }
 
+  /**
+   * Exposes i18n to user
+   * @returns {I18nApi}
+   */
+  get i18n() {
```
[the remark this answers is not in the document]
Cagince: You are right on that. but instead of `getLocalizedText` we should provide `i18n.t` function. doing it now

src/html/index.html
```diff
@@ -249,6 +249,15 @@ <h4 class="card-title">
 //      {title: 'Recommendation 11', url: 'http://bitmovin.com', thumbnail: 'http://placehold.it/300x300', duration: 435},
 //      {title: 'Recommendation 12: Ain\'t no better video than this', url: 'http://bitmovin.com', thumbnail: 'http://placehold.it/300x300', duration: 34}
 //    ],
+    localization: {
```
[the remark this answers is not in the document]
Cagince: by returning a `function` from `i18n.t()` to be called on `component.toDomElement()`  we solve that problem.

src/ts/localization/de.json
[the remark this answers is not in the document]
Cagince: :+1: then we should agree on an interface with correct names, ..etc. before I do the change. 
should we add it to specification issue ?

spec/localization.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+
+    it('uses vocabulary \'it\'', () => {
+      i18n.setConfig({ ...defaultConfig, language: 'it' });
+      expect(i18n.getLocalizedText(i18n.t('test'))).toEqual(successIt);
+    });
+  });
+
+  describe('Language Fallback\'s', () => {
+    it('falls back to `key` if it is not in vocabulary', () => expect(i18n.getLocalizedText(i18n.t('some word'))).toEqual('some word'));
```
stonko1994: Break into normal test syntax as this line is too long. We have 120 characters

src/ts/components/audioqualityselectbox.ts
```diff
@@ -29,7 +30,8 @@ export class AudioQualitySelectBox extends SelectBox {
       this.clearItems();
 
       // Add entry for automatic quality switching (default setting)
-      this.addItem('auto', 'auto');
+      this.addItem('auto', i18n.t('labels.off'));
+
```
stonko1994: Remove extra empty line

src/ts/components/button.ts
```diff
@@ -1,15 +1,16 @@
 import {ComponentConfig, Component} from './component';
 import {DOM} from '../dom';
 import {EventDispatcher, NoArgs, Event} from '../eventdispatcher';
+import { LocalizableText , i18n} from '../localization/i18n';
```
stonko1994: Add space in `{ … }`
```suggestion
import { LocalizableText , i18n } from '../localization/i18n';
```

src/ts/components/listselector.ts
```diff
@@ -103,8 +104,8 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
    * @param key the key of the item to add
    * @param label the (human-readable) label of the item to add
    */
-  addItem(key: string, label: string) {
-    const listItem = { key: key, label: label };
+  addItem(key: string, label: LocalizableText) {
+    const listItem = { key: key, label: i18n.getLocalizedText(label)};
```
stonko1994: Add space `{ … }`
```suggestion
    const listItem = { key: key, label: i18n.getLocalizedText(label) };
```

src/ts/localization/i18nApi.ts
```diff
@@ -0,0 +1,161 @@
...
+  fallbackLanguages?: string[];
+  disableBrowserLanguageDetection?: boolean;
+  translations: BitmovinPlayerUiTranslations;
+}
+
+export default class I18nApi {
```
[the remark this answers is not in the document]
stonko1994: Still open ^

src/ts/localization/i18nApi.ts
```diff
@@ -0,0 +1,170 @@
...
+      let translationString = this.vocabulary[key as string | number];
+
+      if (translationString == null) {
+        console.warn(`We haven't been able to find a translation provided for key: '${key}'... The value of the key will be set to '${key}'.\n Please provide correct value via 'config' if this was not intended.`);
+        translationString = key as string;
+      }
+      return translationString;
+    };
+  }
+
```
stonko1994: Remove empty line

src/ts/localization/i18nApi.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  public getLocalizedText(text: LocalizableText) {
+    return typeof text === 'function' ? text() : text;
+  }
+
```
stonko1994: Remove empty line

src/ts/localization/languages/en.json
```diff
@@ -0,0 +1,64 @@
+{
```
stonko1994: Placeholder comment for previous discussion about format. https://github.com/bitmovin/bitmovin-player-ui/pull/271#discussion_r333871498

src/ts/uiconfig.ts
```diff
@@ -40,7 +41,7 @@ export interface TimelineMarker {
 export interface UIConfig {
   /**
    * Specifies the container in the DOM into which the UI will be added. Can be a CSS selector string or a
-   * HTMLElement object. By default, the player container will be used ({@link PlayerAPI#getContainer}).
+   * HTMLElement object. By default, the player container will be used ({@link PlayerAPI#getContainer})."
```
stonko1994: Remove `"` 
```suggestion
   * HTMLElement object. By default, the player container will be used ({@link PlayerAPI#getContainer}).
```

src/ts/localization/i18nApi.ts
```diff
@@ -0,0 +1,170 @@
...
+export interface BitmovinPlayerUiTranslations {
+  [key: string]: CustomVocabulary<StringKeyMap>;
+}
+
+export interface LocalizationConfig {
+  language: 'en' | 'de' | string;
+  fallbackLanguages?: string[];
+  disableBrowserLanguageDetection?: boolean;
+  translations: BitmovinPlayerUiTranslations;
```
stonko1994: Why isn't that optional?

src/ts/localization/i18nApi.ts
```diff
@@ -0,0 +1,170 @@
...
+
+export type CustomVocabulary<V> = V & Partial<Vocab>;
+type StringKeyMap = {[key: string]: string};
+
+export interface BitmovinPlayerUiTranslations {
+  [key: string]: CustomVocabulary<StringKeyMap>;
+}
+
+export interface LocalizationConfig {
+  language: 'en' | 'de' | string;
```
stonko1994: Why isn't that optional?

src/ts/localization/i18nApi.ts
```diff
@@ -0,0 +1,170 @@
...
+}
+
+export type CustomVocabulary<V> = V & Partial<Vocab>;
+type StringKeyMap = {[key: string]: string};
+
+export interface BitmovinPlayerUiTranslations {
+  [key: string]: CustomVocabulary<StringKeyMap>;
+}
+
+export interface LocalizationConfig {
```
stonko1994: Add a short documentation to this interface.

stonko1994: Can we move this ConfigInterface to `UIConfig.ts` file

src/ts/uiconfig.ts
```diff
@@ -67,4 +68,9 @@ export interface UIConfig {
    * For an example have a look at {@link ErrorMessageOverlayConfig.messages}
    */
   errorMessages?: ErrorMessageMap | ErrorMessageTranslator;
+  /**
+   * Provide configuration to support Multiple UI languages
+   * default language: 'en'
```
stonko1994: Actually the default languages is the browser language.

stonko1994:
I did a bit of manual testing. Doing something like that will not work :/

```ts
let myLabel = new Label({ text: 'foobar.text' });
myLabel.getDomElement();

...

new UIManager…
```

Cagince:
@stonko1994 
This is why we exposed the `translate` function of `UiManager` so that the user will have to use `UiManager.translate('my.key')` if they want to be able to translate a certain key... meaning that the `toDomElement()` fn should be called after the UManager is initialized.
Currently adding the correct `usage` in the Readme.md... 
One thing comes to ming regarding your finding tough. Should we also make the `translate` function of `UIManager` a **static** function??

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,182 @@
...
+  private initializeLanguage(
+    language: string,
+    disableBrowserLanguageDetection: boolean,
+    translations: BitmovinPlayerUiTranslations,
+  ) {
+    const shouldDetectLanguage = !Boolean(disableBrowserLanguageDetection);
+
+    if (shouldDetectLanguage) {
+      let userLanguage = (window.navigator.language);
+      userLanguage = userLanguage.slice(0, 2);
```
protyposis: Language names are not limited to 2 chars, what if it's called `de-at`?

Cagince: it is true, I have been thinking about it after your comment about the multi-language & fallback languages. 
the problem I encountered is there are too many languages and too many variations. and basically if you are localizing a website you localize it for `de` not `de-at` or `en` not `en-us` (theoretically)  for example, for english we have:
```
| English - Australia | en | en-au |
| English - Belize | en | en-bz |
| English - Canada | en | en-ca |
| English - Caribbean | en | en-cb |
| English - Great Britain | en | en-gb | 
| English - India | en | en-in |
| English - Ireland | en | en-ie |
| English - Jamaica | en | en-jm |
| English - New Zealand | en | en-nz |
| English - Philippines | en | en-ph |
| English - Southern Africa | en | en-za |
| English - Trinidad | en | en-tt |
| English - United States | en | en-us |
| English - Zimbabwe | en |
```
also depending on the browser return value can be `en-AU` or `en-au` or just `en`...
should user add the same translation for all or should we have a map containing all of the language maps? 
that is why I taught using just the first `2` chars of the language is the safest way to go and for each [language variation first 2 is same anyways](https://www.science.co.il/language/Locale-codes.php)... 

this supporting the de-at would also create confusion (atleast for me )

user-defined-vocabularies: 'de-DE', 'en-US', 'it'
default-language: 'en-US'
browser-language: 'de-AT'
fallbackLangs in order: 'it', 'de-DE' 
in this case we will not set language to `de-DE` because browser finds `de-AT` and in our vocabulary there is no such thing. so the language will be set to `en-US` even tough the `de` is available...

Cagince: so my suggestion is to standardize this somehow. and `recommend` user to use those standards to get expected results`

protyposis: The "shortnames" are already some kind of standard. Simply check for the full string, e.g. `en-us`, and fallback to `en` otherwise.

src/ts/localization/i18nApi.ts
Cagince: regarding this comment I agree that it is the responsibility of the user to create a merged vocabulary. Current behavior that we provide is: `en` is always the fallback because that is the only language with full static vocabulary atm. (unless user duplicates the en.json and fill everything). 
what we are doing if user defines 'en' for example we merge it with default english like `{...defaultEn, ...userEn}`.

and same goes for fallback languages, basically if user defined 'de', 'it', 'fr' on the fallback languages the merge will be `{...defaultEn, ...defaultDe, ...userFr, ...userIt, ... userDe}`. why is this important?, user might have multiple languages defined in vocabulary, fallback languages are there so that we know in which order we merge them together. if they do not exist, we merge them anyways but using `Object.keys(languages)` which then does not ensure the order of the keys is same as it is defined in the `object`...

why language is single?. because there can be only one language. (maybe I should rename it to defaultLanguage to make it more clear?). why fallBackLanguages multiple? because we give the flexibility to define the order in which fallback languages will be set(maybe I should rename it to fallbackLanguageOrder to make it more clear?)

protyposis: I completely understand what you mean. It seems to me thought that if merging is the responsibility of the user then it should be sufficient to have one single complete set of vocabulary as fallback, and the `fallbackLanguages` array seems unnecessary at all.

Cagince: :+1: you are correct. I will be removing the `fallback languages` then. 
new behavior will be: `{...defaultEn, ...userLanguage}`

README.md
```diff
@@ -240,6 +240,69 @@ const myUiManager = new UIManager(player, myUi, myUiConfig);
...
+      'my.custom.key': 'my custom key',
+      'this will also act as a key': 'my custom key 2',
+    }
+    de: {
+      'my.custom.key': 'my german translation',
+    }
+  }
+};
+
+const label1 = new Label({ text: UIManager.localize('my.custom.key')});
```
stonko1994: Keep code style also in the readme:
```suggestion
const label1 = new Label({ text: UIManager.localize('my.custom.key') });
```

src/ts/components/subtitlesettings/backgroundopacityselectbox.ts
```diff
@@ -18,12 +19,12 @@ export class BackgroundOpacitySelectBox extends SubtitleSettingSelectBox {
...
     super.configure(player, uimanager);
 
-    this.addItem(null, 'default');
-    this.addItem('100', '100%');
-    this.addItem('75', '75%');
-    this.addItem('50', '50%');
-    this.addItem('25', '25%');
-    this.addItem('0', '0%');
+    this.addItem(null, i18n.t('default'));
+    this.addItem('100', i18n.t('percent', { value: 100}));
```
stonko1994: Spaces within `{ … }` (All occurences)
```suggestion
    this.addItem('100', i18n.t('percent', { value: 100 }));
```

src/ts/components/subtitlesettings/characteredgeselectbox.ts
```diff
@@ -18,11 +19,11 @@ export class CharacterEdgeSelectBox extends SubtitleSettingSelectBox {
   configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
     super.configure(player, uimanager);
 
-    this.addItem(null, 'default');
-    this.addItem('raised', 'raised');
-    this.addItem('depressed', 'depressed');
-    this.addItem('uniform', 'uniform');
-    this.addItem('dropshadowed', 'drop shadowed');
+    this.addItem(null, i18n.t('default')); //  'default'
```
stonko1994: remove comment `// 'default'`

src/ts/components/subtitlesettings/fontopacityselectbox.ts
```diff
@@ -18,11 +19,11 @@ export class FontOpacitySelectBox extends SubtitleSettingSelectBox {
   configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
     super.configure(player, uimanager);
 
-    this.addItem(null, 'default');
-    this.addItem('100', '100%');
-    this.addItem('75', '75%');
-    this.addItem('50', '50%');
-    this.addItem('25', '25%');
+    this.addItem(null, i18n.t('default'));
+    this.addItem('100', i18n.t('percent', { value: 100}));
```
stonko1994: Spaces

src/ts/components/subtitlesettings/fontsizeselectbox.ts
```diff
@@ -18,14 +19,14 @@ export class FontSizeSelectBox extends SubtitleSettingSelectBox {
...
-    this.addItem(null, 'default');
-    this.addItem('50', '50%');
-    this.addItem('75', '75%');
-    this.addItem('100', '100%');
-    this.addItem('150', '150%');
-    this.addItem('200', '200%');
-    this.addItem('300', '300%');
-    this.addItem('400', '400%');
+    this.addItem(null, i18n.t('default'));
+    this.addItem('50', i18n.t('percent', {value: 50}));
```
stonko1994: Spaces

src/ts/components/subtitlesettings/windowopacityselectbox.ts
```diff
@@ -18,12 +19,12 @@ export class WindowOpacitySelectBox extends SubtitleSettingSelectBox {
...
     super.configure(player, uimanager);
 
-    this.addItem(null, 'default');
-    this.addItem('100', '100%');
-    this.addItem('75', '75%');
-    this.addItem('50', '50%');
-    this.addItem('25', '25%');
-    this.addItem('0', '0%');
+    this.addItem(null, i18n.t('default'));
+    this.addItem('100', i18n.t('percent', {value: 100}));
```
stonko1994: spaces

src/ts/uimanager.ts
```diff
@@ -10,6 +10,20 @@ import {BrowserUtils} from './browserutils';
...
+import { i18n, CustomVocabulary, BitmovinPlayerUiTranslations } from './localization/i18n';
+
+/**
+ * language: defines the preferred locale.
+ * fallbackLanguages[]: Array of languages which should be used as fallback languages if given word doesn't exist
+ * in the selected vocabulary (in the order they are given in the array.)
+ * browserLanguageDetection: (default: true) will  enable/disable auto-detection and selection of browsers preferred language
+ * translations: key-value pair of 'language' and 'vocabulary' definitions. this is where the user adds their custom languages/vocabularies
+ */
+export interface LocalizationConfig {
```
stonko1994: Please write the documentation of the properties directly above the properties as we do in other places as well. A general documentation for the interface can be above the interface.

Docu is outdated too.

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+    this.initializeLanguage(config.language, browserLanguageDetection, translations);
+    this.initializeVocabulary(translations);
+  }
+
+  private containsKey(obj: object, key: string) {
+    return obj.hasOwnProperty(key);
+  }
+
+  private mergeTranslationsWithDefaultTranslations(translations: BitmovinPlayerUiTranslations = {}) {
+    const rawTranslations: BitmovinPlayerUiTranslations = { ...defaultTranslations, ...translations};
```
stonko1994: spaces

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+    return Object.keys(rawTranslations).reduce((acc, key) => {
+      let translation: CustomVocabulary<Record<string, string>> = rawTranslations[key as string];
+      if (this.containsKey(defaultTranslations, key) && this.containsKey(translations, key)) {
+        translation = {...defaultTranslations[key], ...translations[key]};
```
stonko1994: spaces

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+      }
+      return { ...acc, [key]: translation };
+    }, {} as BitmovinPlayerUiTranslations);
```
stonko1994: is `as BitmovinPlayerUiTranslations` necessary?

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  private initializeLanguage(
+    language: string,
+    browserLanguageDetectionEnabled: boolean,
+    translations: BitmovinPlayerUiTranslations,
+  ) {
+
```
stonko1994: remove empty line

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+    if (browserLanguageDetectionEnabled) {
+      const userLanguage = (window.navigator.language);
```
stonko1994: Why surrounding with brackets?

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+      }
+    }
+
+    this.language = language;
+
+  }
+
+  private initializeVocabulary(translations: BitmovinPlayerUiTranslations) {
+    this.vocabulary = ['en', this.language]
+      .reduce((vocab, lang) => ({...vocab, ...(translations[lang] || {})}), {});
```
stonko1994: spaces

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+      }
+      const shortenedUserLanguage = userLanguage.slice(0, 2);
+      if (translations.hasOwnProperty(shortenedUserLanguage)) {
+        this.language = shortenedUserLanguage;
+        return;
+      }
+    }
+
+    this.language = language;
+
```
stonko1994: remove empty line

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+  private replaceVariableWithPlaceholderIfExists(text: string, config: any) {
+    const matches = Array.from(text.match(new RegExp('{[a-zA-Z]+}')));
+    if (matches.length === 0) {
+      return text;
+    }
+
+    return matches
+      .map((m: string) => ({ match: m, key: m.slice(1, -1)}))
+      .reduce((str, {key, match}) => config.hasOwnProperty(key) ? str.replace(match, config[key]) : str, text);
```
stonko1994: spaces

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  public t<V extends CustomVocabulary<Record<string, string>> = CustomVocabulary<Record<string, string>>>(key: keyof V, config?: any) {
+    return () => {
+      if (key == null) { // because sometimes we call toDomElement() without configuring the component or setting text...
+        return undefined;
+      }
+      let translationString = this.vocabulary[key as string | number];
+
+      if (translationString == null) {
+        // console.warn(`We haven't been able to find a translation provided for key: '${key}'... The value of the key will be set to '${key}'.\n Please provide correct value via 'config' if this was not intended.`);
```
stonko1994: either remove comment or add the warning

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+      return translationString;
+    };
+  }
+
+  public getLocalizedText(text: LocalizableText) {
+    return typeof text === 'function' ? text() : text;
+  }
+}
+
+export const i18n = new I18n(defaultLocalizationConfig);
```
stonko1994: add newline at EOF

src/ts/localization/languages/de.json
```diff
@@ -0,0 +1,37 @@
...
+  "replay": "Wiederholen",
+  "ads.remainingTime": "Diese Anzeige endet in {remainingTime} Sekunden",
+  "default": "standard",
+  "colors.white": "weiß",
+  "colors.black": "schwarz",
+  "colors.red": "rot",
+  "colors.green": "grün",
+  "colors.blue": "blau",
+  "colors.yellow": "gelb"
+}
```
stonko1994: Add newline at EOF

src/ts/localization/languages/en.json
```diff
@@ -0,0 +1,53 @@
...
+  "googleCast": "Google Cast",
+  "vr" : "VR",
+  "off": "off",
+  "auto": "auto",
+  "back" : "Back",
+  "reset": "Reset",
+  "replay": "Replay",
+  "normal": "normal",
+  "default": "default"
+}
```
stonko1994: add newline at EOF

src/ts/uiconfig.ts
```diff
@@ -7,6 +7,7 @@ export interface UIRecommendationConfig {
   duration?: number;
 }
 
+
```
stonko1994: remove empty line

README.md
```diff
@@ -240,6 +240,69 @@ const myUiManager = new UIManager(player, myUi, myUiConfig);
...
 
+### UI Localization
+UI can be localized by passing `localization` object to *uiConfig* while initializing UIManager. By default the player UI supports `english` and `german` and will set the language to `browsers preferred language` by default (if language has a vocabulary defined) if this behavior is not intended, you can easily disable it by adding `disableBrowserLanguageDetection: true` parameter to your Localization Config...
+
+```ts
+interface LocalizationConfig {
+  language?: 'en' | 'de' | string;
+  fallbackLanguages?: string[];
+  disableBrowserLanguageDetection?: boolean;
+  translations?: BitmovinPlayerUiTranslations;
```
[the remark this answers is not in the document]
protyposis: Apart from the previous open question here, do we even need the `translations` property? Why don't we make 'language: string | Vocabulary' - if it is a `string` we try to load a JSON file with this name, if it is a `Vocabulary` object then there's nothing to do.

src/html/index.html
```diff
@@ -253,6 +253,8 @@ <h4 class="card-title">
 
   var player;
   var uiManager;
+  bitmovin.playerui.UIManager.setLocalizationConfig({ language: 'de', browserLanguageDetection: false});
```
protyposis: Demo should be in English

src/ts/uimanager.ts
```diff
@@ -302,6 +316,22 @@ export class UIManager {
     resolveUiVariant(null);
   }
 
+  /**
+   * Exposes i18n.t() function
+   * @returns {I18nApi.t()}
+   */
+  static localize<V extends CustomVocabulary<{[key: string]: string}>>(key: keyof V) {
```
protyposis: Use `Record` type

src/ts/uiconfig.ts
protyposis: Unnecessary blank line

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+      .map((m: string) => ({ match: m, key: m.slice(1, -1)}))
+      .reduce((str, {key, match}) => config.hasOwnProperty(key) ? str.replace(match, config[key]) : str, text);
+  }
+
+  public t<V extends CustomVocabulary<Record<string, string>> = CustomVocabulary<Record<string, string>>>(key: keyof V, config?: any) {
+    return () => {
+      if (key == null) { // because sometimes we call toDomElement() without configuring the component or setting text...
+        return undefined;
+      }
+      let translationString = this.vocabulary[key as string | number];
```
protyposis: Why `| number`? Isn't our vocabulary defined as having only string keys?

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+    if (matches.length === 0) {
+      return text;
+    }
+
+    return matches
+      .map((m: string) => ({ match: m, key: m.slice(1, -1)}))
+      .reduce((str, {key, match}) => config.hasOwnProperty(key) ? str.replace(match, config[key]) : str, text);
+  }
+
+  public t<V extends CustomVocabulary<Record<string, string>> = CustomVocabulary<Record<string, string>>>(key: keyof V, config?: any) {
```
protyposis: Can `config` really be `any`? Shouldn't it be `Record<string, string | number>`?

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  }
+
+  private initializeVocabulary(translations: BitmovinPlayerUiTranslations) {
+    this.vocabulary = ['en', this.language]
+      .reduce((vocab, lang) => ({...vocab, ...(translations[lang] || {})}), {});
+  }
+
+  private replaceVariableWithPlaceholderIfExists(text: string, config: any) {
+    const matches = Array.from(text.match(new RegExp('{[a-zA-Z]+}')));
```
protyposis: No digits allowed?

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+class I18n {
+  private language: string;
+  private vocabulary: CustomVocabulary<Record<string, string>>;
+
+  constructor(config: LocalizationConfig) {
+    this.setConfig(config);
+  }
+
+  public setConfig(config: LocalizationConfig) {
+    const browserLanguageDetection = config.browserLanguageDetection != null ? config.browserLanguageDetection : true;
```
protyposis:
```ts
config = {
  browserLanguageDetection: true,
  ...config,
};
```

src/ts/components/button.ts
```diff
@@ -1,15 +1,16 @@
 import {ComponentConfig, Component} from './component';
 import {DOM} from '../dom';
 import {EventDispatcher, NoArgs, Event} from '../eventdispatcher';
+import {LocalizableText , i18n} from '../localization/i18n';
```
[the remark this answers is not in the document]
Cagince: why do we have spaces in this import when we do not have any of the imports above?

src/ts/components/listselector.ts
```diff
@@ -1,13 +1,14 @@
 import {Component, ComponentConfig} from './component';
 import {EventDispatcher, Event} from '../eventdispatcher';
 import {ArrayUtils} from '../arrayutils';
+import {LocalizableText, i18n} from '../localization/i18n';
```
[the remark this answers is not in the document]
Cagince: same here ?

src/ts/localization/i18n.ts
Cagince: imo. we should not overwrite fn. parameter, even if it makes it easier for our current situation

README.md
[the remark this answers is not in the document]
Cagince: the `translations` there so that user can add multiple languages. 
yes we set only one vocabulary from the `language` user sets but we also check `browserPrefferedLang` so if the user has already added the language we can set it. so imo that doesn't makes sense to have `string | Vocabulary` which than would make localization very limited...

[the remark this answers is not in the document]
Cagince: for the prev. comment. we only use `Localization` and `translations` 
which makes sense because the config is made for `Localization` and the `translations` object is contains the `translations` for the languages. other option would be to renam it to `languageKeyVocabularyPair`

src/ts/uimanager.ts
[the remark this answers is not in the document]
Cagince: @stonko1994 reolved?

src/ts/localization/i18n.ts
protyposis: This has zero side effects, but if you want to play save then do it like this:

```ts
const mergedConfig = {
  browserLanguageDetection: true,
  ...config,
};
```

README.md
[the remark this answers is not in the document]
protyposis: Localization is usually concerned with more aspects that just pure translation that we do here, e.g. number/time formatting, right-to-left. But sure we should not limit our possibilities with a name too narrow.

src/ts/localization/i18n.ts
Cagince: yes it can be done but why do we need to merge it like that? I am basically assigning a default value to variable if it is not user defined. not sure what the downside of  ` ? : ` is

src/html/index.html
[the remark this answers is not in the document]
Cagince: resolved?

src/ts/localization/i18n.ts
protyposis: It is not pretty and difficult to read.

README.md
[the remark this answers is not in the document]
protyposis: > for the prev. comment. we only use `Localization` and `translations`

`i18n`

src/ts/localization/i18n.ts
protyposis: Actually `browserLanguageDetection: true` should be moved to `defaultLocalizationConfig` and then:

```ts
const mergedConfig = {
  ...defaultLocalizationConfig,
  ...config,
};
```

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
+import vocabularyDe from './languages/de.json';
+import vocabularyEn from './languages/en.json';
+import { LocalizationConfig } from '../uimanager.js';
+
+export const defaultTranslations: BitmovinPlayerUiTranslations = {
+  'en': vocabularyEn,
+  'de': vocabularyDe,
+};
+
+export const defaultLocalizationConfig: LocalizationConfig = {
```
protyposis: Why is this exported?

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,173 @@
...
+
+      if (config != null) {
+        translationString = this.replaceVariableWithPlaceholderIfExists(translationString, config);
+      }
+
+      return translationString;
+    };
+  }
+
+  public getLocalizedText(text: LocalizableText) {
```
protyposis: As this is not really a getter but a function that does some kind of transformation of input to output, it might be better called `localize`!?

src/ts/uimanager.ts
```diff
@@ -302,6 +316,22 @@ export class UIManager {
     resolveUiVariant(null);
   }
 
+  /**
+   * Exposes i18n.t() function
+   * @returns {I18nApi.t()}
+   */
+  static localize<V extends CustomVocabulary<Record<string, string>>>(key: keyof V) {
+    return i18n.t(key);
+  }
```
protyposis: I find it really confusing that there are `t`, `localize`, and `getLocalizedText` functions, even more so that `localize` actually calls `t`. These names should be streamlined.

src/ts/components/button.ts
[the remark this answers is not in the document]
stonko1994: We want to change the style to this but we don't want to change the whole codebase at once.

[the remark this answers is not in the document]
Cagince: imo. that kind of changes should be handled in an seperate issue. changing the code style part by part causes more inconsistency. 
I have already created an [issue ](https://github.com/bitmovin/player-issues-web/issues/2523) for such things, maybe add it as a comment there?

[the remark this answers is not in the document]
protyposis: This is simply the way how we have handled this until today. This is an open source repo so this may seem odd but we want to keep changes (and merge conflicts in forks) to a minimum.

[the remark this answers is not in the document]
Cagince: updated

README.md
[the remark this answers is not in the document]
Cagince: i18n is a standard naming in this domain (by standard I mean used by most of the people who does localization), which represents the process and planning of i(nternationalizatio)n process which includes Localization & Translation of words, sentences ..etc.. If we want to look at the semantics it makes sense to call the process(class) `i18n` and to include `localization` and `translation`s inside that class.
As I checked the libraries this was the namings they used and that was the reason I choose to name it i18n.It can also be renamed to `Localization` if you think it might cause confusion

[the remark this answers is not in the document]
Cagince: but in this case I find it more weird to call `localization.localize` or `localization.getLocalizableCallback` ?

src/ts/uimanager.ts
stonko1994: still open ^

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    return matches
+      .map((m: string) => ({ match: m, key: m.slice(1, -1) }))
+      .reduce((str, { key, match }) => config.hasOwnProperty(key) ? str.replace(match, config[key]) : str, text);
+  }
+
+  public getLocalizableCallback<V extends CustomVocabulary<Record<string, string>> = CustomVocabulary<Record<string, string>>>(
+    key: keyof V,
+    config?: Record<string, string | number>,
+    ) {
```
stonko1994: this `) {` should be same intended as the word `public`

src/ts/localization/languages/en.json
```diff
@@ -0,0 +1,56 @@
...
+  "off": "off",
+  "auto": "auto",
+  "back" : "Back",
+  "reset": "Reset",
+  "replay": "Replay",
+  "normal": "normal",
+  "default": "default",
+  "live": "Live",
+  "subtitle.example": "example subtitle",
+  "playingOn": "Playing on <strong>{castDeviceName}</strong>"
```
stonko1994: German translation is missing for these two

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,178 @@
...
+  'playPause': string;
+  'live': string;
+  'subtitle.example': string;
+  'playingOn': string;
+  'connectingTo': string;
+}
+
+export type CustomVocabulary<V> = V & Partial<Vocabulary>;
+
+export interface BitmovinPlayerUiTranslations {
```
protyposis: Can we please simply call this `Translations`? The prefix does not make sense - it should be either everywhere or nowhere. Even better, call it `LanguageMap` or simply `Vocabularies`!

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,178 @@
+import vocabularyDe from './languages/de.json';
+import vocabularyEn from './languages/en.json';
+import { LocalizationConfig } from '../uimanager.js';
+
+export const defaultTranslations: BitmovinPlayerUiTranslations = {
```
protyposis: If the type id changed to `Vocabularies`, please call this `defaultVocabularies`

src/ts/localization/i18n.ts
```diff
@@ -0,0 +1,178 @@
...
+  'de': vocabularyDe,
+};
+
+const defaultLocalizationConfig: LocalizationConfig = {
+  language: 'en',
+  browserLanguageDetection: true,
+  translations: defaultTranslations,
+};
+
+type LocalizableCallback = () => string;
```
protyposis: This is not a callback, because it does not "call back" to anywhere. It is simply a function that creates an output. It is also not a "localizable" because we cannot execute any action on it to localize it, i.e. it does not have a `setLocale` or equivalent. Please rename to `Localizer` - in the sense that this is a function that does emit localized output, and the action it does it to localize a string - so it "localizes" and is therefore a "localizer".

Please also rename the method that generates it from `getLocalizableCallback` to `getLocalizer`.

src/ts/uimanager.ts
```diff
@@ -302,6 +318,22 @@ export class UIManager {
     resolveUiVariant(null);
   }
 
+  /**
+   * Exposes i18n.getLocalizableCallback() function
+   * @returns {I18nApi.getLocalizableCallback()}
+   */
+  static localize<V extends CustomVocabulary<Record<string, string>>>(key: keyof V) {
```
protyposis: If this is simply exposing a method from another class, why is it called totally different? E.g. for `i18n.getLocalizableCallback` I would expect this to be called `getLocalizableCallback` or `getI18nLocalizableCallback` (or any other fitting name depending on the renames).

Cagince: yes correct, my taught process was that: 
- user should not be using `i18n.localize` function because it is not exposed via uiManager
- when user declares a new Label it feels more natural and less confusing (imo.) if the do `new Label({ text: uiManager.localize('myString')})` instead of  `new Label({ text: uiManager.getLocalizer('myString')})` because what user wants to do is to localize, not to get  `Localizer` to use somewhere else.

again the decision is yours @protyposis , if my thinking is wrong in that regard please lmk. so that I can update it too.

protyposis: Yes `uiManager.localize` feels much better than `uiManager.getLocalizer`, the user shouldn't have to care about implementation technicalities. I'd really like to call it `localize`, but we have another `localize` method - one returns the `Localizer` and one executes it - that is highly confusing. So how about calling this one `uiManager.localize` (it calls `i18n.getLocalizer`) and renaming the "internal" `i18n.localize` into `i18n.performLocalization`?

Cagince: :+1:  yes good idea.

src/ts/uimanager.ts
```diff
@@ -10,6 +10,22 @@ import {BrowserUtils} from './browserutils';
...
+
+export interface LocalizationConfig {
+  /**
+   * defines the preferred locale
+   */
+  language?: 'en' | 'de' | string;
+  /**
+   * (default: true) will  enable/disable auto-detection and selection of browsers preferred language
+   */
+  browserLanguageDetection?: boolean;
```
protyposis: I think these two should be exclusive. If I set `language: 'de'`, I want my UI to be in German, the additional need to set `browserLanguageDetection: false` seems like it should not be necessary - also when I choose German, it would be unexpected if the language detection chooses another language instead.

Maybe this could be solved by removing `browserLanguageDetection` and adding an 'auto' option to `language`?

protyposis: Better yet, we introduce a `language: 'auto'` setting which is also the default, i.e. if the `language` property is missing in the config, `'auto'` will be set as default.

bitmovin-kenny:
Bad news: this broke IE since `Array.from` is not supported:

```exception
"TypeError: Das Objekt unterstützt die Eigenschaft oder Methode "from" nicht
   at e.prototype.replaceVariableWithPlaceholderIfExists (https://cdn.bitmovin.com/player/web/nightly/bitmovinplayer-ui.js:16:4023)
   at Anonymous function (https://cdn.bitmovin.com/player/web/nightly/bitmovinplayer-ui.js:16:4352)
   at e.prototype.performLocalization (https://cdn.bitmovin.com/player/web/nightly/bitmovinplayer-ui.js:16:4482)
   at t.prototype.addItem (https://cdn.bitmovin.com/player/web/nightly/bitmovinplayer-ui.js:12:8970)
   at t.prototype.configure (https://cdn.bitmovin.com/player/web/nightly/bitmovinplayer-ui.js:14:14443)
   at Anonymous function (https://cdn.bitmovin.com/player/web/nightly/bitmovinplayer-ui.js:17:11713)
   at n (https://cdn.bitmovin.com/player/web/nightly/bitmovinplayer-ui.js:17:14014)
   at n (https://cdn.bitmovin.com/player/web/nightly/bitmovinplayer-ui.js:17:14108)
   at n (https://cdn.bitmovin.com/player/web/nightly/bitmovinplayer-ui.js:17:14108)
   at n (https://cdn.bitmovin.com/player/web/"
```

PR #276, opened 2019-10-23
Handle durationChanged event

stonko1994:
### Description
With player `v8.19.0` the `DurationChanged` event was introduced which will be fired when a live stream turned into a VoD. In that case, the player still returns the wall-clock-timestamps for `getCurrentTime`. Therefore we need to respect the `seekableRange` to properly translate the timestamps in a user-friendly value and back.

### Solution
This PR will handle the event accordingly and respects the `seekableRange` whenever needed.

### Tests
- New unit tests added
- For manual testing you can use any dash vod and following network API to fake the live part.

```js
  let availabilityStartTime;
  const streamDuration = 210;
  let switchToVod = false;
  config = {
  ...
    network: {
      preprocessHttpResponse: (type, response) => {
        availabilityStartTime = availabilityStartTime || Date.now() - streamDuration * 1000;

        return switchToVod
          ? Promise.resolve(response)
          : convertVodToLive(type, response, availabilityStartTime);
      },
    },
  ...
  };

  function convertVodToLive(type, response, availabilityStartTime = 0) {
    if (type === 'manifest/dash') {
      const mpdDocument = deserializeMpd(response.body);
      const mpd = mpdDocument.getElementsByTagName('MPD')[0];
      const availabilityStartTimeString = new Date(availabilityStartTime).toISOString();

      mpd.setAttribute('type', 'dynamic');
      mpd.setAttribute('availabilityStartTime', availabilityStartTimeString);
      mpd.setAttribute('publishTime', availabilityStartTimeString);
      mpd.setAttribute('minimumUpdatePeriod', 'PT1S');
      mpd.removeAttribute('mediaPresentationDuration');

      response.body = serializeMpd(mpdDocument);
    }

    return Promise.resolve(response);
  }

  function deserializeMpd(mpd) {
    return new DOMParser().parseFromString(mpd, 'text/xml');
  }

  function serializeMpd(mpd) {
    var xmlHeader = '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>\n';
    return (
      xmlHeader +
      (mpd.documentElement.outerHTML ? mpd.documentElement.outerHTML : new XMLSerializer().serializeToString(mpd))
    );
  }

// peroform the switch manually by calling:
switchToVod = true;
```

### DIFF
--- a/src/ts/components/seekbar.ts
+++ b/src/ts/components/seekbar.ts
@@ -349,7 +349,7 @@ export class SeekBar extends Component<SeekBarConfig> {
       const maxTimeShift = this.player.getMaxTimeShift();
       this.player.timeShift(maxTimeShift - (maxTimeShift * (percentage / 100)), 'ui');
     } else {
-      const seekableRangeStart = this.player.getSeekableRange() && this.player.getSeekableRange().start || 0;
+      const seekableRangeStart = PlayerUtils.getSeekableRangeStart(this.player, 0);
       const relativeSeekTarget = this.player.getDuration() * (percentage / 100);
       const absoluteSeekTarget = relativeSeekTarget + seekableRangeStart;
       this.player.seek(absoluteSeekTarget, 'ui');
--- a/src/ts/components/seekbar.ts
+++ b/src/ts/components/seekbar.ts
@@ -171,7 +171,7 @@ export class SeekBar extends Component<SeekBarConfig> {
         this.setBufferPosition(100);
       }
       else {
-        let playbackPositionPercentage = 100 / player.getDuration() * player.getCurrentTime();
+        let playbackPositionPercentage = 100 / player.getDuration() * this.getRelativeCurrentTime();
 
         let videoBufferLength = player.getVideoBufferLength();
         let audioBufferLength = player.getAudioBufferLength();
@@ -391,7 +391,7 @@ export class SeekBar extends Component<SeekBarConfig> {
       currentTimeSeekBar += currentTimeUpdateDeltaSecs;
 
       try {
-        currentTimePlayer = player.getCurrentTime();
+        currentTimePlayer = this.getRelativeCurrentTime();
       } catch (error) {
         // Detect if the player has been destroyed and stop updating if so
         if (error instanceof player.exports.PlayerAPINotAvailableError) {
@@ -426,7 +426,7 @@ export class SeekBar extends Component<SeekBarConfig> {
 
     let startSmoothPlaybackPositionUpdater = () => {
       if (!player.isLive()) {
-        currentTimeSeekBar = player.getCurrentTime();
+        currentTimeSeekBar = this.getRelativeCurrentTime();
         this.smoothPlaybackPositionUpdater.start();
       }
     };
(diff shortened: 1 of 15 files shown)


src/ts/playerutils.ts
```diff
@@ -78,15 +92,10 @@ export namespace PlayerUtils {
 
   /**
    * Detects changes of the stream type, i.e. changes of the return value of the player#isLive method.
-   * Normally, a stream cannot change its type during playback, it's either VOD or live. Due to bugs on some
-   * platforms or browsers, it can still change. It is therefore unreliable to just check #isLive and this detector
-   * should be used as a workaround instead.
    *
-   * Known cases:
+   * @deprecated use `PlayerEvent.DurationChanged` instead
```
protyposis: Please do not forget that the UI is always backwards compatible to all player releases of the major version.

src/ts/playerutils.ts
```diff
@@ -109,10 +118,15 @@ export namespace PlayerUtils {
...
       // HLS live detection workaround for Android:
       // Also re-evaluate during playback, because that is when the live flag might change.
-      // (Doing it only in Android Chrome saves unnecessary overhead on other plattforms)
+      // (Doing it only in Android Chrome saves unnecessary overhead on other platforms)
       if (BrowserUtils.isAndroid && BrowserUtils.isChrome) {
         player.on(player.exports.PlayerEvent.TimeChanged, liveDetector);
       }
+
+      // DurationChanged event was introduced with player v8.19.1
+      if (player.exports.PlayerEvent.hasOwnProperty('DurationChanged')) {
```
protyposis: This can simply be `if (player.exports.PlayerEvent.DurationChanged) {`

src/ts/playerutils.ts
stonko1994: Yes I'm aware of that but I wasn't sure how to write that in a case that it should be treated as deprecated for the next major player version 🤔 

Actually just removing the `@deprecated` should be sufficient or? (When keeping the TODO below)

src/ts/playerutils.ts
stonko1994: ed50331

src/ts/playerutils.ts
protyposis: I would keep all the description, add a sentence that the stream type can indeed switch since version X, make it `@deprecated since vX.Y.Z use `PlayerEvent.DurationChanged` instead` and keep the TODO

stonko1994: 5c5c04f

src/ts/components/seekbar.ts
```diff
@@ -349,7 +349,10 @@ export class SeekBar extends Component<SeekBarConfig> {
       const maxTimeShift = this.player.getMaxTimeShift();
       this.player.timeShift(maxTimeShift - (maxTimeShift * (percentage / 100)), 'ui');
     } else {
-      this.player.seek(this.player.getDuration() * (percentage / 100), 'ui');
+      const seekableRangeStart = this.player.getSeekableRange() && this.player.getSeekableRange().start || 0;
```
jmsn: Did you ever experienced `this.player.getSeekableRange()` to be `null`/`undefined`?

Also, as we have 3 occurrences of this line we might consider putting it into ulils?

stonko1994: When using it together with our mobile SDKs.

stonko1994: c667ec2

PR #281, opened 2019-11-13
Fix live asset

stonko1994:
### Description
The live asset which is currently included is outdated.

### Fix
Update the live asset.


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - UI flickering back and forth on live streams after ads
+- Not playing live asset sample
```
protyposis: "Not playing" sounds odd. Call it "Outdated live asset sample"

PR #282, opened 2019-11-13
Add web accessibility support

vzablock:
Description: Added web accessibility support for the player ui.

Implementation done using following test cases and examples: 
http://oaa-accessibility.org/example/1/
https://www.w3.org/WAI/GL/WCAG20/tests/test1.html


src/ts/components/playbacktimelabel.ts
```diff
@@ -172,16 +174,20 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
...
+        this.getDomElement().attr('aria-label', `Video, total time: ${totalTime}`);
         break;
       case PlaybackTimeLabelMode.CurrentAndTotalTime:
         this.setText(`${currentTime} / ${totalTime}`);
+        this.getDomElement().attr('aria-label', `Video, current and total time: ${currentTime} / ${totalTime}`);
         break;
       case PlaybackTimeLabelMode.RemainingTime:
         let remainingTime = StringUtils.secondsToTime(durationSeconds - playbackSeconds, this.timeFormat);
         this.setText(`${remainingTime}`);
+        this.getDomElement().attr('aria-label', `Video, remaining time: ${remainingTime}`);
```
Cagince: we can also localize those `aria-label` descriptions? I think it makes sense

src/ts/components/button.ts
```diff
@@ -31,12 +31,18 @@ export class Button<Config extends ButtonConfig> extends Component<Config> {
   }
 
   protected toDomElement(): DOM {
-    // Create the button element with the text label
-    let buttonElement = new DOM('button', {
+    let buttonElementAttributes: {[name: string]: string} = {
```
protyposis: Please used `const` in all changed lines where applicable.

src/ts/components/seekbar.ts
```diff
@@ -115,6 +124,7 @@ export class SeekBar extends Component<SeekBarConfig> {
       cssClass: 'ui-seekbar',
       vertical: false,
       smoothPlaybackPositionUpdateIntervalMs: 50,
+      'aria-controls': this.config["aria-controls"],
```
protyposis: We use single quotes everywhere.

src/ts/components/button.ts
stonko1994:
```suggestion
    let buttonElementAttributes: { [name: string]: string } = {
```

src/ts/components/label.ts
```diff
@@ -40,8 +45,11 @@ export class Label<Config extends LabelConfig> extends Component<Config> {
   }
 
   protected toDomElement(): DOM {
-    let labelElement = new DOM('span', {
+    const tagName = this.config.for != null ? 'label' : 'span';
```
stonko1994: To handle also `undefined` values
```suggestion
    const tagName = Boolean(this.config.for) ? 'label' : 'span';
```

src/ts/components/playbacktimelabel.ts
```diff
@@ -119,6 +119,8 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
     let playbackTimeHandler = () => {
       if (!live && player.getDuration() !== Infinity) {
         this.setTime(PlayerUtils.getCurrentTimeRelativeToSeekableRange(player), player.getDuration());
+      } else {
+        this.getDomElement().removeAttr('aria-label');
```
stonko1994: Why do we remove it in case of `live` 🤔 

Please move this to another place as this is executed every `TimeChanged`.

EDIT: Do you need to remove it every time (in live) because the value is changing? (Can we update it)

src/ts/components/seekbar.ts
```diff
@@ -30,6 +31,12 @@ export interface SeekBarConfig extends ComponentConfig {
    * Default: 50 (50ms = 20fps).
    */
   smoothPlaybackPositionUpdateIntervalMs?: number;
+
+  /**
+   * WCAG20 standard - denotes which elements in a page an interactive element or set of elements has control
+   * over and affects them.
+   */
+  'aria-controls'?: string;
```
stonko1994: Why do we change the naming of the property here? Our style would be `ariaControls`

src/ts/components/seekbar.ts
```diff
@@ -567,6 +591,20 @@ export class SeekBar extends Component<SeekBarConfig> {
...
+
+      if (this.sliderType === 'live') {
+        currentSeekValue = Math.floor(this.getOffset(e) * this.player.getMaxTimeShift());
+      } else if (this.sliderType === 'vod') {
+        currentSeekValue = StringUtils.secondsToTime(Math.floor(this.getOffset(e) * this.player.getDuration()));
+      } else {
+        currentSeekValue = Math.floor(this.getOffset(e) * 100);
+      }
+
+      this.seekBar.attr('aria-valuenow', currentSeekValue.toString());
```
stonko1994: `aria-valuenow` seems od for a seekbar 🤔

src/ts/components/seekbar.ts
```diff
@@ -577,7 +615,10 @@ export class SeekBar extends Component<SeekBarConfig> {
       this.setSeekPosition(targetPercentage);
       this.setPlaybackPosition(targetPercentage);
       this.onSeekPreviewEvent(targetPercentage, true);
+
+      setAriaSliderValue(e);
```
stonko1994: Is it intended that the value changes when the user just hovers the bar?

Why exactly does this value mean? The current position? Or also preview values?

src/ts/components/seekbarlabel.ts
```diff
@@ -130,6 +130,7 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
    */
   setText(text: string) {
     this.timeLabel.setText(text);
+    this.timeLabel.getDomElement().attr('aria-label', `Thumbnail time: ${text}`);
```
stonko1994: I'm not sure if these texts are defined somewhere.
```suggestion
    this.timeLabel.getDomElement().attr('aria-label', `Thumbnail at time: ${text}`);
```

src/ts/components/settingspanelitem.ts
```diff
@@ -38,7 +38,7 @@ export class SettingsPanelItem extends Container<ContainerConfig> {
       if (label instanceof Component) {
         this.label = label;
       } else {
-        this.label = new Label({text: label} as LabelConfig);
+        this.label = new Label({text: label, for: this.setting.getConfig().id} as LabelConfig);
```
stonko1994:
```suggestion
        this.label = new Label({ text: label, for: this.setting.getConfig().id } as LabelConfig);
```

src/ts/components/settingspanelpageopenbutton.ts
```diff
@@ -10,12 +10,15 @@ export class SettingsPanelPageOpenButton extends SettingsPanelPageNavigatorButto
     this.config = this.mergeConfig(config, {
       cssClass: 'ui-settingspanelpageopenbutton',
       text: i18n.getLocalizer('open'),
+      role: 'img'
```
stonko1994:
```suggestion
      role: 'img',
```

src/ts/components/volumetogglebutton.ts
```diff
@@ -16,6 +16,7 @@ export class VolumeToggleButton extends ToggleButton<ToggleButtonConfig> {
       text: i18n.getLocalizer('settings.audio.mute'),
       onClass: 'muted',
       offClass: 'unmuted',
+      role: 'img'
```
stonko1994:
```suggestion
      role: 'img',
```

src/ts/components/watermark.ts
```diff
@@ -18,6 +18,8 @@ export class Watermark extends ClickOverlay {
     this.config = this.mergeConfig(config, {
       cssClass: 'ui-watermark',
       url: 'http://bitmovin.com',
+      role: 'img',
+      text: 'bitmovin-logo'
```
stonko1994:
```suggestion
      text: 'bitmovin-logo',
```

stonko1994: In a forked UI it does not need to be the `bitmovin-logo`

src/ts/uifactory.ts
```diff
@@ -90,10 +90,12 @@ export namespace UIFactory {
       text: i18n.getLocalizer('open'),
     });
 
+    const subtitleSelectBox = new SubtitleSelectBox();
+
     mainSettingsPanelPage.addComponent(
       new SettingsPanelItem(
-        new SubtitleSettingsLabel({text: i18n.getLocalizer('settings.subtitles'), opener: subtitleSettingsOpenButton}),
-        new SubtitleSelectBox(),
+        new SubtitleSettingsLabel({text: i18n.getLocalizer('settings.subtitles'), opener: subtitleSettingsOpenButton, for: subtitleSelectBox.getConfig().id}),
```
stonko1994:
```suggestion
        new SubtitleSettingsLabel({
          text: i18n.getLocalizer('settings.subtitles'), 
          opener: subtitleSettingsOpenButton, 
          for: subtitleSelectBox.getConfig().id,
        }),
```

src/ts/components/label.ts
vzablock: `!= null` is falsy for `null` and `undefined` values, `!== null` is falsy only for `null`

src/ts/components/seekbar.ts
vzablock: Seekbar is a slider in a sense and for web accessibility purposes we need to show the user what is the current value he is on

src/ts/components/seekbar.ts
vzablock: Defines only the current position so users with screen readers would know where they are located on the slider

src/ts/components/watermark.ts
vzablock: We should have a text content inside of the watermark button for the accessibility validation. Not sure what text we should put here

src/ts/components/label.ts
stonko1994: True. Missed the missing second `=` 👍

src/ts/components/watermark.ts
stonko1994: just `logo`?

src/ts/components/seekbar.ts
stonko1994: What I was thinking about is assume we have our default seekbar during a VOD (0 - 100). The `currentTime` is about 10 but the user now hovers over the seekbar at position 50. What should this value be now? 10 where we are currently playing, or 50 where the user hovers over 🤔

vzablock: Only if the slider is pressed and then moved, the current value will be displayed

stonko1994: 👍

src/ts/components/seekbar.ts
```diff
@@ -48,6 +50,60 @@ export interface SeekBarMarker {
   duration?: number;
 }
 
+enum SeekBarType {
```
stonko1994: Can we extract this whole code block into a separate file? Maybe a `SeekBarArrowKeyController` or something. (`seekbar.ts` is already really big (+1.1k LOC))

src/ts/components/seekbar.ts
```diff
@@ -48,6 +50,60 @@ export interface SeekBarMarker {
...
+}
+
+interface ArrowKeyControls {
+  left: () => void;
+  right: () => void;
+  up: () => void;
+  down: () => void;
+}
+
+const arrowKeyControls = (currentValue: number, set: (number: number) => any, range: {min: number, max: number}): ArrowKeyControls => {
```
stonko1994: this line is too long. When extracting this can be a function then we could break the line.

src/ts/components/seekbar.ts
```diff
@@ -48,6 +50,60 @@ export interface SeekBarMarker {
...
+  const controlValue = Math.floor(currentValue);
+
+  return {
+    left: () => {
+      if (controlValue > range.min) {
+        set(controlValue - 1);
```
stonko1994: Who defines those steps? `1` and `5`. Can we extract this and maybe make them configurable. I'm not sure if needed but if you have a loooooong VOD, with 1 and 5 sec steps you can seek for a long time.

src/ts/components/seekbar.ts
stonko1994: `set` is maybe not an ideal name 🤔 What about `applyFunction` / `valueUpdateFunction`

src/ts/components/volumeslider.ts
```diff
@@ -30,6 +30,8 @@ export class VolumeSlider extends SeekBar {
     this.config = this.mergeConfig(config, <VolumeSliderConfig>{
       cssClass: 'ui-volumeslider',
       hideIfVolumeControlProhibited: true,
+      'aria-controls': 'volume',
```
stonko1994: Where is this config applied?

src/ts/uiutils.ts
```diff
@@ -21,4 +21,133 @@ export namespace UIUtils {
     // Walk and configure the component tree
     recursiveTreeWalker(component);
   }
+
+  // From: https://github.com/nfriend/ts-keycode-enum/blob/master/Key.enum.ts
+  export enum KeyDownKey {
```
stonko1994: Do we need all of those? 🤔 Currently, we just care about the arrow keys right?
naming suggestion
```suggestion
  export enum KeyCode {
```

stonko1994:
Manual testing:
- Seekbar does not have a focus outline
- UI not becoming visible when pressing a key
- settingsPanel does not get focus

Will those issues be addressed in a separate PR?

vzablock:
@stonko1994 The other related issues will be adressed in this PR, I just wanted to finish the issue with the seekbar controls before moving to other stuff

stonko1994:
@vzablock maybe do it in a PR which builds on this one to have smaller PRs which are easier to review

src/scss/skin-modern/components/_seekbar.scss
```diff
@@ -150,6 +150,10 @@ $seekbar-height: .3125em;
       }
     }
   }
+
+  &:focus {
+    box-shadow: 0 0 0 2px rgba(27,127,204,.8);
```
stonko1994:
```suggestion
    box-shadow: 0 0 0 2px rgba(27, 127, 204, .8);
```

stonko1994: Please extract to our `_variables.scss` file

src/ts/uiutils.ts
```diff
@@ -21,4 +21,16 @@ export namespace UIUtils {
...
+
+  // From: https://github.com/nfriend/ts-keycode-enum/blob/master/Key.enum.ts
+  export enum KeyCode {
+    LeftArrow = 37,
+    UpArrow = 38,
+    RightArrow = 39,
+    DownArrow = 40,
+
+    Space = 32,
+    End = 35,
```
stonko1994: What is the `End` and `Home` key? 🤔

src/ts/uifactory.ts
```diff
@@ -313,7 +324,7 @@ export namespace UIFactory {
         new Container({
           components: [
             new PlaybackTimeLabel({ timeLabelMode: PlaybackTimeLabelMode.CurrentTime, hideInLivePlayback: true }),
-            new SeekBar({ smoothPlaybackPositionUpdateIntervalMs: -1 }),
+            new SeekBar({ smoothPlaybackPositionUpdateIntervalMs: -1  }),
```
stonko1994:
```suggestion
            new SeekBar({ smoothPlaybackPositionUpdateIntervalMs: -1 }),
```

src/ts/components/seekbar.ts
```diff
@@ -9,6 +9,8 @@ import TimeShiftAvailabilityChangedArgs = PlayerUtils.TimeShiftAvailabilityChang
 import LiveStreamDetectorEventArgs = PlayerUtils.LiveStreamDetectorEventArgs;
 import { TimelineMarker } from '../uiconfig';
 import { PlayerAPI, PlayerEventBase } from 'bitmovin-player';
+import { StringUtils } from '../stringutils';
+import {SeekBarType, setSeekBarControls} from './seekbarcontrols';
```
stonko1994:
```suggestion
import { SeekBarType, setSeekBarControls } from './seekbarcontrols';
```

src/ts/uiutils.ts
vzablock: Set the slider to the min and max value, required by the standard

src/scss/skin-modern/components/_controlbar.scss
```diff
@@ -7,6 +7,8 @@
   @include hidden-animated;
   @include layout-align-bottom;
 
+  visibility: unset !important;
```
stonko1994: Is there any impact adding this? Maybe also for potential users who forked the UI.

stonko1994: Do we need the `!important` here 🤔

src/ts/uifactory.ts
```diff
@@ -215,13 +224,22 @@ export namespace UIFactory {
...
       text: i18n.getLocalizer('open'),
     });
 
+    const subtitleSelectBox = new SubtitleSelectBox();
+
     mainSettingsPanelPage.addComponent(
       new SettingsPanelItem(
-        new SubtitleSettingsLabel({text: i18n.getLocalizer('settings.subtitles'), opener: subtitleSettingsOpenButton}),
-        new SubtitleSelectBox(),
+          new SubtitleSettingsLabel({
```
stonko1994: This is one indentation too much.

src/scss/skin-modern/_variables.scss
```diff
@@ -12,6 +12,8 @@ $color-background: #111 !default;
 $color-background-highlight: transparentize(mix($color-black, $color-highlight, 75%), .3) !default;
 $color-background-bars: transparentize($color-black, .3) !default;
```
protyposis: Why this empty line? Why is the focus color not part of the colors block?

src/scss/skin-modern/_variables.scss
```diff
@@ -20,3 +22,5 @@ $subtitle-text-border-color: #000 !default;
 
 $animation-duration: .3s !default;
 $animation-duration-short: $animation-duration / 2 !default;
+
+$box-shadow: 0 0 0 2px rgba($color-focus, .8);
```
protyposis: This name is way too generic. Please assign a meaningful name.

src/scss/skin-modern/components/_controlbar.scss
protyposis: `!important` is a no go and brings our customers a lot of trouble. Please find a better way.

src/ts/components/button.ts
```diff
@@ -11,6 +11,11 @@ export interface ButtonConfig extends ComponentConfig {
    * The text as string or localize callback on the button.
    */
   text?: LocalizableText;
+
+  /**
+   * WCAG20 standard for defining info about the component (usually the name)
+   */
+  ariaLabel?: LocalizableText;
```
protyposis: This is already in the `ComponentConfig`

src/ts/components/button.ts
```diff
@@ -27,16 +32,29 @@ export class Button<Config extends ButtonConfig> extends Component<Config> {
...
+      /**
+      * WCAG20 standard to display if a button is pressed or not
+      */
+      'aria-pressed': 'false',
+      'tabindex': this.config.tabindex.toString(),
+    };
+
+    if (this.config.role != null) {
+      buttonElementAttributes['role'] = this.config.role;
+    }
```
protyposis: Is the `if` necessary? Can't you simply assign the `role` whether it is set or not?

src/ts/components/casttogglebutton.ts
```diff
@@ -14,6 +14,7 @@ export class CastToggleButton extends ToggleButton<ToggleButtonConfig> {
     this.config = this.mergeConfig(config, {
       cssClass: 'ui-casttogglebutton',
       text: i18n.getLocalizer('googleCast'),
+      ariaLabel: i18n.getLocalizer('googleCast'),
```
protyposis: Can't you set the `text` as default `ariaLabel`? Seems unnecessary to do this assignment over and over again.

src/ts/components/component.ts
```diff
@@ -46,6 +47,21 @@ export interface ComponentConfig {
...
+
+  /**
+   * Specifies the component role for WCAG20 standard
+   */
+  role?: string;
+
+  /**
+   * WCAG20 requirement for screen reader navigation
+   */
+  tabindex?: number;
```
protyposis: Shouldn't this be called `tabIndex` (camel case)?

src/ts/components/seekbar.ts
```diff
@@ -111,10 +125,17 @@ export class SeekBar extends Component<SeekBarConfig> {
   constructor(config: SeekBarConfig = {}) {
     super(config);
 
+    const keyStepIncrements = this.config.keyStepIncrements != null ? this.config.keyStepIncrements : {
```
protyposis: ` != null` is unnecessary. Actually you can do `this.config.keyStepIncrements || { ...`. best would be `??` but that needs TS 3.7.

src/ts/components/seekbar.ts
```diff
@@ -80,6 +88,12 @@ export class SeekBar extends Component<SeekBarConfig> {
 
   private player: PlayerAPI;
 
+  protected seekBarType: SeekBarType;
+
+  protected getSeekBarType = () => {
```
protyposis: Why an arrow function? And why above the constructor?

src/ts/components/seekbar.ts
```diff
@@ -211,6 +262,14 @@ export class SeekBar extends Component<SeekBarConfig> {
     // update bufferlevel when a segment has been downloaded
     player.on(player.exports.PlayerEvent.SegmentRequestFinished, playbackPositionHandler);
 
+    player.on(player.exports.PlayerEvent.SourceLoaded, () => {
+      if (player.isLive()) {
```
protyposis: This flag is not necessarily known in `SourceLoaded`, and can also change during playback. So you need to additionally use the `LiveStreamDetector`.

src/ts/components/seekbar.ts
```diff
@@ -520,6 +579,8 @@ export class SeekBar extends Component<SeekBarConfig> {
     let seekBarContainer = new DOM('div', {
       'id': this.config.id,
       'class': this.getCssClasses(),
+      'role': 'slider',
+      'tabindex': this.config.tabindex.toString(),
```
protyposis: `toString()` on a `number`??

src/ts/components/seekbar.ts
```diff
@@ -567,6 +628,25 @@ export class SeekBar extends Component<SeekBarConfig> {
...
+      if (this.seekBarType === SeekBarType.Live) {
+        currentSeekValue = -(Math.floor(this.getOffset(e) * this.player.getMaxTimeShift()) - this.player.getMaxTimeShift());
+        currentSeekText = `Timeshift value: ${currentSeekValue}`;
+      } else if (this.seekBarType === SeekBarType.Vod) {
+        currentSeekValue = Math.floor(this.getOffset(e) * this.player.getDuration());
+        currentSeekText = `${StringUtils.secondsToTime(currentSeekValue)} out of ${StringUtils.secondsToTime(this.player.getDuration())}`;
+      } else {
+        currentSeekValue = Math.floor(this.getOffset(e) * 100);
+        currentSeekText = `Value: ${currentSeekValue}`;
+      }
```
protyposis: Isn't this a lot of duplicate code that is already written somewhere else?

src/ts/components/seekbar.ts
```diff
@@ -591,6 +674,8 @@ export class SeekBar extends Component<SeekBarConfig> {
       this.setSeeking(false);
       seeking = false;
 
+      setAriaSliderValue(e);
```
protyposis: Wouldn't this be better placed in `onSeekedEvent`?

src/ts/components/seekbar.ts
```diff
@@ -614,6 +699,8 @@ export class SeekBar extends Component<SeekBarConfig> {
       // Fire seeked event
       this.onSeekEvent();
 
+      setAriaSliderValue(e);
```
protyposis: Wouldn't this be better placed in `onSeekEvent`?

src/ts/components/seekbarcontrols.ts
```diff
@@ -0,0 +1,108 @@
+import { UIUtils } from '../uiutils';
+import { DOM } from '../dom';
+import { PlayerAPI } from 'bitmovin-player';
+import { VolumeController } from '../volumecontroller';
+
```
protyposis: Why isn't this a `class SeekBarController` to fit the remainder of the architecture?

src/ts/components/seekbarcontrols.ts
```diff
@@ -0,0 +1,108 @@
...
+        return arrowKeyControls(player.getTimeShift(), { min: player.getMaxTimeShift(), max: 0 }, keyStepIncrements, player.timeShift);
+    } else if (type === SeekBarType.Vod) {
+        return arrowKeyControls(player.getCurrentTime(), { min: 0, max: player.getDuration() }, keyStepIncrements, player.seek);
+    } else if (type === SeekBarType.Volume && volumeController != null) {
+        const volumeTransition = volumeController.startTransition();
+        return arrowKeyControls(player.getVolume(), { min: 0, max: 100 }, keyStepIncrements, volumeTransition.finish.bind(volumeTransition));
+    }
+};
+
+export const setSeekBarControls = (
```
protyposis: Why is this specifically attached to the seek bar? I think this should be a separate controller, e.g. attached to the `UIContainer`, that issues commands directly on the player. The `SeekBar`, like any other `Component` in this UI framework, will then simply listen to and react upon the player events.

src/ts/components/selectbox.ts
```diff
@@ -13,6 +13,7 @@ import { i18n } from '../localization/i18n';
  *     </select>
  * </code>
  */
+
```
protyposis: No empty line

src/ts/components/settingstogglebutton.ts
```diff
@@ -41,7 +41,20 @@ export class SettingsToggleButton extends ToggleButton<SettingsToggleButtonConfi
...
       settingsPanel: null,
       autoHideWhenNoActiveSettings: true,
+      role: 'pop-up button',
+      ariaLabel: i18n.getLocalizer('settings'),
     }, <SettingsToggleButtonConfig>this.config);
+
+    /**
+     * WCAG20 standard defines which popup menu (element id) is owned by the button
+     */
+    this.getDomElement().attr('aria-owns', config.settingsPanel.getActivePage().getConfig().id);
```
protyposis: Is WCAG20 really supposed to work with the ID of an element? What if we have multiple UI instances?

src/ts/uiutils.ts
```diff
@@ -21,4 +21,16 @@ export namespace UIUtils {
...
     recursiveTreeWalker(component);
   }
+
+  // From: https://github.com/nfriend/ts-keycode-enum/blob/master/Key.enum.ts
+  export enum KeyCode {
+    LeftArrow = 37,
+    UpArrow = 38,
+    RightArrow = 39,
+    DownArrow = 40,
+
```
protyposis: Why this empty line? What's the meaning?

src/ts/components/settingstogglebutton.ts
vzablock: https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships#aria-labelledby

All aria relationships work on element id's.

Aren't the id's always dynamically generated and unique for each instance of the UI?

src/ts/components/seekbarcontrols.ts
vzablock: We discussed this with @dweinber, we decided that the UI wouldn't have controls on the whole player because our customers had different requirements and specifications for it. That is why we agreed on adding controls on on the sliders

src/ts/components/settingstogglebutton.ts
protyposis: You are right, they are (usually).

src/ts/components/seekbarcontrols.ts
protyposis: Then this should be configurable. Needing to focus the seekbar to issue unrelated actions is bad UX.

src/scss/skin-modern/components/_controlbar.scss
vzablock: The `visibility` property removes the current focus location and position of navigation.
As seen in the usage of other players they are just changing the opacity of the elements and never hiding them. In that way the focus is still kept on the current element and you are able to manipulate with that element with the keyboard.
I can't unfortunately think of any other way of maintaining the focus on the current element when the controls are `hidden`.

src/ts/components/casttogglebutton.ts
vzablock: Unfortunately not, because in some cases the aria label is not the same as the text. E.g. `Play/Pause` is confusing for the users when it is shown on the button and the aria label `Play` would make more sense. When the button is active the user automatically implies that `Play` is active

src/ts/components/seekbar.ts
vzablock: `string is not assignable to number`

protyposis: It looks to me that you are calling `toString()` on a `number` (which is a primitive, not an object with functions). Is `this.config.tabindex` not a `number`?

src/ts/components/casttogglebutton.ts
protyposis: That's why I meant as default. Of course it can be overwritten where necessary.

src/ts/components/seekbar.ts
vzablock: It is a number but `DOM` accepts only strings as values

protyposis: 👍

src/ts/components/seekbar.ts
vzablock: The scope of the class gets lost when the method is passed to the controller. I can either do it as arrow function or do `getSeekBarType.bind(this)`. Not sure which one is preferred

src/ts/components/button.ts
vzablock: If the role is not defined (not needed in a specific case) the attribute will be still added to the element with empty value

vzablock:
I tested the performance of the UI when removing the visibility in the controlbar and titlebar. Didn’t see any performance drops testing with 6x CPU slowdown on chrome and as well on chromecast for TV’s.
Also tested web pages of some customers that use a custom version of our UI and with the applied changes nothing was breaking or making issues on the pages.

src/ts/components/seekbar.ts
```diff
@@ -138,18 +166,39 @@ export class SeekBar extends Component<SeekBarConfig> {
     // (the call must be up here to be executed for the volume slider as well)
     this.setPosition(this.seekBarBackdrop, 100);
 
+    // Add seekbar controls to the seekbar
+    const seekBarController = new SeekBarController(this.config.keyStepIncrements, player, uimanager.getConfig().volumeController);
+
+    seekBarController.setSeekBarControls(this.getDomElement(), this.getSeekBarType);
+
     if (!configureSeek) {
+      this.seekBarType = SeekBarType.Volume;
```
stonko1994: Wouldn't that fit better in the subclass (Volume Slider)?

src/ts/components/seekbarcontroller.ts
```diff
@@ -0,0 +1,111 @@
...
+
+export enum SeekBarType {
+    Vod,
+    Live,
+    Volume,
+}
+
+const changeRangeValue = (
+    value: number,
+    range: { min: number, max: number },
```
stonko1994: It maybe would make sense to extract `{ min: number, max: number }` into a `Range` interface.

src/ts/components/seekbarcontroller.ts
```diff
@@ -0,0 +1,111 @@
...
+        cb(range.min);
+    } else if (value > range.max) {
+        cb(range.max);
+    } else {
+        cb(value);
+    }
+};
+
+export class SeekBarController {
+    protected keyStepIncrements: { leftRight: number, upDown: number };
```
stonko1994: It maybe would make sense to extract `{ leftRight: number, upDown: number }` into a `KeyStepIncrementsConfig` interface (or with a similar name).

src/ts/components/selectbox.ts
```diff
@@ -76,4 +78,4 @@ export class SelectBox extends ListSelector<ListSelectorConfig> {
       this.updateDomItems(value);
     }
   }
-}
+}
```
stonko1994: Keep the newline at EOF

src/ts/components/settingspanel.ts
```diff
@@ -92,16 +92,6 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
         this.hide();
         this.hideHoveredSelectBoxes();
       });
-
-      this.onShow.subscribe(() => {
```
stonko1994: This could be added to the changelog as bugfix 🤔

src/ts/components/uicontainer.ts
```diff
@@ -153,6 +156,17 @@ export class UIContainer extends Container<UIContainerConfig> {
       handler: () => {
         showUi();
       },
+    }, {
+      name: 'focusin',
+      handler: (e) => {
+        console.warn(e);
```
stonko1994: Remove warning

src/ts/components/settingspanelpage.ts
```diff
@@ -73,6 +74,11 @@ export class SettingsPanelPage extends Container<ContainerConfig> {
 
   onActiveEvent() {
     this.settingsPanelPageEvents.onActive.dispatch(this);
+    const activeItems = this.getItems().filter((item) => item.isActive() === true);
```
stonko1994:
```suggestion
    const activeItems = this.getItems().filter((item) => item.isActive());
```

src/ts/stringutils.ts
```diff
@@ -32,6 +33,26 @@ export namespace StringUtils {
         .replace('ss', leftPadWithZeros(seconds, 2));
   }
 
+  export function secondsToText(totalSeconds: number): string {
```
stonko1994: why isn't `secondsToTime` sufficient?

vzablock: That only displays the time in format `00:00:00`. For voice over that doesn't actually help users understand what those numbers mean so `secondsToText` displays the format in `00 hours 00 minutes 00 seconds`

src/ts/components/seekbar.ts
vzablock: Yes but I wanted to keep everything in one place regarding Web Accessibility since the `aria` attributes for the volume are defined here. In the `Volumeslider` it is not that easy to add those labels

src/ts/stringutils.ts
stonko1994: 👍

src/ts/components/seekbar.ts
stonko1994: 👍

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Web Content Accessibility (WCAG) 2.0
 - Slider controls for Web Content Accessibility
+- Fixed issue where settings panel show subscription was ignored on small UI screens
```
stonko1994: Please move under `Fixed` section without `fixed` prefix

CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Added
+
```
protyposis: no blank line

CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
...
+- Web Content Accessibility (WCAG) 2.0
+- Slider controls for Web Content Accessibility
+
+### Fixed
+
```
protyposis: no blank line

CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Added
+
+- Web Content Accessibility (WCAG) 2.0
+- Slider controls for Web Content Accessibility
```
protyposis: Isn't the latter included in the first one?

CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
...
+
+### Fixed
+
+- Settings panel show subscription ignored on small UI screens
```
protyposis: I don't understand what that means, please rephrase

src/scss/skin-modern/_mixins.scss
```diff
@@ -22,6 +22,16 @@
   }
 }
 
+@mixin hidden-animated-focusable($duration: $animation-duration) {
+  opacity: 1;
+  transition: hidden-animated-transition($duration);
+
+  &.#{$prefix}-hidden {
+    opacity: 0;
+    transition: opacity $duration, visibility $duration;
```
protyposis: `visibility`?

src/ts/components/seekbar.ts
protyposis: Please convert to a conventional class function and pass as `() => this.getSeekBarType()` instead of `this.getSeekBarType`

src/ts/components/seekbar.ts
```diff
@@ -138,18 +166,39 @@ export class SeekBar extends Component<SeekBarConfig> {
...
+      this.setAriaSliderMinMax('0', '100');
+      this.getDomElement().attr('aria-label', i18n.performLocalization(i18n.getLocalizer('settings.ariaLabel.volume')));
+      this.getDomElement().attr('aria-valuenow', `${player.getVolume()}`);
+      this.getDomElement().attr('aria-valuetext', `${i18n.performLocalization(i18n.getLocalizer('seekBar.value'))}: ${player.getVolume()}`);
       // The configureSeek flag can be used by subclasses to disable configuration as seek bar. E.g. the volume
       // slider is reusing this component but adds its own functionality, and does not need the seek functionality.
       // This is actually a hack, the proper solution would be for both seek bar and volume sliders to extend
       // a common base slider component and implement their functionality there.
+
       return;
```
protyposis: * A `SeekBar` should not configure a `VolumeSlider` - they are totally unrelated
 * Why is the added code above the describing comment of the conditional flag `configureSeek`?
 * Why is there a blank line added?

src/ts/components/seekbar.ts
```diff
@@ -164,12 +213,20 @@ export class SeekBar extends Component<SeekBarConfig> {
...
           let playbackPositionPercentage = 100 - (100 / player.getMaxTimeShift() * player.getTimeShift());
           this.setPlaybackPosition(playbackPositionPercentage);
+          this.setAriaSliderMinMax(player.getMaxTimeShift().toString(), '0');
+        }
+
+        if (this.isUiShown) {
+          const timeshiftValue = Math.ceil(this.player.getTimeShift()).toString();
+          this.getDomElement().attr('aria-valuenow', timeshiftValue);
+          this.getDomElement().attr('aria-valuetext', `${i18n.performLocalization(i18n.getLocalizer('seekBar.timeshift'))} ${i18n.performLocalization(i18n.getLocalizer('seekBar.value'))}: ${timeshiftValue}`);
         }
```
protyposis: Why are these values only set then the UI is shown? Are they ever reset when the UI is hidden? This should be implemented in a reactive way, directly depending on the events. Setting a flag in the events and then reacting at some undefined later point in time leads to ugly edge cases.

src/ts/components/seekbar.ts
```diff
@@ -194,6 +251,14 @@ export class SeekBar extends Component<SeekBarConfig> {
...
 
         this.setBufferPosition(playbackPositionPercentage + bufferPercentage);
+
+        this.setAriaSliderMinMax('0', playerDuration.toString());
+
+        if (this.isUiShown) {
+          const ariaValueText = `${StringUtils.secondsToText(this.player.getCurrentTime())} ${i18n.performLocalization(i18n.getLocalizer('seekBar.durationText'))} ${StringUtils.secondsToText(playerDuration)}`;
+          this.getDomElement().attr('aria-valuenow', Math.floor(this.player.getCurrentTime()).toString());
+          this.getDomElement().attr('aria-valuetext', ariaValueText);
+        }
```
protyposis: As above

src/ts/components/seekbarcontroller.ts
```diff
@@ -0,0 +1,121 @@
...
+    min: number;
+    max: number;
+}
+
+interface KeyStepIncrementsConfig {
+    leftRight: number;
+    upDown: number;
+}
+
+const changeRangeValue = (
```
protyposis: Methods like this one are usually called something with "coerce", e.g. `coerceValueIntoRange`

src/ts/components/seekbarcontroller.ts
```diff
@@ -0,0 +1,121 @@
...
+            return this.arrowKeyControls(this.player.getTimeShift(), { min: this.player.getMaxTimeShift(), max: 0 }, this.player.timeShift);
+        } else if (type === SeekBarType.Vod) {
+            return this.arrowKeyControls(this.player.getCurrentTime(), { min: 0, max: this.player.getDuration() }, this.player.seek);
+        } else if (type === SeekBarType.Volume && this.volumeController != null) {
+            const volumeTransition = this.volumeController.startTransition();
+            return this.arrowKeyControls(this.player.getVolume(), { min: 0, max: 100 }, volumeTransition.finish.bind(volumeTransition));
+        }
+    }
+
+    public setSeekBarControls (domElement: DOM, type: () => SeekBarType) {
```
protyposis: Whitespacing

src/ts/localization/i18n.ts
```diff
@@ -41,6 +41,12 @@ interface Vocabulary {
   'settings.subtitles.font.family.smallcapital': string;
   'settings.subtitles.window.color': string;
   'settings.subtitles.window.opacity': string;
+  'settings.ariaLabel.volume': string;
+  'settings.ariaLabel.mute': string;
+  'settings.ariaLabel.play': string;
```
protyposis: Please make them generic labels. Constraining them to "aria" labels unnecessarily reduces their versatility.

src/ts/components/seekbar.ts
vzablock: They stay as they are when the UI is hidden. We don't want to update the attributes because the user would never hear the video if it keeps updating that is why we stop updating until the UI is shown . We are listening to the `uimanager.onControlsShow`

src/ts/components/seekbarcontroller.ts
vzablock: Not sure why if you can explain? It just changes the value of the slider

src/ts/localization/i18n.ts
vzablock: For play I can do it but for volume not that easy since we have defined `"settings.audio.mute" : "Volume/Mute"` Which should actually be `settings.audio.volumeMute` and I can't define both of them separately

CHANGELOG.md
vzablock: Yes but it was added as a separate issue. And would benefit users to know that there are slider controls now implemented (people who don't need WCAG

protyposis: Issues are internal and unrelated to the public changelog. "Slider controls" is a super generic term and could mean anything - this doesn't help anybody in understanding what exactly has been added. E.g. "keyboard controls" would be much more helpful.

src/ts/components/seekbar.ts
vzablock: https://github.com/bitmovin/bitmovin-player-ui/pull/282#discussion_r354857696

protyposis: I have seen this comment - but that doesn't make the issue go away ;) I'm sure this can be written in a better way.

src/ts/components/seekbar.ts
protyposis: > They stay as they are when the UI is hidden.

So you deliberately want to keep values which are then wrong and misrepresent the player state?

src/ts/components/seekbarcontroller.ts
protyposis: Because a coercion is what you do here. You force a value into a specific range.

src/ts/components/seekbar.ts
vzablock: Correct because otherwise it would be unintuitive for the user

src/ts/localization/i18n.ts
protyposis: Then make the text labels on those buttons stateful as well?

src/ts/components/seekbar.ts
protyposis: Keeping outdated values is more intuitive than resetting an outdated state?

vzablock: But to what would the values reset to? They get updated when the UI is shown again. If we change the values in any way when the UI is hidden the user will hear that. The user doesn't need that info as it will make noise for hime while he is listening to the video

src/ts/localization/i18n.ts
vzablock: We cannot make it stateful as it doesn't make sense for the client:

https://github.com/bitmovin/player-issues-web/issues/2586#issuecomment-560999083

Feedback number 4.

src/ts/components/seekbar.ts
vzablock: Only way I see is separating the seek slider from the volume slider, not extending one from another like we do now in this case

src/ts/components/seekbar.ts
protyposis: > But to what would the values reset to?

The state they were is before being populated first. When I look into the DOM I don't want to see irrelevant outdated values.

Will the screenreader talk about values being removed?

src/ts/components/seekbar.ts
protyposis: What about a `protected configureAria(...) {. .. }` method that can be overwritten in the volume slider?

src/ts/localization/i18n.ts
protyposis: The answer you lined is exactly demanding an aria "state", which is what I meant with "stateful". if that is implemented for accessibility, why not do it for the "normal" label as well?

src/ts/components/seekbar.ts
vzablock: Yes it reads content that is not contained inside the element. I checked out other players for various sources and all of them use the same or similar system as I implemented here

src/ts/localization/i18n.ts
vzablock: Because we already have it open for other clients and changing the "normal" labels can be api breaking 😞

src/scss/skin-modern/_mixins.scss
```diff
@@ -28,7 +28,7 @@
 
   &.#{$prefix}-hidden {
     opacity: 0;
-    transition: opacity $duration, visibility $duration;
+    transition: opacity $duration $duration;
```
stonko1994: Is this valid CSS? I don't think this are the correct properties for the `transition`

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Added
+- Web Content Accessibility (WCAG) 2.0
+- Keyboard controls for sliders when focused
+
+### Fixed
+- Settings panel show and hide listener ignored on small UI screens
```
```suggestion
- `onActive` and `onInactive` not fired for root `SettingsPanelPage` when `SettingsPanel` was configured with `hideDelay` of `-1`
```

src/ts/localization/i18n.ts
protyposis: The ID is called `settings.audio.mute` so please change the string from `Volume/Mute` to `Mute`, and then add a `settings.audio.volume` with string `Volume`. Also add a `settings.audio.unmute` => `Unmute` label which is the label that should be set while the player is muted.

PR #284, opened 2019-11-25
Add web acessibility menu controls and labels

vzablock:
Issue: Support web accessibility labels and controls

Task: Added web accessibility labels and controls for navigating through the player with keyboard only controls


src/scss/skin-modern/components/_button.scss
```diff
@@ -37,6 +37,11 @@
   }
 
   @include hidden;
+
+  &:focus {
+    outline: none;
+    box-shadow: 0 0 0 2px rgba(27,127,204,.8);
```
stonko1994: Please extract to `_variables.scss`

stonko1994:
```suggestion
    box-shadow: 0 0 0 2px rgba(27, 127, 204, .8);
```

src/scss/skin-modern/components/_hugeplaybacktogglebutton.scss
```diff
@@ -81,4 +81,8 @@
       }
     }
   }
+
+  &:focus {
+    box-shadow: 0 0 0 2px rgba(27,127,204,.8);
```
stonko1994: Please extract to `_variables.scss`

stonko1994:
```suggestion
    box-shadow: 0 0 0 2px rgba(27, 127, 204, .8);
```

src/scss/skin-modern/components/_selectbox.scss
```diff
@@ -17,4 +17,8 @@
       color: $color-highlight;
     }
   }
+
+  &:focus {
+    box-shadow: 0 0 0 2px rgba(27,127,204,.8);
```
stonko1994: Please extract to `_variables.scss`

stonko1994:
```suggestion
    box-shadow: 0 0 0 2px rgba(27, 127, 204, .8);
```

src/ts/components/button.ts
```diff
@@ -16,6 +16,11 @@ export interface ButtonConfig extends ComponentConfig {
    * WCAG20 standard for defining info about the component (usually the name)
    */
   ariaLabel?: string;
+
+  /**
+   * WCAG20 standard to display if a button is pressed or not
+   */
+  ariaPressed?: string;
```
stonko1994: Can we change this to `boolean` and make the `toString` internally? Or why is this a `string` if it's suffixed with `pressed`

src/ts/components/button.ts
```diff
@@ -40,6 +46,7 @@ export class Button<Config extends ButtonConfig> extends Component<Config> {
       'id': this.config.id,
       'aria-label': this.config.ariaLabel,
       'class': this.getCssClasses(),
+      'aria-pressed': 'false'
```
stonko1994: Why not respecting the config?

src/ts/components/container.ts
```diff
@@ -10,6 +10,8 @@ export interface ContainerConfig extends ComponentConfig {
    * Child components of the container.
    */
   components?: Component<ComponentConfig>[];
+
+  ariaLabel?: string;
```
stonko1994: Please add a little documentation

src/ts/components/hugeplaybacktogglebutton.ts
```diff
@@ -17,7 +17,7 @@ export class HugePlaybackToggleButton extends PlaybackToggleButton {
       cssClass: 'ui-hugeplaybacktogglebutton',
       text: i18n.getLocalizer('playPause'),
       role: 'button',
-      ariaLabel: 'playback',
+      ariaLabel: 'Playback',
```
stonko1994: Can you rename those on the base PR? (Applies for all)

src/ts/components/settingspanelitem.ts
```diff
@@ -32,7 +32,7 @@ export class SettingsPanelItem extends Container<ContainerConfig> {
 
     this.config = this.mergeConfig(config, {
       cssClass: 'ui-settings-panel-item',
-      role: 'menuitem',
+      role: this.config.role != null ? this.config.role : 'menuitem',
```
stonko1994: The following should be sufficient? 🤔 This is just the default config and does not override what is in the `config`
```suggestion
      role: 'menuitem',
```

src/ts/dom.ts
```diff
@@ -169,6 +169,15 @@ export class DOM {
     return new DOM(allChildElements);
   }
 
+  /**
+   * Focuses to the first input element
+   */
+  focusToFirstInput(): DOM {
```
stonko1994: Does this need a return value?

src/ts/uiutils.ts
```diff
@@ -28,6 +28,9 @@ export namespace UIUtils {
     UpArrow = 38,
     RightArrow = 39,
     DownArrow = 40,
+
     Space = 32,
+    End = 35,
```
stonko1994: What are the `End` and the `Home` key?

src/ts/components/settingspanelitem.ts
vzablock: The problem is that I need to override with config if there is a `menubar` present (bar with inputs inside of it). Since this component is used as a wrapper for input fields and as well for menu bars

vzablock:
@stonko1994 
Should I show the UI when we are seeking with keys?

stonko1994:
> Should I show the UI when we are seeking with keys?

I would say yes. I would treat this like a normal UI interaction which interrupts the `clearTimeout`

vzablock:
@stonko1994 https://stackoverflow.com/questions/1848390/safari-ignoring-tabindex

Safari ignores tab navigation by default (not sure why only the sliders are working) 😄

stonko1994:
@vzablock 🤦‍♂ wasn't aware of this ... sry 😄

src/ts/components/seekbar.ts
```diff
@@ -169,7 +169,7 @@ export class SeekBar extends Component<SeekBarConfig> {
     if (!configureSeek) {
       this.seekBarType = SeekBarType.Volume;
       this.setAriaSliderMinMax('0', '100');
-      this.getDomElement().attr('aria-label', 'volume');
+      this.getDomElement().attr('aria-label', 'Volume');
```
stonko1994: Why not using i18n here too?

src/ts/components/seekbar.ts
```diff
@@ -184,7 +184,7 @@ export class SeekBar extends Component<SeekBarConfig> {
     let isUserSeeking = false;
     let isPlayerSeeking = false;
 
-    this.getDomElement().attr('aria-label', 'seek');
+    this.getDomElement().attr('aria-label', 'Video timeline');
```
stonko1994: i18n?

src/ts/components/settingspanelitem.ts
stonko1994: I don't get it. What is a `menubar` in this context?

src/ts/components/button.ts
```diff
@@ -32,14 +32,21 @@ export class Button<Config extends ButtonConfig> extends Component<Config> {
 
     this.config = this.mergeConfig(config, {
       cssClass: 'ui-button',
+      role: 'button',
+      tabindex: '0',
```
stonko1994: you changed the type to number right?

src/ts/components/settingspanelitem.ts
vzablock: Screenshot 2019-11-27 at 15 08 08
Screenshot 2019-11-27 at 15 08 16


These two are actually menu bars, they contain more than one control element

stonko1994: Ah ok makes sense. But anyway once you set a config when initializing (e.g. in the `SubtitleSettingsPanelPage`) the role from this config will be applied. So no need to change the default value.

spec/components/seekbar.spec.ts
```diff
@@ -20,6 +20,7 @@ describe('SeekBar', () => {
     it('when live had no time shift', () => {
       jest.spyOn(playerMock, 'getMaxTimeShift').mockReturnValue(0);
       jest.spyOn(playerMock, 'isLive').mockReturnValue(true);
+      jest.spyOn(playerMock, 'getDuration').mockReturnValue(888);
```
stonko1994: Please return `Infinite` for the duration in case of a live.

spec/components/seekbar.spec.ts
```diff
@@ -35,6 +36,8 @@ describe('SeekBar', () => {
   describe('disables smoothPlaybackPositionUpdater on Live Streams', () => {
     it('clears smoothPlaybackPositionUpdater when AD is finished', () => {
       jest.spyOn(playerMock, 'isLive').mockReturnValue(true);
+      jest.spyOn(playerMock, 'getMaxTimeShift').mockReturnValue(-60);
+      jest.spyOn(playerMock, 'getDuration').mockReturnValue(60);
```
stonko1994: Please return `Infinite` for the duration in case of a live.

src/scss/skin-modern/components/_hugeplaybacktogglebutton.scss
```diff
@@ -81,4 +81,8 @@
       }
     }
   }
+
+  &:focus {
```
stonko1994: Would it make sense to extract these to some mixins? `focusable` maybe

vzablock:
@stonko1994 Shouldn't a players default behaviour be that if audio changes the player should be automatically unmuted? If that is the case that should be handled in the player itself not in the UI

stonko1994:
@vzablock In the end it should not matter. It should just be the same behaviour then clicking on the volume slider. You would need to use the `volumeTransition` like in the `VolumeSlider.ts`

src/ts/components/seekbarcontrols.ts
```diff
@@ -50,7 +50,15 @@ const seekBarControls = (
     } else if (type === SeekBarType.Vod) {
         return arrowKeyControls(player.getCurrentTime(), { min: 0, max: player.getDuration() }, keyStepIncrements, player.seek);
     } else {
-        return arrowKeyControls(player.getVolume(), { min: 0, max: 100 }, keyStepIncrements, player.setVolume);
+        const setVolume = (volume: number) => {
```
stonko1994: Why are you not using the already existing `volumeController` which is handling this?

vzablock: I need to pass the `uimanager` to the controls as well then just to use the `volumeController`. Not sure if that is really needed since the slider should only unmute the player if it is being used

stonko1994: sure but why having the logic of unmuting in two places? And what I can remember there are some edge cases which are handled in the `VolumeController`. That's why we have it.

vzablock:
@stonko1994 Safari now properly focuses the element but there is another issue regarding the default `select` box skin for safari. It is not possible to make a outline or focus effect on it

src/scss/skin-modern/components/_uicontainer.scss
```diff
@@ -29,6 +29,12 @@
     }
   }
 
+  &.#{$prefix}-controls-shown {
+    .#{$prefix}-ui-hugeplaybacktogglebutton:focus {
+      box-shadow: inset -4px -3px 2px 9px rgba(27, 127, 204, 1) !important;
```
stonko1994: - The `!important` here should not be necessary or? 🤔 
- If this is the bitmovin color you could use `rgba($color-highlight, .8);` If not, extract the color and use `rgba(color, alphy)`

src/ts/components/closebutton.ts
```diff
@@ -24,6 +25,7 @@ export class CloseButton extends Button<CloseButtonConfig> {
     this.config = this.mergeConfig(config, {
       cssClass: 'ui-closebutton',
       text: 'Close',
```
stonko1994: use the new translation for the `text` as well.

PR #285, opened 2019-12-17
Feature/disable focus for ios

vzablock:
Disable focus for iOS devices because it automatically opens select inputs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - `onActive` and `onInactive` not fired for root `SettingsPanelPage` when `SettingsPanel` was configured with `hideDelay` of `-1`
+- Disabled focus on settings panel when clicked for iOS, behavior opens keyboard controls when focused 
 
 ## [3.9.1]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - `onActive` and `onInactive` not fired for root `SettingsPanelPage` when `SettingsPanel` was configured with `hideDelay` of `-1`
-- Disabled focus on settings panel when clicked for iOS, behavior opens keyboard controls when focused 
+- Automatic opening of the first select box when the `SettingsPanel` becomes visible on iOS devices
 
 ## [3.9.1]
 
--- a/src/ts/browserutils.ts
+++ b/src/ts/browserutils.ts
@@ -29,7 +29,7 @@ export class BrowserUtils {
     if (!this.windowExists()) {
       return false;
     }
-    return navigator && navigator.userAgent && /iPad|iPhone|iPod/.test(navigator.userAgent);
+    return navigator && navigator.userAgent && (/iPad|iPhone|iPod/.test(navigator.userAgent) || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1));
   }
 
   private static windowExists(): boolean {
--- a/src/ts/browserutils.ts
+++ b/src/ts/browserutils.ts
@@ -25,6 +25,13 @@ export class BrowserUtils {
     return navigator && navigator.userAgent && /Android/.test(navigator.userAgent);
   }
 
+  static get isIOS(): boolean {
+    if (!this.windowExists()) {
+      return false;
+    }
+    return navigator && navigator.userAgent && /iPad|iPhone|iPod/.test(navigator.userAgent);
+  }
+
   private static windowExists(): boolean {
     return typeof window !== 'undefined';
   }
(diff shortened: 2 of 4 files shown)


src/ts/components/settingspanelpage.ts
```diff
@@ -73,10 +73,12 @@ export class SettingsPanelPage extends Container<ContainerConfig> {
   }
 
   onActiveEvent() {
-    this.settingsPanelPageEvents.onActive.dispatch(this);
+    const isIOS = /iPad|iPhone|iPod/.test(navigator.userAgent);
```
protyposis: Please integrate this into https://github.com/bitmovin/bitmovin-player-ui/blob/master/src/ts/browserutils.ts

src/ts/browserutils.ts
```diff
@@ -25,6 +25,13 @@ export class BrowserUtils {
     return navigator && navigator.userAgent && /Android/.test(navigator.userAgent);
   }
 
+  static get isIOS(): boolean {
+    if (!this.windowExists()) {
+      return false;
+    }
+    return navigator && navigator.userAgent && /iPad|iPhone|iPod/.test(navigator.userAgent);
```
stonko1994: We have one more problem:
- iPadOS (iOS 13 on an iPad) will be recognized as `macOS` as they don't have `iPad` in the `userAgent`. We should also check for those devices.

I think we will need to make the check like: 
`isIos` || (`isMacOs` && `isMobile`) you could check for the `touch` events. Or research quickly if there is already a way to detect this.

src/ts/browserutils.ts
```diff
@@ -25,6 +25,13 @@ export class BrowserUtils {
     return navigator && navigator.userAgent && /Android/.test(navigator.userAgent);
   }
 
+  static get isIOS(): boolean {
+    if (!this.windowExists()) {
+      return false;
+    }
+    return navigator && navigator.userAgent && (/iPad|iPhone|iPod/.test(navigator.userAgent) || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1));
```
protyposis: Please move that `navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1` to a separate named method. Now looking at this, nobody knows what the actual purpose of this is.

stonko1994: I would prefer to split this up to prevent confusions in the future:
- `isIOS`
- `isMacOS` / `isMacIntel`
- `isTouchSupported` (extract from `seekbar.ts`)

and then use it like `isIOS || (isMacOS && isTouchSupported)`

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - `onActive` and `onInactive` not fired for root `SettingsPanelPage` when `SettingsPanel` was configured with `hideDelay` of `-1`
+- Disabled focus on settings panel when clicked for iOS, behavior opens keyboard controls when focused
```
```suggestion
- Automatic opening of the first select box when the `SettingsPanel` becomes visible on iOS devices
```

src/ts/browserutils.ts
```diff
@@ -25,6 +25,27 @@ export class BrowserUtils {
...
+    return navigator && navigator.userAgent && /iPad|iPhone|iPod/.test(navigator.userAgent);
+  }
+
+  static get isMacIntel(): boolean {
+    if (!this.windowExists()) {
+      return false;
+    }
+    return navigator && navigator.userAgent && navigator.platform === 'MacIntel';
+  }
+  // https://hacks.mozilla.org/2013/04/detecting-touch-its-the-why-not-the-how/
```
stonko1994: Please add an empty line between method end and comment

src/ts/components/settingspanelpage.ts
```diff
@@ -73,10 +74,11 @@ export class SettingsPanelPage extends Container<ContainerConfig> {
   }
 
   onActiveEvent() {
-    this.settingsPanelPageEvents.onActive.dispatch(this);
     const activeItems = this.getItems().filter((item) => item.isActive());
 
-    if (activeItems.length > 0) {
+    this.settingsPanelPageEvents.onActive.dispatch(this);
+    // Disable focus for iOS and iPad OS 13 because it automatically opens select inputs
```
stonko1994:
```suggestion
    // Disable focus for iOS and iPadOS 13 because it opens a select inputs
```

src/ts/components/settingspanelpage.ts
```diff
@@ -73,10 +74,11 @@ export class SettingsPanelPage extends Container<ContainerConfig> {
   }
 
   onActiveEvent() {
-    this.settingsPanelPageEvents.onActive.dispatch(this);
     const activeItems = this.getItems().filter((item) => item.isActive());
 
-    if (activeItems.length > 0) {
+    this.settingsPanelPageEvents.onActive.dispatch(this);
+    // Disable focus for iOS and iPadOS 13 because it opens a select input
```
protyposis:
```suggestion
    // Disable focus for iOS and iPadOS 13. They open select boxes automatically on focus and we want to avoid that.
```

protyposis:
@vzablock why do you merge with a rejected review?

vzablock:
Sorry only got the notification that it was approved 😞

vzablock:
@protyposis closing PR change was made in develop

PR #287, opened 2020-01-22
Fix TTML subtitle background on Chromecast

vzablock:
Added a fix for empty transparent boxes when TTML subtitles are used with Chromecast devices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-### Development
+## [develop]
 
 ### Fixed
 - Subtitle background when TTML subtitles are used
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+### Development
```
stonko1994: Please us heading lvl 2 and write it as below:

```suggestion
## [develop]
```

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- Subtitle background when TTML subtitles are used
```
protyposis: The fix is only for the Chromecast UI so please mention that here as well
```suggestion
- Empty background boxes with TTML subtitles on Chromecast
```

PR #288, opened 2020-01-23
Fix UI hiding while seeking or changing volume

vzablock:
Fixed UI hiding while seeking or changing volume.
Holding and dragging the sliders to change values won't make the UI hide anymore.

### DIFF
--- a/src/ts/components/seekbar.ts
+++ b/src/ts/components/seekbar.ts
@@ -92,7 +92,6 @@ export class SeekBar extends Component<SeekBarConfig> {
   protected seekBarType: SeekBarType;
 
   protected isUiShown: boolean;
-  protected uiManager: UIInstanceManager;
 
   /**
    * Buffer of the the current playback position. The position must be buffered in case the element
@@ -169,7 +168,6 @@ export class SeekBar extends Component<SeekBarConfig> {
     super.configure(player, uimanager);
 
     this.player = player;
-    this.uiManager = uimanager;
 
     // Apply scaling transform to the backdrop bar to have all bars rendered similarly
     // (the call must be up here to be executed for the volume slider as well)
--- a/src/ts/components/seekbar.ts
+++ b/src/ts/components/seekbar.ts
@@ -92,6 +92,7 @@ export class SeekBar extends Component<SeekBarConfig> {
   protected seekBarType: SeekBarType;
 
   protected isUiShown: boolean;
+  protected uiManager: UIInstanceManager;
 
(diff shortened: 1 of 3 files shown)


vzablock:
@stonko1994 the `stopPropagation` would disable the `mousemove` event for showing and hiding UI that was bound in `uicontainer`.

src/ts/components/seekbar.ts
```diff
@@ -689,10 +693,6 @@ export class SeekBar extends Component<SeekBarConfig> {
       e.preventDefault();
 
       if (seeking) {
-        // During a seek (when mouse is down or touch move active), we need to stop propagation to avoid
-        // the VR viewport reacting to the moves.
-        e.stopPropagation();
```
protyposis: Why don't you make it conditional here, like above?

src/ts/components/seekbar.ts
```diff
@@ -689,10 +693,6 @@ export class SeekBar extends Component<SeekBarConfig> {
...
-        // Because the stopped propagation inhibits the event on the document, we need to call it from here
```
protyposis: This comment indicates that the next call was only present because of the propagation stop. Are you sure it makes sense to keep it?

src/ts/components/uicontainer.ts
```diff
@@ -158,7 +158,7 @@ export class UIContainer extends Container<UIContainerConfig> {
       },
     }, {
       name: 'focusin',
-      handler: (e) => {
+      handler: () => {
```
protyposis: This change is completely unrelated to this PR.

vzablock: It is I agree but the argument is unused in the method (a bit of code cleaning doesn't hurt in the long run)

src/ts/components/seekbar.ts
vzablock: The method `mouseTouchMoveHandler` has the `e.stopPropagation();` call. Doesn't make sense to have the method being called 2 times. Same can be discussed regarding `e.preventDefault()`

src/ts/components/seekbar.ts
vzablock: We have the `mouseTouchHandler` being called which has the conditional in it

src/ts/components/uicontainer.ts
protyposis: Code cleaning, when if unrelated to the PR, should be done in a separate PR.

PR #289, opened 2020-02-03
Add VTT style properties and regions

vzablock:
Added styling of new VTT properties and regions.

VTT subtitles now have supported styling that comes from the `.vtt` file.
Reference:
 - https://www.w3.org/TR/webvtt1/
 - https://www.speechpad.com/captions/webvtt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [develop]
 
 ### Added
-- Subtitle stlying with VTT properties and regions
+- Support for regions in VTT subtitles
 
 ### Fixed
 - UI hiding when actively using seek or volume slider
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
+### Added
+- Subtitle stlying with VTT properties and regions
+
 ### Fixed
 - UI hiding when actively using seek or volume slider
 - Empty background boxes with TTML subtitles on Chromecast
--- a/spec/vttutils.spec.ts
+++ b/spec/vttutils.spec.ts
@@ -4,10 +4,6 @@ import { VTTRegionProperties, VTTProperties } from 'bitmovin-player';
 import { MockHelper } from './helper/MockHelper';
 
 describe('Vtt Utils', () => {
-  beforeEach(() => {
-    // Setup DOM Mock
-
-  });
   describe('Vtt Region', () => {
     it('should set region css properties', () => {
         const mockRegionContainer = generateSubtitleRegionContainerMock();
--- a/spec/vttutils.spec.ts
+++ b/spec/vttutils.spec.ts
@@ -11,7 +11,7 @@ describe('Vtt Utils', () => {
         mockRegionContainer.getDomElement = () => mockDomElement;
         const spyCss = jest.spyOn(mockRegionContainer.getDomElement(), 'css');
     
-        VttUtils.setVttRegionStyles(mockRegionContainer, vttRegionProps, {width: 1000, height: 800});
+        VttUtils.setVttRegionStyles(mockRegionContainer, vttRegionProps, { width: 1000, height: 800 });
 
         expect(spyCss).toHaveBeenCalledTimes(8);
         expect(spyCss).toHaveBeenNthCalledWith(1, 'position', 'absolute');
@@ -35,18 +35,7 @@ describe('Vtt Utils', () => {
       VttUtils.setVttCueBoxStyles(mockRegionContainer);
       
       expect(spyCss).toHaveBeenCalledTimes(11);
-      expect(spyCss).toHaveBeenNthCalledWith(8, 'text-align', 'left');
-    });
-
-    it('should not set default position', () => {
-      const mockRegionContainer = generateSubtitleCueBoxMock(false, { position: 50 } as VTTProperties);
-      const mockDomElement = MockHelper.generateDOMMock();
-      mockRegionContainer.getDomElement = () => mockDomElement;
-      const spyCss = jest.spyOn(mockRegionContainer.getDomElement(), 'css');
-
-      VttUtils.setVttCueBoxStyles(mockRegionContainer);
-      
-      expect(spyCss).toHaveBeenCalledTimes(9);
+      expect(spyCss).toHaveBeenNthCalledWith(9, 'text-align', 'left');
     });
 
     describe('Default Cue Box Styles', () => {
@@ -116,7 +105,7 @@ describe('Vtt Utils', () => {
             VttUtils.setVttCueBoxStyles(mockRegionContainer);
             
             expect(spyCss).toHaveBeenCalledTimes(12);
-            expect(spyCss).toHaveBeenNthCalledWith(8, 'top', '50%');
+            expect(spyCss).toHaveBeenNthCalledWith(9, 'top', '50%');
           });
 
           it('should set positive line positioning', () => {
@@ -128,7 +117,7 @@ describe('Vtt Utils', () => {
             VttUtils.setVttCueBoxStyles(mockRegionContainer);
             
             expect(spyCss).toHaveBeenCalledTimes(12);
-            expect(spyCss).toHaveBeenNthCalledWith(8, 'top', '112px');
+            expect(spyCss).toHaveBeenNthCalledWith(9, 'top', '112px');
           });
 
           it('should set negative line positioning', () => {
@@ -140,7 +129,7 @@ describe('Vtt Utils', () => {
             VttUtils.setVttCueBoxStyles(mockRegionContainer);
             
             expect(spyCss).toHaveBeenCalledTimes(12);
-            expect(spyCss).toHaveBeenNthCalledWith(8, 'bottom', '112px');
+            expect(spyCss).toHaveBeenNthCalledWith(9, 'bottom', '112px');
           });
 
           describe('Line Alignment', () => {
@@ -164,7 +153,7 @@ describe('Vtt Utils', () => {
               VttUtils.setVttCueBoxStyles(mockRegionContainer);
               
               expect(spyCss).toHaveBeenCalledTimes(13);
-              expect(spyCss).toHaveBeenNthCalledWith(9, 'margin-top', '-14px');
+              expect(spyCss).toHaveBeenNthCalledWith(10, 'margin-top', '-14px');
             });
 
             it('should do end line alignment', () => {
@@ -176,7 +165,7 @@ describe('Vtt Utils', () => {
               VttUtils.setVttCueBoxStyles(mockRegionContainer);
               
               expect(spyCss).toHaveBeenCalledTimes(13);
-              expect(spyCss).toHaveBeenNthCalledWith(9, 'margin-top', '-28px');
+              expect(spyCss).toHaveBeenNthCalledWith(10, 'margin-top', '-28px');
             });
           })
         });
@@ -191,7 +180,7 @@ describe('Vtt Utils', () => {
   
           VttUtils.setVttCueBoxStyles(mockRegionContainer);
           
-          expect(spyCss).toHaveBeenCalledTimes(11);
+          expect(spyCss).toHaveBeenCalledTimes(12);
           expect(spyCss).toHaveBeenNthCalledWith(7, 'writing-mode', 'vertical-lr');
           expect(spyCss).toHaveBeenNthCalledWith(8, 'right', '0');
         });
@@ -205,7 +194,7 @@ describe('Vtt Utils', () => {
     
             VttUtils.setVttCueBoxStyles(mockRegionContainer);
             
-            expect(spyCss).toHaveBeenCalledTimes(11);
+            expect(spyCss).toHaveBeenCalledTimes(12);
           });
 
           it('should set percentage line positioning', () => {
@@ -216,8 +205,8 @@ describe('Vtt Utils', () => {
     
             VttUtils.setVttCueBoxStyles(mockRegionContainer);
             
-            expect(spyCss).toHaveBeenCalledTimes(12);
-            expect(spyCss).toHaveBeenNthCalledWith(9, 'right', '50%');
+            expect(spyCss).toHaveBeenCalledTimes(13);
+            expect(spyCss).toHaveBeenNthCalledWith(10, 'right', '50%');
           });
 
           it('should set positive line positioning', () => {
@@ -228,8 +217,8 @@ describe('Vtt Utils', () => {
     
             VttUtils.setVttCueBoxStyles(mockRegionContainer);
             
-            expect(spyCss).toHaveBeenCalledTimes(12);
-            expect(spyCss).toHaveBeenNthCalledWith(9, 'right', '112px');
+            expect(spyCss).toHaveBeenCalledTimes(13);
+            expect(spyCss).toHaveBeenNthCalledWith(10, 'right', '112px');
           });
 
           it('should set negative line positioning', () => {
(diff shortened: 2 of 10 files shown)


src/ts/components/subtitleoverlay.ts
```diff
@@ -67,9 +69,14 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
         this.subtitleContainerManager.removeLabel(this.previewSubtitle);
       }
 
-      this.subtitleContainerManager.addLabel(labelToAdd);
-      this.updateComponents();
       this.show();
```
stonko1994: Why moving this after the show now?

src/ts/components/subtitleoverlay.ts
```diff
@@ -453,8 +466,8 @@ export class SubtitleRegionContainerManager {
    * If the subtitle has positioning information it is added to the container.
    * @param label The subtitle label to wrap
    */
-  addLabel(label: SubtitleLabel): void {
-    const regionName = label.region || 'default';
+  addLabel(label: SubtitleLabel, overlaySize?: {width: number, height: number}): void {
```
stonko1994:
```suggestion
  addLabel(label: SubtitleLabel, overlaySize?: { width: number, height: number }): void {
```

src/ts/components/subtitleoverlay.ts
stonko1994: Why moving after show

src/ts/components/subtitleoverlay.ts
```diff
@@ -453,8 +466,8 @@ export class SubtitleRegionContainerManager {
...
+    const regionName = label.vtt && label.vtt.region ? label.vtt.region.id : label.region || 'default';
```
stonko1994: Would it make sense to add brackets here? How is this `||` interpreted if e.g. `region.id` would be undefined?

src/ts/components/subtitleoverlay.ts
```diff
@@ -474,22 +487,20 @@ export class SubtitleRegionContainerManager {
...
-    this.subtitleRegionContainers[regionName].addLabel(label);
+    this.subtitleRegionContainers[regionName].addLabel(label, overlaySize);
   }
 
   /**
    * Removes a subtitle label from a container.
    */
   removeLabel(label: SubtitleLabel): void {
-    const region = label.region || 'default';
-    for (const regionName in this.subtitleRegionContainers) {
```
stonko1994: do you know why this loop was there before?

src/ts/components/subtitleoverlay.ts
```diff
@@ -516,19 +528,80 @@ class SubtitleRegionContainer extends Container<ContainerConfig> {
...
-  addLabel(labelToAdd: SubtitleLabel) {
+  private addCueBoxContainer(labelToAdd: SubtitleLabel) {
+    const id = labelToAdd.getConfig().id;
+    this.cueBoxContainers[id] = new SubtitleCueBoxContainer();
+    this.cueBoxContainers[id].addLabel(labelToAdd);
+    setVttCueBoxStyles(this.cueBoxContainers[id], labelToAdd.vtt);
+    this.addComponent(this.cueBoxContainers[id]);
+    this.updateComponents();
+
+    if (labelToAdd.vtt.region && labelToAdd.vtt.region.scroll) {
```
stonko1994: for backwards compatiliby also check if `labelToAdd.vtt` is not empty.

stonko1994: What does scroll mean exactly? (There is no documentation in the player)

src/ts/components/subtitleoverlay.ts
```diff
@@ -505,8 +516,9 @@ export class SubtitleRegionContainerManager {
   }
 }
 
-class SubtitleRegionContainer extends Container<ContainerConfig> {
+export class SubtitleRegionContainer extends Container<ContainerConfig> {
   private labelCount = 0;
```
stonko1994: I guess the `labelCount` can be removed now

src/ts/components/subtitleoverlay.ts
```diff
@@ -516,19 +528,80 @@ class SubtitleRegionContainer extends Container<ContainerConfig> {
     }, this.config);
   }
 
-  addLabel(labelToAdd: SubtitleLabel) {
+  private addCueBoxContainer(labelToAdd: SubtitleLabel) {
+    const id = labelToAdd.getConfig().id;
+    this.cueBoxContainers[id] = new SubtitleCueBoxContainer();
```
stonko1994: Is there always just a 1-to-1 relationship as you always create a new container?

src/ts/components/subtitleoverlay.ts
```diff
@@ -516,19 +528,80 @@ class SubtitleRegionContainer extends Container<ContainerConfig> {
...
   }
 
   public isEmpty(): boolean {
     return this.labelCount === 0;
   }
-}
\ No newline at end of file
+}
+
+export class SubtitleCueBoxContainer extends Container<ContainerConfig> {
```
stonko1994: why is this needed? I think the idea of the `SubtitleRegionContainer` is exactly that.

src/ts/components/subtitlevtt.ts
```diff
@@ -0,0 +1,142 @@
+import {SubtitleRegionContainer, SubtitleCueBoxContainer} from './subtitleoverlay';
+import { VTTProperties, VTTRegionProperties } from 'bitmovin-player/types/subtitles/vtt/API';
+
+type Direction = 'top' | 'bottom' | 'left' | 'right';
+
+const lineHeight = 28;
```
stonko1994: where are those numbers coming from? Maybe add a reference to the vtt-spec.

src/ts/components/subtitlevtt.ts
```diff
@@ -0,0 +1,142 @@
+import {SubtitleRegionContainer, SubtitleCueBoxContainer} from './subtitleoverlay';
```
stonko1994:
```suggestion
import { SubtitleRegionContainer, SubtitleCueBoxContainer } from './subtitleoverlay';
```

src/ts/components/subtitlevtt.ts
```diff
@@ -0,0 +1,142 @@
...
+import { VTTProperties, VTTRegionProperties } from 'bitmovin-player/types/subtitles/vtt/API';
+
+type Direction = 'top' | 'bottom' | 'left' | 'right';
```
stonko1994: Why not using an enum here?

src/ts/components/subtitlevtt.ts
```diff
@@ -0,0 +1,142 @@
...
+    cueContainer.getDomElement().css('right', 'unset');
+    setVttPositionAlign(cueContainer, vtt, 'left');
+  } else if (vtt.position !== 'auto' && vtt.vertical != null) {
+    cueContainer.getDomElement().css('height', `${containerSize}%`);
+    setVttPositionAlign(cueContainer, vtt, 'top');
+  }
+};
+
+// https://www.w3.org/TR/webvtt1/#regions
+export const setVttRegionStyles = (regionContainer: SubtitleRegionContainer | SubtitleRegionContainer, region: VTTRegionProperties, overlaySize: {width: number, height: number}) => {
```
stonko1994: duplicate declaration (same type)

src/ts/components/subtitlevtt.ts
stonko1994: How does this play together with our subtitle customization?

src/ts/components/subtitlevtt.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  setVttWritingDirection(cueContainer, vtt);
+
+  // https://w3.org/TR/webvtt1/#webvtt-cue-text-alignment
+  const textAlign = vtt.align === 'middle' ? 'center' : vtt.align;
+  cueContainer.getDomElement().css('text-align', textAlign);
+
+  // https://w3.org/TR/webvtt1/#webvtt-cue-size
+  const containerSize = vtt.size;
+  if (vtt.position !== 'auto' && vtt.vertical == null) {
```
stonko1994: According to the player interface `vtt.vertical` is non-optional

src/ts/components/subtitleoverlay.ts
vzablock: In order to get the relative sizes and positioning calculations of the `VTTRegion` we have to show the element in the DOM to get those relative values as pixels for calculations

src/ts/components/subtitleoverlay.ts
vzablock: Not sure but @hanneslinder said that is is fine for us to remove them as they do not affect anything

src/ts/components/subtitleoverlay.ts
vzablock: Means that the bottom text pushes the top one (in our css means that we scroll the subtitle container to the bottom) and hide the subtitles on the top

src/ts/components/subtitleoverlay.ts
vzablock: There is a public method using the count to remove the region container itself if there are no more labels inside

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
+### Added
+- Subtitle stlying with VTT properties and regions
```
```suggestion
- Support for regions in VTT subtitles
```

To be consistent with the TTML entry

src/ts/components/subtitleoverlay.ts
```diff
@@ -453,43 +466,65 @@ export class SubtitleRegionContainerManager {
...
+        cssClasses
       });
 
-      this.subtitleRegionContainers[regionName] = regionContainer;
+      this.subtitleRegionContainers[regionContainerId] = regionContainer;
 
       if (label.regionStyle) {
         regionContainer.getDomElement().attr('style', label.regionStyle);
+      } else if (label.vtt && !label.vtt.region) {
+        regionContainer.getDomElement().css('position', 'unset');
```
stonko1994: Maybe add a short comment why this is needed.

src/ts/components/subtitleoverlay.ts
```diff
@@ -516,8 +551,18 @@ class SubtitleRegionContainer extends Container<ContainerConfig> {
     }, this.config);
   }
 
-  addLabel(labelToAdd: SubtitleLabel) {
+  addLabel(labelToAdd: SubtitleLabel, overlaySize?: {width: number, height: number}) {
     this.labelCount++;
+
+    if (labelToAdd.vtt) {
+      debugger;
```
stonko1994: Please remove debugger statement.

src/ts/components/subtitleoverlay.ts
```diff
@@ -516,8 +551,18 @@ class SubtitleRegionContainer extends Container<ContainerConfig> {
     }, this.config);
   }
 
-  addLabel(labelToAdd: SubtitleLabel) {
+  addLabel(labelToAdd: SubtitleLabel, overlaySize?: {width: number, height: number}) {
```
stonko1994:
```suggestion
  addLabel(labelToAdd: SubtitleLabel, overlaySize?: { width: number, height: number }) {
```

src/ts/components/subtitleoverlay.ts
```diff
@@ -453,43 +466,65 @@ export class SubtitleRegionContainerManager {
    * If the subtitle has positioning information it is added to the container.
    * @param label The subtitle label to wrap
    */
-  addLabel(label: SubtitleLabel): void {
-    const regionName = label.region || 'default';
-    if (!this.subtitleRegionContainers[regionName]) {
+  addLabel(label: SubtitleLabel, overlaySize?: {width: number, height: number}): void {
```
stonko1994:
```suggestion
  addLabel(label: SubtitleLabel, overlaySize?: { width: number, height: number }): void {
```

src/ts/components/subtitlevtt.ts
```diff
@@ -0,0 +1,157 @@
...
+  [Direction.Top, Direction.Bottom],
+  [Direction.Left, Direction.Right],
+  [Direction.Right, Direction.Left],
+]);
+
+/**
+ * Sets the default standardized styles for the Cue Box
+ * https://w3.org/TR/webvtt1/#applying-css-properties
+ */
+const setDefaultVttStyles = (cueContainer: SubtitleLabel) => {
```
stonko1994: I think we usually add such methods into a namespace.

src/ts/components/subtitlevtt.ts
```diff
@@ -0,0 +1,157 @@
+import { SubtitleRegionContainer, SubtitleLabel } from './subtitleoverlay';
```
stonko1994: As this is no component please move one folder level up. (Or think about introducing a helper folder)

src/ts/components/subtitlevtt.ts
```diff
@@ -0,0 +1,157 @@
...
+    cueContainer.getDomElement().css('width', `${containerSize}%`);
+    setVttPositionAlign(cueContainer, vtt, Direction.Left);
+  } else {
+    cueContainer.getDomElement().css('height', `${containerSize}%`);
+    setVttPositionAlign(cueContainer, vtt, Direction.Top);
+  }
+};
+
+// https://www.w3.org/TR/webvtt1/#regions
+export const setVttRegionStyles = (regionContainer: SubtitleRegionContainer, region: VTTRegionProperties, overlaySize: {width: number, height: number}) => {
```
stonko1994:
```suggestion
export const setVttRegionStyles = (regionContainer: SubtitleRegionContainer, region: VTTRegionProperties, overlaySize: { width: number, height: number }) => {
```

src/ts/components/subtitlevtt.ts
```diff
@@ -0,0 +1,157 @@
...
+      cueContainer.getDomElement().css(DirectionPair.get(direction), `${vtt.position}%`);
+      break;
+    default:
+      cueContainer.getDomElement().css(direction, `${vtt.position}%`);
+      cueContainer.getDomElement().css(DirectionPair.get(direction), 'auto');
+  }
+};
+
+export const setVttCueBoxStyles = (cueContainer: SubtitleLabel, vtt: VTTProperties) => {
+  console.warn(vtt);
```
stonko1994: Please remove console output

stonko1994:
@vzablock please also request a second reviewer.

src/ts/vttutils.ts
```diff
@@ -0,0 +1,164 @@
...
+  [Direction.Right, Direction.Left],
+]);
+
+/**
+ * Sets the default standardized styles for the Cue Box
+ * https://w3.org/TR/webvtt1/#applying-css-properties
+ */
+const setDefaultVttStyles = (cueContainer: SubtitleLabel) => {
+  if (cueContainer.vtt.region) {
+    cueContainer.getDomElement().css('position', 'relative');
```
hanneslinder: I think it would be better to pass the actual DOM Element instead of the `cueContainer` to all the helper functions in `vttutils` as all you ever need from it is `cueContainer.getDomElement()`.

vzablock: The problem is that the type gets lost in that case, the parameter required then would only be `DOM` and not `SubtitleLabel`

spec/vttutils.spec.ts
```diff
@@ -0,0 +1,572 @@
...
+
+describe('Vtt Utils', () => {
+  describe('Vtt Region', () => {
+    it('should set region css properties', () => {
+        const mockRegionContainer = generateSubtitleRegionContainerMock();
+        const mockDomElement = MockHelper.generateDOMMock();
+        mockRegionContainer.getDomElement = () => mockDomElement;
+        const spyCss = jest.spyOn(mockRegionContainer.getDomElement(), 'css');
+    
+        VttUtils.setVttRegionStyles(mockRegionContainer, vttRegionProps, {width: 1000, height: 800});
```
stonko1994:
```suggestion
        VttUtils.setVttRegionStyles(mockRegionContainer, vttRegionProps, { width: 1000, height: 800 });
```

spec/vttutils.spec.ts
```diff
@@ -0,0 +1,572 @@
...
+      const spyCss = jest.spyOn(mockRegionContainer.getDomElement(), 'css');
+
+      VttUtils.setVttCueBoxStyles(mockRegionContainer);
+      
+      expect(spyCss).toHaveBeenCalledTimes(11);
+      expect(spyCss).toHaveBeenNthCalledWith(8, 'text-align', 'left');
+    });
+
+    it('should not set default position', () => {
+      const mockRegionContainer = generateSubtitleCueBoxMock(false, { position: 50 } as VTTProperties);
```
stonko1994: Shouldn't that be `auto` now instead of `50`?

vzablock: The test says "it should NOT set default" 😄

PR #293, opened 2020-02-18
Clear selected CC/Subtitles on source unload

TeslaAdis:
If we playback an HLS stream twice. Notice that on the second playback, CC is not rendered although it shows as enabled in the Bitmovin UI.

The current implementation of `refreshSubtitles` that is called on `SourceUnloaded` does clear/sync `listElements`. But it does not clear `selectedItem`. Doing load or unload will constantly persist this `selectedItem`, and if we have the same subtitle `identifier` between sources, it will be rendered as `selected` in the UI. 

We are now adding `clearSubtitles()` which is being run on `PlayerEvent.SourceUnloaded`, and it will clear `listElements` and `selectedItem`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - UI hiding when actively using seek or volume slider
 - Empty background boxes with TTML subtitles on Chromecast
+- Clear subtitles list when source is unloaded
 
 ## [3.9.2]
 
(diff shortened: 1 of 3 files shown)


TeslaAdis:
> Why does `refreshSubtitles` not work correctly? An explanation is missing (e.g. internal issue).
> 
> It seems to me that this is either a bug in the player (subtitles returned after unload), or a wrong handling of the subtitles API (if the namespace is missing after unload). Either way, it looks like this additional method is not necessary and this can be solved by either fixing the player bug or updating `refreshSubtitles`.

Please take a look at the updated description. Utilizing `clearItems` makes more sense to me than modifying `refreshSubtitles`, as for the UI, we should expect to remove all subtitles/tracks when we do unload, and it clears `selectedItem` with it.

PR #294, opened 2020-03-17
Fix TTML region class names

vzablock:
Issue: TTML region subtitles were missing region class names. TTML `displayAlign = after` case styling was missing as well.

Solution: Fixed CSS configuration of the container adding the class name on subtitle region.
Added CSS case for TTML `displayAlign = after`

### DIFF
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -25,6 +25,12 @@
       left: 3em;
       right: 3em;
     }
+
+    &.#{$prefix}-subtitle-position-bottom>div {
+      position: absolute;
+      bottom: 0;
+      width: 100%;
+    }
   }
 
   .#{$prefix}-ui-subtitle-label {
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -26,10 +26,12 @@
       right: 3em;
     }
 
-    &.#{$prefix}-subtitle-position-bottom>div {
-      position: absolute;
-      bottom: 0;
-      width: 100%;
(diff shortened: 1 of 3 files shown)


src/ts/components/subtitleoverlay.ts
```diff
@@ -549,10 +549,6 @@ export class SubtitleRegionContainer extends Container<ContainerConfig> {
 
   constructor(config: ContainerConfig = {}) {
     super(config);
-
```
stonko1994: This solution will fix this particular use-case but we still have the potential for further bugs in this area as long as we use the `Object.assign` method in `mergeConfig` as the `cssClasses` is an Array which will just be overridden. E.g. it could be that customer passes in `cssClasses` to whatever component and then our default class would be gone and result in unexpected side effects.

If it's applicable I would go for a fix in `margeConfig` to properly respect the `cssClasses` `Array`.

protyposis: The problem lies within `SubtitleRegionContainer`. It sets its class `subtitle-region-container` with the `cssClasses` property instead of `cssClass` (and then inheriting the base class `ui-container` in SCSS). If you fix that root cause, this problem should be fixed too.

vzablock:
`cssClasses` are being ignored in some cases if they come from the config and in some cases when we do a merge. A new issue will be created to check why is this behaving so strangely.

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -25,6 +25,12 @@
       left: 3em;
       right: 3em;
     }
+
+    &.#{$prefix}-subtitle-position-bottom>div {
```
protyposis: Add spacing

vzablock:
> There's still two SCSS lint warnings that you introduced here.

@protyposis  Is there an easy way to check out which ones?

PR #301, opened 2020-05-25
Make seekbar snapping range configurable

aygee:
I would like to be able to control the snapping range, rather than the preset 1%. In my use case, I would like to disable it by setting it to 0.

### DIFF
--- a/src/ts/components/seekbar.ts
+++ b/src/ts/components/seekbar.ts
@@ -39,6 +39,11 @@ export interface SeekBarConfig extends ComponentConfig {
    * Used for seekBar control increments and decrements
    */
   keyStepIncrements?: { leftRight: number, upDown: number };
+
+  /**
+   * Used for seekBar marker snapping range percentage
+   */
+  snappingRange?: number;
 }
 
 /**
@@ -133,6 +138,7 @@ export class SeekBar extends Component<SeekBarConfig> {
       keyStepIncrements,
       ariaLabel: i18n.getLocalizer('seekBar'),
       tabIndex: 0,
+      snappingRange: 1,
     }, this.config);
 
     this.label = this.config.label;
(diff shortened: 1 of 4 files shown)


src/ts/components/seekbar.ts
```diff
@@ -253,6 +259,10 @@ export class SeekBar extends Component<SeekBarConfig> {
       if (this.isUiShown) {
         this.setAriaSliderValues();
       }
+
+      if (typeof uimanager.getConfig().seekbarSnappingRange !== 'undefined') {
```
bitmovin-kenny, line 263:
```suggestion
      if (typeof uimanager.getConfig().seekbarSnappingRange === 'number') {
```
would make more sense IMO

aygee:
> Please add an entry to CHANGELOG.md

@bitmovin-kenny What version should I add it to? v3.14.0?

bitmovin-kenny:
We collect new features in a `development` section until a new version is released. So something like this:

```md
## [development]

### Added
- Seekbar snapping range is now configurable
```

aygee:
@bitmovin-kenny Done. Thanks.

src/ts/components/seekbar.ts
```diff
@@ -253,6 +259,10 @@ export class SeekBar extends Component<SeekBarConfig> {
       if (this.isUiShown) {
         this.setAriaSliderValues();
       }
+
+      if (typeof uimanager.getConfig().seekbarSnappingRange === 'number') {
+        this.config.snappingRange = uimanager.getConfig().seekbarSnappingRange;
+      }
```
bitmovin-kenny, line 265: I didn't see it the first time, but we should not need to do this on every playback position update.

I would suggest moving it to after the marker config initializiation in line 389:

```
    // Add markers when a source is loaded or update when a marker is added or removed
    uimanager.getConfig().events.onUpdated.subscribe(() => {
      playbackPositionHandler();
    });
```

aygee: Good point, I didn't realise that. I've moved the code. Thanks!

aygee:
@bitmovin-kenny I've created a new PR to develop on #302. Closing this PR.

PR #304, opened 2020-06-22
Change default keyStepIncrements to 5

kazuhide:
Changed the default keyStepIncrements to 5
  - When the keyStepIncrements:leftRight value was 1, the volume slidebar couldn't be increased due to a floating point error (over 28). 
  - Setting the value to >1 solved the issue. Although the value can be customized on the app UI, I'd like to suggest to change the default value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Seekbar snapping range is now configurable
 
+### Changed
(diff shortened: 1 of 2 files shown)


felix-hoc (changes requested):
Increasing the step size would essentially change the behavior. Instead, I'd suggest to fix it by either making sure to return only integers from the Player, or looking into using `Math.round()` instead of `floor()`. But we'd have to test it to make sure it doesn't behave weirdly.

PR #306, opened 2020-07-07
Respect isViewModeAvailable in FullscreenToggleButton

stonko1994:
## Description
On our mobile SDKs `ViewMode.Fullscreen` is not always available. We have an API to check if fullscreen is available but that was not used in the `FullscreenToggleButton`.

## Fix
Respect the return value of `isViewModeAvailable` for `ViewMode.Fullscreen` and hide / show the button accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [develop]
 ### Fixed
 - Dead documentation link in README.md
+- `FullscreenToggleButton` is no longer visible if `ViewMode.Fullscreen` is not available
 
 ## [3.14.0]
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [develop]
 ### Fixed
 - Dead documentation link in README.md
-- `FullscreenToggleButton` is no longer visible if `ViewMode.Fullscreen` is not available
+- `FullscreenToggleButton` being visible although `ViewMode.Fullscreen` is not available
 
 ## [3.14.0]
 ### Added
--- a/src/ts/components/fullscreentogglebutton.ts
+++ b/src/ts/components/fullscreentogglebutton.ts
@@ -49,7 +49,7 @@ export class FullscreenToggleButton extends ToggleButton<ToggleButtonConfig> {
         }
       } else {
         if (console) {
-          console.log('PIP unavailable');
+          console.log('Fullscreen unavailable');
         }
       }
     });
--- a/src/ts/components/fullscreentogglebutton.ts
+++ b/src/ts/components/fullscreentogglebutton.ts
@@ -41,17 +41,19 @@ export class FullscreenToggleButton extends ToggleButton<ToggleButtonConfig> {
     player.on(player.exports.PlayerEvent.ViewModeChanged, fullscreenStateHandler);
 
     this.onClick.subscribe(() => {
-      if (player.isViewModeAvailable(player.exports.ViewMode.Fullscreen)) {
-        if (player.getViewMode() === player.exports.ViewMode.Fullscreen) {
-          player.setViewMode(player.exports.ViewMode.Inline);
-        } else {
-          player.setViewMode(player.exports.ViewMode.Fullscreen);
-        }
-      } else {
+      if (!player.isViewModeAvailable(player.exports.ViewMode.Fullscreen)) {
         if (console) {
           console.log('Fullscreen unavailable');
         }
+        return;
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [develop]
 ### Fixed
 - Dead documentation link in README.md
+- `FullscreenToggleButton` is no longer visible if `ViewMode.Fullscreen` is not available
```
HalfbyteHeroes, line 10:
```suggestion
- `FullscreenToggleButton` being visible although `ViewMode.Fullscreen` is not available
```

src/ts/components/fullscreentogglebutton.ts
```diff
@@ -28,17 +28,34 @@ export class FullscreenToggleButton extends ToggleButton<ToggleButtonConfig> {
       }
     };
 
+    let fullscreenAvailableHandler = () => {
+      if (player.isViewModeAvailable(player.exports.ViewMode.Fullscreen)) {
+        this.show();
+      } else {
+        this.hide();
+      }
+    };
```
HalfbyteHeroes, line 37: It seems a bit counter-intuitive to me, that a `fullscreenAvailableHandler` needs to check, whether fullscreen is available or not, as the name suggests that it is. Is this check actually necessary? If so, I think `fullscreenAvailabilityChangedHandler` would be a more expressive name.

src/ts/components/fullscreentogglebutton.ts
```diff
@@ -28,17 +28,34 @@ export class FullscreenToggleButton extends ToggleButton<ToggleButtonConfig> {
...
+
+    uimanager.getConfig().events.onUpdated.subscribe(fullscreenAvailableHandler);
```
HalfbyteHeroes, line 39: I suppose that the `onUpdated` event of the `UiManager` signals, that the general config of the UI manager has changed? In that case, `uiConfigUpdatedHandler` may be an even more expressive name for the `fullscreenAvailableHandler`.

src/ts/components/fullscreentogglebutton.ts
```diff
@@ -28,17 +28,34 @@ export class FullscreenToggleButton extends ToggleButton<ToggleButtonConfig> {
...
+          player.setViewMode(player.exports.ViewMode.Inline);
+        } else {
+          player.setViewMode(player.exports.ViewMode.Fullscreen);
+        }
       } else {
-        player.setViewMode(player.exports.ViewMode.Fullscreen);
+        if (console) {
+          console.log('Fullscreen unavailable');
+        }
       }
```
HalfbyteHeroes, line 54: With our current player code style, I'd collapse that `if` a little bit to get rid of the nested `if` s:

```ts
if (player.isViewModeAvailable(player.exports.ViewMode.Fullscreen)) {
  const targetViewMode =
    player.getViewMode() === player.exports.ViewMode.Fullscreen
      ? player.exports.ViewMode.Inline
      : player.exports.ViewMode.Fullscreen;
  
  player.setViewMode(targetViewMode);
} else if (console) {
  console.log('Fullscreen unavailable');
}
```

stonko1994: Or what about an early return 🤔 Not sure what the code-style says about that 🤔 

ATM: this is just consistent with our code / naming style what we have e.g. in the `PictureInPictureToggleButton`

HalfbyteHeroes: kk. Sure, an early return would also be possible. Is just a suggestion anyway, not a change request 😉

src/ts/components/fullscreentogglebutton.ts
stonko1994: At the moment the naming matches in my opinion. As the availability can't change (yet - pr incoming) it does exactly what it says.

This will be changed soon.

src/ts/components/fullscreentogglebutton.ts
stonko1994: Most of the time a component just needs to update one particular state when the `UIManager` signals an update. Therefore we pass the callback directly. If there are multiple callbacks which need to be called `onUpdate` I agree with you.

src/ts/components/fullscreentogglebutton.ts
stonko1994: Sure I totally agree that those nested `if`s don't read nicely.

CHANGELOG.md
stonko1994: 82881cd

src/ts/components/fullscreentogglebutton.ts
stonko1994: 251a0cd

PR #307, opened 2020-07-07
React to viewModeAvailabilityChanged event

stonko1994:
## Description
In our native SDKs the availability of a `ViewMode` can change after initialization of the player / UI. Therefore we introduced the `ViewModeAvailabilityChanged` event.

## Changes
Listen to the new event and change the visibility accordingly.
To enable backwards compatibility there is a check added if the event is available.

### DIFF
--- a/spec/helper/PlayerEventEmitter.ts
+++ b/spec/helper/PlayerEventEmitter.ts
@@ -18,8 +18,14 @@ import {
   SubtitleTrack,
   TimeChangedEvent,
   TimeShiftEvent,
-  VideoPlaybackQualityChangedEvent
+  VideoPlaybackQualityChangedEvent, ViewMode,
 } from 'bitmovin-player';
+
+// TODO: remove once available in the type definitions of the player
+export interface ViewModeAvailabilityChangedEvent extends PlayerEventBase {
+  viewMode: ViewMode;
+  available: Boolean;
+}
 
 export class PlayerEventEmitter {
   private eventHandlers: { [eventType: string]: PlayerEventCallback[]; } = {};
@@ -347,5 +353,14 @@ export class PlayerEventEmitter {
       timestamp: Date.now(),
       type: PlayerEvent.AudioChanged,
     } as AudioChangedEvent);
+  }
+
+  fireViewModeAvailabilityChangedEvent(viewMode: ViewMode, available: boolean): void {
+    this.fireEvent<ViewModeAvailabilityChangedEvent>({
+      timestamp: Date.now(),
+      viewMode: viewMode,
+      available: available,
+      type: 'viewmodeavailabilitychanged' as any,
+    });
   }
 }
--- a/spec/helper/PlayerEventEmitter.ts
+++ b/spec/helper/PlayerEventEmitter.ts
@@ -18,7 +18,8 @@ import {
   SubtitleTrack,
   TimeChangedEvent,
   TimeShiftEvent,
-  VideoPlaybackQualityChangedEvent, ViewMode,
+  VideoPlaybackQualityChangedEvent,
+  ViewMode,
 } from 'bitmovin-player';
 
 // TODO: remove once available in the type definitions of the player
--- a/src/ts/components/fullscreentogglebutton.ts
+++ b/src/ts/components/fullscreentogglebutton.ts
@@ -1,6 +1,6 @@
 import { ToggleButton, ToggleButtonConfig } from './togglebutton';
 import { UIInstanceManager } from '../uimanager';
-import { PlayerAPI, ViewMode } from 'bitmovin-player';
+import { PlayerAPI } from 'bitmovin-player';
 import { i18n } from '../localization/i18n';
 
 /**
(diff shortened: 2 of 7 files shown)


spec/helper/PlayerEventEmitter.ts
```diff
@@ -18,9 +18,15 @@ import {
   SubtitleTrack,
   TimeChangedEvent,
   TimeShiftEvent,
-  VideoPlaybackQualityChangedEvent
+  VideoPlaybackQualityChangedEvent, ViewMode,
```
jarhoax, line 21: Move to next line

src/ts/components/fullscreentogglebutton.ts
```diff
@@ -20,28 +20,32 @@ export class FullscreenToggleButton extends ToggleButton<ToggleButtonConfig> {
...
+      isFullScreenAvailable() ? this.show() : this.hide();
+    };
 
     player.on(player.exports.PlayerEvent.ViewModeChanged, fullscreenStateHandler);
 
+    // Available only in our native SDKs for now
+    if ((player.exports.PlayerEvent as any).ViewModeAvailabilityChanged) {
+      player.on(
+        (player.exports.PlayerEvent as any).ViewModeAvailabilityChanged,
+        fullscreenAvailabilityChangedHandler,
```
jarhoax, line 41: You could actually check for `event.available` here, right?

src/ts/components/pictureinpicturetogglebutton.ts
```diff
@@ -20,43 +20,48 @@ export class PictureInPictureToggleButton extends ToggleButton<ToggleButtonConfi
...
+      isPictureInPictureAvailable() ? this.show() : this.hide();
+    };
+
+    player.on(player.exports.PlayerEvent.ViewModeChanged, pictureInPictureStateHandler);
+
+    // Available only in our native SDKs for now
+    if ((player.exports.PlayerEvent as any).ViewModeAvailabilityChanged) {
+      player.on(
+        (player.exports.PlayerEvent as any).ViewModeAvailabilityChanged,
+        pictureInPictureAvailabilityChangedHandler,
```
jarhoax, line 41: Same as before

src/ts/components/fullscreentogglebutton.ts
stonko1994: Yes I could do that some reasons I decided against doing it:
- we have this callback already and call it also initially
- we are doing it like that on many other places as well
- we don't have the type of the event yet. So we would either need to declare the type somewhere or use `any`

spec/helper/PlayerEventEmitter.ts
stonko1994: c924172

src/ts/components/fullscreentogglebutton.ts
```diff
@@ -20,28 +20,32 @@ export class FullscreenToggleButton extends ToggleButton<ToggleButtonConfig> {
   configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
     super.configure(player, uimanager);
 
-    let fullscreenStateHandler = () => {
-      if (player.getViewMode() === player.exports.ViewMode.Fullscreen) {
-        this.on();
-      } else {
-        this.off();
-      }
+    let isFullScreenAvailable = () => {
```
unki2aut, line 23: You could use `const` for functions that won't change

src/ts/components/fullscreentogglebutton.ts
```diff
@@ -20,28 +20,32 @@ export class FullscreenToggleButton extends ToggleButton<ToggleButtonConfig> {
...
 
+    // Available only in our native SDKs for now
+    if ((player.exports.PlayerEvent as any).ViewModeAvailabilityChanged) {
+      player.on(
+        (player.exports.PlayerEvent as any).ViewModeAvailabilityChanged,
+        fullscreenAvailabilityChangedHandler,
+      );
+    }
+
+    uimanager.getConfig().events.onUpdated.subscribe(fullscreenAvailabilityChangedHandler);
```
jdmartins, line 45: Regarding the subscribes, should we unsubscribe them here? 
or this is done somewhere else?

stonko1994: They get cleared in `UIManager.clearEventHandlers` when the UI itself gets removed. Otherwise, I don't think there is need to unsubscribe 🤔

src/ts/components/fullscreentogglebutton.ts
stonko1994: 53e0f15

src/ts/components/fullscreentogglebutton.ts
jarhoax: Hm, k. Fine 4 me as well. Point 2 is not a valid reason imo 😉

src/ts/components/fullscreentogglebutton.ts
unki2aut: 💪

PR #310, opened 2020-07-10
Support TimelineMarkers for live streams

TeslaAdis:
### Description

Adding support for timeline markers on live streams. We are updating markers on time changed to the correct position. If markers have duration, it will be shrunk as it reaches the end of the timeshift range (It will be shrunk if added to timeshift 0 with duration).

Some refactoring was involved to make this more readable and accessible for further changes. 

Marker updates for live stream are on `timechanged` event. As this is already touching a lot of code, I did not refactor this, but it should be done somewhere in the future so we can optimize and introduce smooth marker update (To achieve this we need to be able to match marker in DOM and update its position).


src/ts/components/seekbar.ts
```diff
@@ -1074,4 +1011,130 @@ export class SeekBar extends Component<SeekBarConfig> {
...
+
+function  shouldProcessMarkers(player: PlayerAPI, uimanager: UIInstanceManager): boolean {
+  // Don't generate timeline markers if we don't yet have a duration
+  // The duration check is for buggy platforms where the duration is not available instantly (Chrome on Android 4.3)
+  const validToProcess = player.getDuration() !== Infinity || player.isLive();
+  const hasMarkers = this.uimanager.getConfig().metadata.markers.length > 0;
+
+  return validToProcess && hasMarkers;
 }
+// #endregion
```
ravloony, line 1140: As a first step in refactoring this, I think it would be a good idea to move this responsibility to a delegate class.

src/ts/components/seekbarlabel.ts
```diff
@@ -110,6 +105,12 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
       this.setThumbnail(this.player.getThumbnail(absoluteSeekTarget));
     }
 
+    if (args.marker) {
+      this.setTitleText(args.marker.marker.title);
+    } else {
+      this.setTitleText(null);
+    }
```
ravloony, line 112: How about 
```suggestion
    this.setTitleText(args.marker?.marker?.title ?? null);
```
although I would recommend refactoring `setTitleText` to take an empty string to set the title to nothing, instead of a `null`

TeslaAdis: We would need to upgrade typescript to `3.7.0` for `optional chaining`. And as the `pr` is already pretty big, that should be separate PR.

src/ts/components/seekbar.ts
```diff
@@ -408,7 +408,14 @@ export class SeekBar extends Component<SeekBarConfig> {
     if (this.config.smoothPlaybackPositionUpdateIntervalMs !== SeekBar.SMOOTH_PLAYBACK_POSITION_UPDATE_DISABLED) {
       this.configureSmoothPlaybackPositionUpdater(player, uimanager);
     }
-    this.configureMarkers(player, uimanager);
+
+    // Initialize markers
+    const timelineMarkerConfig = {
```
NikolaNikushev, line 413: Nit:  would move these 3 lines of code in a separate function by itself instead of adding the additional comment on top

NikolaNikushev: In the future-future,  even a refactor of the top bit and moving the top lines into their own separate functions as well

stonko1994:
Please don't post internal links in the PR description. This repo is open source.

src/ts/components/seekbarlabel.ts
ravloony: What about the second suggestion?

TeslaAdis: I mean sure, just trying to keep this PR as small as possible :D keep seeing stuff to refactor, it is driving me insane :D

stonko1994: The reason why we always try to keep changes as small as possible as that others are forking this repo and we never know what they are changing. So to avoid merge conflicts we usually don't change stuff like that.

src/ts/components/timelinemarkers.ts
```diff
@@ -0,0 +1,208 @@
...
+import { Timeout } from '../timeout';
+
+export interface MarkersConfig extends ComponentConfig {
+  /**
+   * Used for seekBar marker snapping range percentage
+   */
+  snappingRange?: number;
+}
+
+export class TimelineMarkers extends Component<MarkersConfig> {
```
stonko1994, line 17: I'm not fully happy with the name of this class.

So this is now a new `component` holding all markers right? Although it does not render itself? The seekbar is still handling / rendering the markers right?

That indicates to me that this is not a `Component`.

This looks like a `TimelineMarkersHelper` / `TimelineMarkersHandler` or something like that.

It also doesn't need to extend the `Component` as it's not added to the `Components` tree. That's not how this is designed to work.

src/ts/components/timelinemarkers.ts
```diff
@@ -0,0 +1,208 @@
...
+  private markersContainer: DOM;
+  private timelineMarkers: SeekBarMarker[];
+  private player: PlayerAPI;
+  private uimanager: UIInstanceManager;
+  private pausedTimeshiftUpdater: Timeout;
+  private getSeekBarWidth: () => number;
+  protected config: MarkersConfig;
+
```
stonko1994, line 25: Remove empty line.

src/ts/components/timelinemarkers.ts
```diff
@@ -0,0 +1,208 @@
...
+    this.config = config;
+    this.getSeekBarWidth = getSeekBarWidth;
+    this.markersContainer = markersContainer;
+  }
+
+  public configure(player: PlayerAPI, uimanager: UIInstanceManager) {
+    this.player = player;
+    this.uimanager = uimanager;
+    this.configureMarkers();
+
```
stonko1994, line 38: Remove empty line.

src/ts/components/timelinemarkers.ts
```diff
@@ -0,0 +1,208 @@
...
+    this.player.on(this.player.exports.PlayerEvent.SourceLoaded, () => {
+      if (this.player.isLive()) {
+        // Update marker position as timeshift range changes
+        this.player.on(this.player.exports.PlayerEvent.TimeChanged, () => this.updateMarkers());
+        // Update marker postion when paused as timeshift range changes
+        this.configureLivePausedTimeshiftUpdater(() => this.updateMarkers());
+      }
+    });
+    this.uimanager.getConfig().events.onUpdated.subscribe(() => this.updateMarkers());
+    this.uimanager.onRelease.subscribe(() => this.uimanager.getConfig().events.onUpdated.unsubscribe(() => this.updateMarkers()));
```
stonko1994, line 56: The main reason why we use inline arrow functions everywhere is that we can pass them without a 'wrapper' arrow function to the `PlayerEventCallback`

src/ts/components/timelinemarkers.ts
```diff
@@ -0,0 +1,208 @@
...
+          && percentage <= marker.position + snappingRange) {
+          return marker;
+        }
+      }
+    }
+
+    return null;
+  }
+
+  private clearMarkers = (): void  => {
```
stonko1994, line 84: Either do arrow functions or not everywhere.

src/ts/components/timelinemarkers.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  private updateMarkers(): void {
+    this.clearMarkers();
+
+    if (!shouldProcessMarkers(this.player, this.uimanager)) {
+      return;
+    }
+
+    for (let marker of this.uimanager.getConfig().metadata.markers) {
```
stonko1994, line 100:
```suggestion
    this.uimanager.getConfig().metadata.markers.forEach((marker) => ...
```

src/ts/components/timelinemarkers.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  private updateMarkersDOM(): void {
+    // TODO: For live streams we are removing all markers, and re-adding them.
+    // we should track markers by some internal IDs and just update their positions
+    // This would give us ability to make their pos updating smooth with css transition
+    this.markersContainer.empty();
+
+    const seekBarWidthPx = this.getSeekBarWidth();
+
+    for (let marker of this.timelineMarkers) {
```
stonko1994, line 122:
```suggestion
    this.timelineMarkers.forEach(...)
```

src/ts/components/timelinemarkers.ts
```diff
@@ -0,0 +1,208 @@
...
+        'data-marker-title': String(marker.marker.title),
+      }).css(cssProperties));
+    }
+  }
+
+  private configureLivePausedTimeshiftUpdater(
+    handler: () => void,
+  ): void {
+    // Regularly update the marker position while the timeout is active
+    this.pausedTimeshiftUpdater = new Timeout(1000, handler, true);
```
stonko1994, line 148: Do we really need two of those? 🤔 One here and the very same is still in the `seekbar.ts`

src/ts/components/timelinemarkers.ts
```diff
@@ -0,0 +1,208 @@
...
+        this.pausedTimeshiftUpdater.start();
+      }
+    });
+
+    // Stop updater when playback continues (no matter if the updater was started before)
+    this.player.on(this.player.exports.PlayerEvent.Play, () => this.pausedTimeshiftUpdater.clear());
+  }
+}
+
+function getMarkerPositions(player: PlayerAPI, marker: TimelineMarker) {
```
stonko1994, line 161: We usually don't do this top-level functions (yet) but I guess it's fine 👍

src/ts/components/timelinemarkers.ts
```diff
@@ -0,0 +1,208 @@
...
+    // This would give us ability to make their pos updating smooth with css transition
+    this.markersContainer.empty();
+
+    const seekBarWidthPx = this.getSeekBarWidth();
+
+    for (let marker of this.timelineMarkers) {
+      const markerClasses = ['seekbar-marker'].concat(marker.marker.cssClasses || [])
+        .map(cssClass => this.prefixCss(cssClass));
+
+      const cssProperties: {[propertyName: string]: string} = {
```
stonko1994, line 126:
```suggestion
      const cssProperties: { [propertyName: string]: string } = {
```

src/ts/components/timelinemarkers.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  return duration - (PlayerUtils.getSeekableRange(player).end - marker.time);
+}
+
+function getDuration (player: PlayerAPI): number {
+  if (!player.isLive()) {
+    return player.getDuration();
+  }
+  const {start, end} = PlayerUtils.getSeekableRange(player);
```
stonko1994, line 192:
```suggestion
  const { start, end } = PlayerUtils.getSeekableRange(player);
```

src/ts/playerutils.ts
```diff
@@ -59,6 +59,29 @@ export namespace PlayerUtils {
     return player.getSeekableRange() && player.getSeekableRange().start || defaultValue;
   }
 
+  /**
+   * Returns seekable time range.
+   * For live streams it will calculate timestamp timerange if tweak
+   * `enable_seek_for_vod` is not present
```
stonko1994, line 65: I don't know what `enable_seek_for_vod` means but I guess you meant `enable_seek_for_live`.

src/ts/playerutils.ts
```diff
@@ -59,6 +59,29 @@ export namespace PlayerUtils {
...
+   */
+  export function getSeekableRange(player: PlayerAPI): TimeRange {
+    let {start, end} = player.getSeekableRange();
```
stonko1994, line 68: The mobile SDKs will not return anything here.

Please see `getSeekableRangeStart` and respect accordingly.

src/ts/playerutils.ts
```diff
@@ -59,6 +59,29 @@ export namespace PlayerUtils {
...
+    const currentTimeshift = -player.getTimeShift();
+    const maxTimeshift = -player.getMaxTimeShift();
+    const currentTime = player.getCurrentTime();
+
+    end = currentTime + (currentTimeshift);
+    start = currentTime - (maxTimeshift - currentTimeshift);
+
+    return {start, end};
+  }
+
```
stonko1994, line 84: I don't fully get why this is needed 🤔 Shouldn't the player `player.getSeekablerange()` return what we need 🤔

spec/playerutils.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { PlayerUtils } from '../src/ts/playerutils';
+import { PlayerAPI } from 'bitmovin-player';
+
+let PlayerMockClass: jest.Mock<PlayerAPI> = jest.fn().mockImplementation(() => ({
+  isLive: jest.fn(),
+  getSeekableRange: jest.fn().mockReturnValue({ start: -1, end: -1 }),
+  getTimeShift: jest.fn(),
+  getMaxTimeShift: jest.fn(),
+  getCurrentTime: jest.fn(),
+}));
```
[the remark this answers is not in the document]
TeslaAdis: Adapted in https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/12a949b0831c85de9325986278d9b0f0a70e5c30

spec/playerutils.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+describe('PlayerUtils', () => {
+  let player: PlayerAPI;
```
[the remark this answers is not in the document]
TeslaAdis: Adapted in https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/12a949b0831c85de9325986278d9b0f0a70e5c30

spec/playerutils.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  beforeEach(() => {
+    player = new PlayerMockClass;
+  });
+
+  describe('getSeekableRange', () => {
+    it('it should return seekable range from playerAPI if available', () => {
+      const seekableRange = { start: 0, end: 15 };
+      player.getSeekableRange = jest.fn().mockReturnValue(seekableRange);
```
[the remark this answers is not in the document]
TeslaAdis: Adapted in https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/12a949b0831c85de9325986278d9b0f0a70e5c30

spec/playerutils.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    ${0}      | ${-260}       | ${1594646367} | ${1594646107} | ${1594646367}
+    ${-25}    | ${-260}       | ${1594646367} | ${1594646132} | ${1594646392}
+    ${-60}    | ${-260}       | ${1594646367} | ${1594646167} | ${1594646427}
+  `(
+      'should calculate start=$expectedStart and end=$expectedEnd seekable range',
+      ({ timeshift, maxTimeshift, currentTime, expectedStart, expectedEnd }) => {
+        player.isLive = jest.fn().mockReturnValue(true);
+        player.getTimeShift = jest.fn().mockReturnValue(timeshift);
+        player.getMaxTimeShift = jest.fn().mockReturnValue(maxTimeshift);
+        player.getCurrentTime = jest.fn().mockReturnValue(currentTime);
```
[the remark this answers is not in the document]
TeslaAdis: Adapted in https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/12a949b0831c85de9325986278d9b0f0a70e5c30

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Dead documentation link in README.md
 - `FullscreenToggleButton` being visible although `ViewMode.Fullscreen` is not available
 
+### Added
+- Support for `TimelineMarkers` on live streams
```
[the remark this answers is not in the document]
TeslaAdis: Adapted in https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/183c0ec91cdbee1fda3da2ef47d4fc58b5ff9a50

src/ts/components/seekbar.ts
```diff
@@ -369,9 +369,9 @@ export class SeekBar extends Component<SeekBarConfig> {
     });
     let timeShiftDetector = new PlayerUtils.TimeShiftAvailabilityDetector(player);
     timeShiftDetector.onTimeShiftAvailabilityChanged.subscribe((sender, args: TimeShiftAvailabilityChangedArgs) => {
-        hasTimeShift = args.timeShiftAvailable;
-        switchVisibility(isLive, hasTimeShift);
-      },
+      hasTimeShift = args.timeShiftAvailable;
+      switchVisibility(isLive, hasTimeShift);
+    },
     );
```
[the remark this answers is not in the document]
TeslaAdis: Auto-format acts weird o0

src/ts/components/timelinemarkers.ts
TeslaAdis: https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/61a805b60a52d8f5cfe13b0a69ae598ddb34abc1

src/ts/playerutils.ts
TeslaAdis: Interestingly enough, it will return `start -1 and end -1` for live streams, unless we have `enable_seek_for_live` enabled. So as a fallback we need to make sure to get these values somehow.

src/ts/playerutils.ts
TeslaAdis: Hmm, I can do an additional check for covering `mobile SDK`, but not sure about using `getSeekableRangeStart`, it returns 0 by default? In what cases do we not get `start` and `end` here?

src/ts/components/timelinemarkers.ts
TeslaAdis: Well, `seekbar` is holding container, but `TimelineMarkers` are the ones responsible for filling in that container. 
I was not sure of my self. I wanted `TimelineMarkers` to be handling all of this, but it was not possible :/ 
Will create it as a helper and define `prefixCss` inside, to cover that as well.

TeslaAdis: Done in https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/a5e96ef084255a89445df0900476b6978394f847 and https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/dd1b366fa704601e44eaa0964bb8871a34d0b6af

src/ts/components/timelinemarkers.ts
TeslaAdis: It gives you same result, I guess it depends on preference.

src/ts/components/timelinemarkers.ts
TeslaAdis: Indeed, did not want to pass the method in as a parameter. I could extract its functionality outside and re-use the method?

src/ts/components/seekbarlabel.ts
```diff
@@ -110,6 +105,12 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
       this.setThumbnail(this.player.getThumbnail(absoluteSeekTarget));
     }
 
+    if (args.marker) {
+      this.setTitleText(args.marker.marker.title);
+    } else {
+      this.setTitleText(null);
```
NikolaNikushev, line 111: shouldn't this one be '' as the default for `setTitleText(text = '')` as the type is expecting a string?  
or just this.setTitleText()

src/ts/components/timelinemarkershandler.ts
```diff
@@ -0,0 +1,209 @@
...
+        this.timelineMarkers.push({ marker, position: markerPosition, duration: markerDuration });
+      }
+    });
+
+    // Populate the timeline with the markers
+    this.updateMarkersDOM();
+  }
+
+  private updateMarkersDOM(): void {
+    // TODO: For live streams we are removing all markers, and re-adding them.
```
NikolaNikushev, line 112: Can you add a jira task for the follow up?

TeslaAdis: Will communicate out if needed.

src/ts/playerutils.ts
TeslaAdis: Is there any documentation I can find regarding this on mobile SDK?

src/ts/components/seekbarlabel.ts
mimcz: Perhaps good old ternary is more compact :)
```ts
args.marker
  ? this.setTitleText(args.marker.marker.title)
  : this.setTitleText(null)
```

src/ts/components/timelinemarkershandler.ts
```diff
@@ -0,0 +1,209 @@
...
+    if (this.timelineMarkers.length > 0) {
+      this.timelineMarkers.forEach(marker => {
+        // Handle interval markers
+        if (marker.duration > 0
+          && percentage >= marker.position - snappingRange
+          && percentage <= marker.position + marker.duration + snappingRange) {
+          return marker;
+        }
+        // Handle position markers
+        else if (percentage >= marker.position - snappingRange
```
mimcz, line 71: nit: No need for `else` as there is return before

src/ts/components/timelinemarkershandler.ts
```diff
@@ -0,0 +1,209 @@
...
+    return null;
+  }
+
+  private clearMarkers(): void {
+    this.timelineMarkers = [];
+    this.markersContainer.empty();
+  }
+
+  private removeMarker(marker: TimelineMarker): void {
+    this.uimanager.getConfig().metadata.markers = this.uimanager.getConfig().metadata.markers.filter(_marker => marker !== _marker);
```
mimcz, line 87: Do we really want to remove marker from the configuration it self? I think we should rather map it for internal manipulation.

src/ts/components/timelinemarkershandler.ts
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  private updateMarkers(): void {
+    this.clearMarkers();
+
+    if (!shouldProcessMarkers(this.player, this.uimanager)) {
+      return;
+    }
+
+    this.uimanager.getConfig().metadata.markers.forEach(marker => {
```
mimcz, line 97: Same here. We can probably just `map` and `filter` it?

src/ts/components/timelinemarkershandler.ts
TeslaAdis: This is a single point of truth for all markers in this codebase.

src/ts/components/timelinemarkers.ts
stonko1994: 👍

src/ts/components/timelinemarkers.ts
stonko1994: We could introduce another helper class which is handling this. 🤔 

What about adding a `PausedTimeshiftUpdatedHandler` where you can attach listeners. So the seekbar and this component would attach a listener 🤔 That would result in just having one `Timeout`.

But not sure if this isn't an overkill here 😄

src/ts/playerutils.ts
stonko1994: No there is no documentation. 

But mobile SDK versions don't even implement `getSeekableRange`. They will return `undefined` which will return in a crash here.

TeslaAdis: I see, will go with simple `!isLive()` check to return `seekableRange`, otherwise, it will get calculated. This calculation is only possible for live streams and is used for markers on live stream anyway.

src/ts/components/timelinemarkers.ts
TeslaAdis: Little overkill, but sure, I will make it as a `singleton` class.

spec/playerutils.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+      const utilSeekableRange = PlayerUtils.getSeekableRange(playerMock);
+
+      expect(utilSeekableRange).toEqual(seekableRange);
+    });
+
+    test.each`
+    timeshift | maxTimeshift | currentTime    | expectedStart  | expectedEnd
+    ${0}      | ${-260}       | ${1594646367} | ${1594646107} | ${1594646367}
+    ${-25}    | ${-260}       | ${1594646367} | ${1594646132} | ${1594646392}
+    ${-60}    | ${-260}       | ${1594646367} | ${1594646167} | ${1594646427}
```
stonko1994, line 26: nit: Using smaller numbers for `currentTime` would simplify the `expected` value calculation.

src/ts/components/seekbar.ts
```diff
@@ -408,7 +407,18 @@ export class SeekBar extends Component<SeekBarConfig> {
...
+    this.initializeTimelineMarkers(player, uimanager);
+  }
+
+  private initializeTimelineMarkers(player: PlayerAPI, uimanager: UIInstanceManager): void {
+    const timelineMarkerConfig = {
+      cssPrefix: this.config.cssPrefix,
+      snappingRange: this.config.snappingRange,
+    };
+    this.timelineMarkersHandler = new TimelineMarkersHandler(timelineMarkerConfig, () => this.seekBar.width(), this.seekBarMarkersContainer);
+    this.timelineMarkersHandler.configure(player, uimanager);
```
stonko1994, line 421: You pass the `player` and `uiManager` now in the constructor.

src/ts/playerutils.ts
```diff
@@ -59,6 +59,29 @@ export namespace PlayerUtils {
     return player.getSeekableRange() && player.getSeekableRange().start || defaultValue;
   }
 
+  /**
+   * Returns seekable time range.
+   * For live streams it will calculate timestamp timerange if tweak
+   * `enable_seek_for_live` is not present
```
stonko1994, line 65: Maybe shortly describe why you need this in case this tweak isn't set. Like you described in the PR comment

TeslaAdis:
> I quickly manually tested it and the marker is displayed properly. But when hovering over it the title is not displayed and the snapping to it does also not work. 🤔

Does not like this commit https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/2890a34e6dbf06ef03b9c14f2bdce741c6073862 :(

src/ts/playerutils.ts
TeslaAdis: Updated https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/2e8de6c57c7fd4553f633acbe01eea708f9bbe4d

src/ts/components/timelinemarkers.ts
TeslaAdis: Introduced in https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/2f59f252295301384efae0529a6e626050f6e5c5 and https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/c82874e6b323c0b4d7e2fb87e1ddc8c299e9e9c4

src/ts/components/seekbar.ts
TeslaAdis: Trying to keep the number of parameters to maximum 3, or are we OK with passing more. Feel like having `configure` seems ok here to avoid this problem.

src/ts/playerutils.ts
TeslaAdis: Updated https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/99ce8795e5818a4a50470324265f1b5a7e0840e4

spec/playerutils.spec.ts
TeslaAdis: I try to use expected numbers (in this case timestamps) in these types of unit tests.

TeslaAdis:
> I quickly manually tested it and the marker is displayed properly. But when hovering over it the title is not displayed and the snapping to it does also not work. 🤔

Went bonkers with replacing `forEach`, should be fixed now 🙈

src/ts/components/seekbar.ts
stonko1994: I think we are ok with passing more. However fine 4 me but then please rename the method. `configure` is our 'configure' method for real components and one might get confused here. Maybe rename to `setup` or `initialize` but still I would move them into the constructor.

src/ts/pausedtimeshiftupdatehandler.ts
```diff
@@ -0,0 +1,58 @@
...
+  private constructor() {}
+
+  public static getInstance(player: PlayerAPI) {
+    if (!this.instance) {
+      this._player = player;
+      this.instance = new PausedTimeshiftUpdatedHandler();
+    }
+
+    return this.instance;
+  }
```
stonko1994: We can't do that 🤔 What if there are two player instances on one page (two UI instances). As they might play different content we would have the same handler (same player) 🤔

src/ts/components/seekbar.ts
TeslaAdis: https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/55763804214bfb6e2faa0ade8a2ce3455b51d5b8

src/ts/pausedtimeshiftupdatehandler.ts
```diff
@@ -0,0 +1,58 @@
+import { PlayerAPI } from 'bitmovin-player';
+import { Timeout } from './timeout';
+
+export class PausedTimeshiftUpdatedHandler {
+  private static instance: PausedTimeshiftUpdatedHandler;
+  private static _player: PlayerAPI;
+  private pausedTimeshiftUpdater: Timeout;
+  private handlers: CallableFunction[] = [];
```
stonko1994: I was more thinking about using our `EventDispatcher` implementation for this 🤔 (sry should have been more clear on that)

src/ts/playerutils.ts
```diff
@@ -60,26 +60,25 @@ export namespace PlayerUtils {
...
    */
-  export function getSeekableRange(player: PlayerAPI): TimeRange {
-    let {start, end} = player.getSeekableRange();
-
-    // If it is VOD or we have `enable_seek_for_vod` enabled
-    if (!player.isLive() || end > -1 && start > -1) {
-      return {start, end};
+  export function getSeekableRangeForLive(player: PlayerAPI): TimeRange {
+    if (!player.isLive()) {
+      return player.getSeekableRange();
```
stonko1994, line 71: will still return `undefined` on mobile SDKs. You need to check if this is implemented and I don't know return an empty object or `0` values 😕 Whatever works for the seekbar.

src/ts/playerutils.ts
```diff
@@ -60,26 +60,25 @@ export namespace PlayerUtils {
...
+   *
+   * @param player
    */
-  export function getSeekableRange(player: PlayerAPI): TimeRange {
-    let {start, end} = player.getSeekableRange();
-
-    // If it is VOD or we have `enable_seek_for_vod` enabled
-    if (!player.isLive() || end > -1 && start > -1) {
-      return {start, end};
+  export function getSeekableRangeForLive(player: PlayerAPI): TimeRange {
```
stonko1994, line 69: naming suggestion: `getSeekableRangeRespectingLive` as it covers `VOD` and `Live` case. Still, implicit knowledge needed 🤔

src/ts/pausedtimeshiftupdatehandler.ts
TeslaAdis: I mean I can extend it so it uses event dispatcher, but doesn't that just bring in more complexity to this?

src/ts/playerutils.ts
TeslaAdis: Well, it is basically supposed to work only for live, that is some fallback there. In this implementation, we are using it on markers only if we are live as well, so we should not be facing any problems here, as it will always go through calculation.

src/ts/pausedtimeshiftupdatehandler.ts
TeslaAdis: Correct, reverted.

src/ts/playerutils.ts
TeslaAdis: Done in https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/b288dd3592345f25efc4f4da516e4e526a79da49

src/ts/components/seekbarlabel.ts
ravloony: Updated TS in https://github.com/bitmovin/bitmovin-player-ui/pull/312

src/ts/components/timelinemarkershandler.ts
TeslaAdis: Refactored in https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/3a8e0f67034bd8f6e735f0ae54c8f15fec0ac5e4

TeslaAdis:
Config to test with -- >

<details>
  <summary>HTML</summary>

  ```html
<!DOCTYPE html>
<html lang="en">
<head>
    <title>Bitmovin Player UI Demo</title>
    <meta charset="UTF-8"/>
    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">

    <!-- Bitmovin Player -->
    <script type="text/javascript" src="https://cdn.bitmovin.com/player/web/8/bitmovinplayer.js"></script>

    <!-- UI -->
    <script src="js/bitmovinplayer-ui.js"></script>
    <!-- <script src="https://cdn.bitmovin.com/player/web/8/bitmovinplayer-ui.js"></script> -->
    <link rel="stylesheet" href="css/bitmovinplayer-ui.css">

    <style>
        .player-wrapper {
            width: 95%;
            margin: 20px auto;
            box-shadow: 0 0 30px rgba(0,0,0,0.7);
        }
    </style>
</head>
<body>
<div class="container">
    <div class="content">
        <a href="index.html"><- Back to playground</a>
        <div class="player-wrapper">
            <div id="player"></div>
        </div>
    </div>
</div>
<script type="text/javascript">
    var conf = {
        key: '',
        ui: false,
        playback: {
            muted: true,
            autoplay: true,
        },
        tweaks: {
          // enable_seek_for_live: true,
        }
    };

    var source3 = {
      dash: 'https://akam.stream.peacocktv.com/live/vc1018c8297d1/dash/ctr-ua/master.mpd',
      drm: {
        widevine: {
          LA_URL: 'https://ovp.peacocktv.com/drm/widevine/acquirelicense?bt=7357-4WrTeKFu1U6Fxm65ObsXzyLolk3Zq0m8hnsYTUusntBXyCGjxPXKeX7OwXjEnBR-nFL0iqTcE2nOvJfAkW2fLrkDBOtdskjcsfELnnbkC2pymCGU0ri8b5VVzBwy0NKJeHt1vPx83T2U_AI-7rtX8C6WOx1y4BSzcGTeslLv5GezIij11MZJM-13Sg3_RALUYVAKaxTAMsgw1baXjnRnpN27sKovwtzw-sZTbHm-rNzsIbqeUEJG27D_tPUsDP9vUqUet75yEQ=='
        }
      }
    }

    var player = new bitmovin.player.Player(document.getElementById('player'), conf);
    var uiManager = bitmovin.playerui.UIFactory.buildDefaultUI(player, {});
    player.load(source3);

    player.on('ready', () => {
      const currentTime = player.getCurrentTime();
      const maxTimeshift = -player.getMaxTimeShift();
      const markers =[
        { time: currentTime + 25, duration: 15, title: 'I come from the future!' }, // Marker outside timeshift range
        { time: currentTime - 25, duration: 5 ,title: 'Tree' },
        { time: currentTime - 125, title: 'Two' },
        { time: currentTime - maxTimeshift + 10, duration: 15, title: 'One!' },
      ];

      markers.forEach(marker => {
        uiManager.addTimelineMarker(marker);
      });
    });

</script>
</body>
</html>

  ```
</details>

spec/components/timelinemarkershandler.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  it('should return TimelineMarker for LIVE', () => {
+    const expectedMarker = getTimelineMarker(180);
+    uimanagerMock.getConfig().metadata.markers.push(expectedMarker);
+    jest.spyOn(playerMock, 'getCurrentTime').mockReturnValue(200);
+    jest.spyOn(playerMock, 'getTimeShift').mockReturnValue(0);
+    jest.spyOn(playerMock, 'getMaxTimeShift').mockReturnValue(-100);
+    jest.spyOn(playerMock, 'isLive').mockReturnValue(true);
+    jest.spyOn(playerMock, 'getDuration').mockReturnValue(Infinity);
```
stonko1994, line 51: Could be extracted into a `beforeEach` block inside a own `describe('TimelineMarkers for Live')`

src/ts/components/timelinemarkershandler.ts
```diff
@@ -100,7 +99,7 @@ export class TimelineMarkersHandler {
 
       if (shouldRemoveMarker(markerPosition, markerDuration)) {
         this.removeMarker(marker);
-      } else {
+      } else if (markerPosition < 100) {
```
stonko1994, line 102:
```suggestion
      } else if (markerPosition <= 100) {
```

What if a marker is configured at `player.duration` (in a VOD of course)

src/ts/playerutils.ts
```diff
@@ -61,10 +61,12 @@ export namespace PlayerUtils {
 
   /**
    * Calculates player seekable time range for live.
-   * For live streams we calculate range since we receive {start: -1, end: -1} unless
-   * `enable_seek_for_live` is present
+   * For live streams we need to calculate range since we receive { start: -1, end: -1 }
+   * unless `enable_seek_for_live` tweak is present
```
stonko1994, line 65:
```suggestion
   * As the player returns `{ start: -1, end: -1 }` for live streams we need to calculate the `seekableRange` based on `maxTimeshift`.
```

src/ts/playerutils.ts
stonko1994: 👍

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Dead documentation link in README.md
 - `FullscreenToggleButton` being visible although `ViewMode.Fullscreen` is not available
 
+### Added
+- Support `TimelineMarkers` in live streams
+
```
stonko1994, line 14: Missed that previously. Please move the `Added` section before the `Fixed` section.

TeslaAdis: Addressed in https://github.com/bitmovin/bitmovin-player-ui/pull/310/commits/08b99dd7f70524693ae72bea92a180ff40fa4c7d

PR #311, opened 2020-07-15
Fix `npm ci` build failure on node 14

ravloony:
I have added `node-sass` as a direct dependency. This is recommended by the `gulp-sass` project.

I don't think this needs a CHANGELOG entry as it doesn't do anything to the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ The UI framework is also available in the NPM repository and comes with all sour
 ## Getting Started with Development
 
  0. Clone Git repository
- 1. Install node.js (version 10.x required)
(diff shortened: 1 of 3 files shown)


ravloony:
> We have a `README` entry that we require `node 10.x`
> 
> `1. Install node.js (version 10.x required)`
> 
> I think we could remove this now right

Looks like, as `npm ci` and the `gulp` commands now work. I'll remove it.

ravloony:
> Your branch name is missing the `feature/` prefix, setup itself works.

:see_no_evil: oops

PR #312, opened 2020-07-15
Update Typescript to 3.9.6

ravloony:
Like it says on the tin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Listen to `ViewModeAvailabilityChanged` event and toggle visibility of `FullscreenToggleButton` and `PictureInPictureToggleButton` accordingly
 
+### Changed
+- TypeScript update to 3.9.6
+
 ### Fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Listen to `ViewModeAvailabilityChanged` event and toggle visibility of `FullscreenToggleButton` and `PictureInPictureToggleButton` accordingly
 
+### Changed
+- TypeScript update to 3.9.6
```
unki2aut, line 13: Isn't this more of an `## Internal` change?

unki2aut:
/testautomation run --target "browser=chrome&deviceType=desktop"

CHANGELOG.md
ravloony: I thought so too, but there are no `## Internal` entries in this Changelog (maybe because it is public?) and the last TS upgrade was also in `## Changed`

unki2aut: Sry, didn't look at the repository name 🙈

PR #313, opened 2020-07-16
add npx build

NikolaNikushev:
The build fails because the gulp is not up-to-date.

If we add NPX it installs needed dependencies.


NikolaNikushev:
Not needed. The build was failing as it was loading from Master instead of DEVELOP

PR #333, opened 2020-10-23
Fix subtitle display issue for IE/tizen 2016 when no regions are present

kishorens:
**Description**: Captions are not displayed in IE 11/tizen 2016 when there are no regions present in WebVTT.

**Problem**: When region labels are not present for the subtitle we are removing the absolute position set to the region container using CSS `position: unset` which is not supported in IE11 and also Tizen 2016.

**Fix**: Use CSS position property `'static'` to remove the position style.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Live-indicator stays active after stalling in live streams
+- Subtitles not displayed in IE/tizen 2016 when no regions are present
 
 ## [3.17.0] - 2020-08-18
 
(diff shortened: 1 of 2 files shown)


bitmovin-kenny:
Do you have an asset ready to verify this fix?

kishorens:
> Do you have an asset ready to verify this fix?

The issue is reproducible in our [demo site](https://bitmovin.com/demos/caption-styling), you can use our demo asset to verify.

var source = {
  dash: 'https://bitmovin-a.akamaihd.net/content/sintel/sintel.mpd',
  hls: 'https://bitmovin-a.akamaihd.net/content/sintel/hls/playlist.m3u8',
  poster: 'https://bitmovin-a.akamaihd.net/content/sintel/poster.png'
};

bitmovin-kenny (approved):
Wow, just verified, that witout your fix, the subtitles don't show on IE. It's hard to believe that this went unnoticed till now. Nice catch 💪

kishorens:
Thanks for the approval @bitmovin-kenny.

kazuhide (approved):

hanneslinder (approved):

PR #335, opened 2020-11-02
Fix Vast `unknown` AdError when UI is not present

Cagince:
Object.assign is removed and now present only in bitdash.


stonko1994:
Why this change? I think the UI should add the polyfills which are needed. In theory, the UI can be used without including our player at all (it just works with the interface).

Cagince:
The polyfill we have in the UI was causing certain capabilities of `player` being dependent on the inclusion of UI, 
not sure what type of use cases there are to use bitmovin player UI without bitmovin player? If that is a valid concern we should keep it as is.

But the purpose of this PR is to put all of the polyfills in one place to avoid bugs/errors/warnings we might miss because we are using the UI or any other module, but in general  all features of player should function without UI and UI should not have any side effects on player...

Cagince:
closing the PR because there might be a use case of player-ui being used without player.

stonko1994:
> but in general all features of player should function without UI and UI should not have any side effects on player...

agree but therefor we have the `if (typeof Object.assign !== 'function') {` check. So it would be pretty easy to add this in the player as well.

PR #337, opened 2020-11-10
Add button type to button component

kishorens:
Description: In most modern browsers (with the exception of IE), the default button type is submit this causes an issue when we embed our player in a form, on play button press the form is submitted.

Problem: button type needs to be defined, so that browser will not consider it as submit. 

Fix: Add  `'type' : 'button'` to button component

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+
+## develop
+
 ### Added
-- Add `'type' : 'button'` to buttons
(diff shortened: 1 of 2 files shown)


ravloony (approved):

hanneslinder (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+### Added
+- Add `'type' : 'button'` to buttons
+
```
stonko1994, line 9: We usually omit the redundant `Add` keyword in the `Added` section. (And the `develop` header is missing)

```suggestion

## develop

### Added
- `'type' : 'button'` to buttons to prevent submitting forms

```

flofischer (approved):

PR #340, opened 2020-11-12
Fix CI build errors

jmsn:
Fixing two build-breaking errors:

1. https://github.com/bitmovin/bitmovin-player-ui/pull/312 broke wrapping getters and setters in `uimanager.ts` and made the corresponding unit tests fail.
  The reason is that since TypeScript 3.9 getters and setters are no longer enumerable properties and thus the `for...in` loop we are using does not iterate over such properties anymore.
  https://github.com/bitmovin/bitmovin-player-ui/blob/f7847b64f6323a8a027cf7f998bf3c116857ace8/src/ts/uimanager.ts#L771-L776
  Side note: Currently all getters/setters on the player object are actually defined as enumerable properties which means the cases that the broken unit tests are covering are currently not relevant, they might have been in the past or become in the future.

2. https://github.com/bitmovin/bitmovin-player-ui/pull/307 had an [import error](https://travis-ci.org/github/bitmovin/bitmovin-player-ui/builds/743004941) in a spec file

### DIFF
--- a/src/ts/uimanager.ts
+++ b/src/ts/uimanager.ts
@@ -771,19 +771,7 @@ export class PlayerWrapper {
     // Collect all members of the player (public API methods and properties of the player)
     const objectProtoPropertyNames = Object.getOwnPropertyNames(Object.getPrototypeOf({}));
     const namesToIgnore = ['constructor', ...objectProtoPropertyNames];
-    const getAllPropertyNames = ((target: PlayerAPI) => {
-      let names: string[] = [];
-
-      while (target) {
-        names = names.concat(Object.getOwnPropertyNames(target).filter(name => namesToIgnore.indexOf(name) === -1));
-        // go up prototype chain
-        target = Object.getPrototypeOf(target);
-      }
-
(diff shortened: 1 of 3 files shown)


Cagince (approved):

src/ts/uimanager.ts
```diff
@@ -902,3 +899,15 @@ export class PlayerWrapper {
   }
 }
 
+function getAllPropertyNames(target: PlayerAPI): string[] {
```
stonko1994, line 902: Would it make sense to enable it for 'any-object' (in case it is needed for a different type than the `PlayerAPI`)
```suggestion
function getAllPropertyNames(target: Object): string[] {
```

stonko1994 (approved):

fabiandev (approved):
Should we also add a test for that?

jmsn:
> Should we also add a test for that?

Well, the fix is covered by the unit test that caused the CI build failure so should be fine IMHO 🙂

PR #350, opened 2020-12-15
Configurable hide UI when hovered

luckygoyal-bitmovin:
**Issue**
Currently controls bar auto hides in hide delay seconds even if user is hovering over any of the elements like settings panel etc. There is no configuration option to disable auto hiding in this case. 

For a little background, the default behaviour was disabling auto hide when hovered but this was changed with #272 

The reason for this change was behaviour on touch devices(Android, iOS) where tapping on an element in control bar would make it stay forever until user again taps to dismiss it. But at the same time the behaviour changed for non-touch devices like Desktop causing controls to auto hide.

**Fix**
- Bring in changes removed in #272 + additionally limit the change to non-mobile devices
- Add a configuration option to allow disabling auto hiding on hover instead of making it default behaviour. Default 
   behaviour still remains to auto hide.
- The mobile devices are already good with default behaviour as some of important elements like SettingsPanel 
   that would benefit with this configuration are configured to not auto-hide by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## develop
 
 ### Added
-- Add `UIConfig#disableAutoHideWhenHovered` config property to disable auto hiding of UI when an element in `ControlBar` is currently hovered
+- `UIConfig#disableAutoHideWhenHovered` config property to disable auto hiding of UI when an element in `ControlBar` is currently hovered
 
 ### Changed
 - Exported builder method for `modernUI`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## develop
 
+### Added
+- Add `UIConfig#disableAutoHideWhenHovered` config property to disable auto hiding of UI when an element in `ControlBar` is currently hovered
+
 ### Changed
 - Exported builder method for `modernUI`
 
--- a/src/ts/components/controlbar.ts
+++ b/src/ts/components/controlbar.ts
@@ -72,7 +72,5 @@ export class ControlBar extends Container<ControlBarConfig> {
     uimanager.onControlsHide.subscribe(() => {
       this.hide();
     });
-
-
   }
 }
--- a/src/ts/components/controlbar.ts
+++ b/src/ts/components/controlbar.ts
@@ -36,28 +36,26 @@ export class ControlBar extends Container<ControlBarConfig> {
     // Counts how many components are hovered and block hiding of the control bar
     let hoverStackCount = 0;
 
-    if (uimanager.getConfig().disableAutoHideWhenHovered) {
-      // only enabling this for non-mobile platforms without touch input. enabling this
-      // for touch devices causes the UI to not disappear after the standard few seconds.
-      // Instead, it will stay visible until another manual action is performed.
-      if (!BrowserUtils.isMobile) {
-        // Track hover status of child components
-        UIUtils.traverseTree(this, (component) => {
-          // Do not track hover status of child containers or spacers, only of 'real' controls
-          if (component instanceof Container || component instanceof Spacer) {
-            return;
-          }
+    // only enabling this for non-mobile platforms without touch input. enabling this
+    // for touch devices causes the UI to not disappear after hideDelay seconds.
+    // Instead, it will stay visible until another manual interaction is performed.
+    if (uimanager.getConfig().disableAutoHideWhenHovered && !BrowserUtils.isMobile) {
+      // Track hover status of child components
+      UIUtils.traverseTree(this, (component) => {
+        // Do not track hover status of child containers or spacers, only of 'real' controls
+        if (component instanceof Container || component instanceof Spacer) {
+          return;
+        }
 
-          // Subscribe hover event and keep a count of the number of hovered children
-          component.onHoverChanged.subscribe((sender, args) => {
-            if (args.hovered) {
-              hoverStackCount++;
-            } else {
-              hoverStackCount--;
-            }
-          });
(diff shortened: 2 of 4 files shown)


src/ts/components/controlbar.ts
```diff
@@ -32,10 +33,42 @@ export class ControlBar extends Container<ControlBarConfig> {
...
     super.configure(player, uimanager);
 
+    // Counts how many components are hovered and block hiding of the control bar
+    let hoverStackCount = 0;
+
+    if (uimanager.getConfig().disableAutoHideWhenHovered) {
+      // only enabling this for non-mobile platforms without touch input. enabling this
+      // for touch devices causes the UI to not disappear after the standard few seconds.
+      // Instead, it will stay visible until another manual action is performed.
+      if (!BrowserUtils.isMobile) {
```
bitmovin-kenny, line 43: These statements can be combined:
```ts
    // only enabling this for non-mobile platforms without touch input. enabling this
    // for touch devices causes the UI to not disappear after the standard few seconds.
    // Instead, it will stay visible until another manual action is performed.
    if (uimanager.getConfig().disableAutoHideWhenHovered && !BrowserUtils.isMobile) {
      // Track hover status of child components
      UIUtils.traverseTree(this, (component) => {
        // Do not track hover status of child containers or spacers, only of 'real' controls
        if (component instanceof Container || component instanceof Spacer) {
          return;
        }

        // Subscribe hover event and keep a count of the number of hovered children
        component.onHoverChanged.subscribe((_, args) => {
          if (args.hovered) {
            hoverStackCount++;
          } else {
            hoverStackCount--;
          }
        });
      });
    }
```

src/ts/components/controlbar.ts
```diff
@@ -32,10 +33,42 @@ export class ControlBar extends Container<ControlBarConfig> {
...
+        // Track hover status of child components
+        UIUtils.traverseTree(this, (component) => {
+          // Do not track hover status of child containers or spacers, only of 'real' controls
+          if (component instanceof Container || component instanceof Spacer) {
+            return;
+          }
+
+          // Subscribe hover event and keep a count of the number of hovered children
+          component.onHoverChanged.subscribe((sender, args) => {
```
bitmovin-kenny, line 52: I think it's a common pattern to use `_` for unused parameters in the UI.

bitmovin-kenny (changes requested):
Basically looks good to me, but the code can be shortened.

Would it be possible to add unit tests for this?

src/ts/components/controlbar.ts
luckygoyal-bitmovin: Updated as suggested.

src/ts/components/controlbar.ts
luckygoyal-bitmovin: Updated as suggested

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## develop
 
+### Added
+- Add `UIConfig#disableAutoHideWhenHovered` config property to disable auto hiding of UI when an element in `ControlBar` is currently hovered
```
stonko1994, line 11: We usually omit the redundant `Add / Fixed` at the beginning of changelog entries.

```suggestion
- `UIConfig#disableAutoHideWhenHovered` config property to disable auto hiding of UI when an element in `ControlBar` is currently hovered
```

src/ts/uiconfig.ts
```diff
@@ -68,6 +68,11 @@ export interface UIConfig {
    * Default: true
    */
   playbackSpeedSelectionEnabled?: boolean;
+  /**
+   * Specifies if the player controls including `SettingsPanel` should auto hide when still hovered
+   * Default: false
+   */
+  disableAutoHideWhenHovered?: boolean;
```
stonko1994, line 75: Please add a short remark that this does not have any effect on mobile and a short description why.

OR: change the default value base on 'if-mobile' 🤔 The we would have full control about that also on mobile.

luckygoyal-bitmovin: @stonko1994 , thanks for review comment. I did not fully get the second suggestion regarding "change the default value base on 'if-mobile'". Can you please elaborate?

bitmovin-kenny (approved):
LGTM

I was thinking there might be a potential problem, that a hovered component might get removed without triggering a `onHoverChanged` event. But even if that would be possible, it's still not the default behavior.

Unit tests would still be nice though?

src/ts/uiconfig.ts
stonko1994: I meant to have `default: true` on web and `default: false` on mobile 🤔 

Doing so would allow us to remove the `&& !BrowserUtils.isMobile` in the `controlbar.ts` and gives full control over this via the config.

---

As it is right now you can set `disableAutoHideWhenHovered` to `true` on mobile but the config would be ignored on mobile at all.
This feels unnatural cause you configured it. (_not sure if a use-case though_)

luckygoyal-bitmovin: thanks @stonko1994 for clarifying. I would prefer adding a remark that it does not have any effect for mobile. 
- Removing `!BrowserUtils.isMobile` will bring back the original issue due to which auto hiding was made the default behaviour.
- Also I am not sure if we should/could use the same configuration to have different default values for mobile and non-mobile platforms.

luckygoyal-bitmovin: Updated the documentation of `disableAutoHideWhenHovered` to specify that it does not apply to mobile platforms

stonko1994: Sounds legit 👍 🙂

stonko1994 (approved):
LGTM 👍 Unit tests would be nice but I can imagine that those are not that easy for this use-case.

luckygoyal-bitmovin:
Thanks @bitmovin-kenny and @stonko1994 for the approval. Will attempt adding unit tests for `controlbar.ts ` as a follow up.

PR #361, opened 2021-02-02
Handle PlaylistTransition event for Mobile v3

stonko1994:
## Description
As the Mobile SDKs introduce a new major version that supports playlists we need to add support for this in the UI as well.

### Problem
The UI doesn't know about the playlist context for now and does not update the components.

### Solution
Listen to the `PlaylistTransition` event (introduced with mobile v3) which indicates that the playlist transitioned to a new source. When this event occurs we need to update the UI accordingly to display the data from the new source. 
To achieve this we reuse the `onUpdated` event of the `UIManager` which is already in place and dispatched in the UI when a `SourceLoaded` event occurs. We can't reuse the `SourceLoaded` event for this as this event got a new behaviour in v3 and is fired for each source individually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## develop
+
+### Added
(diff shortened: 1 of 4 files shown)


flofischer (approved):
lgtm 👍

Reevn (approved):
👍

PR #363, opened 2021-02-11
Restrict focus visibility of controls to keyboard interactions

jmsn:
## Problem

There are inconsistencies across browsers on different platforms regarding if a button element receives focus upon click, see [here](https://zellwk.com/blog/inconsistent-button-behavior/#summary-of-the-results).

This causes UI controls to show a focus indicator upon clicking on certain platforms, like Chrome on Android. This is unintended behavior as the focus indicator should only be visible upon interaction via keyboard.

## Fix

There would be the CCS pseudo-class [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#selectively_showing_the_focus_indicator) that we could use and it applies only on elements where `:focus` applies and the focus should be indicated to the user e.g. caused by keyboard interaction [and similar](https://drafts.csswg.org/selectors-4/#the-focus-visible-pseudo).https://github.com/bitmovin/bitmovin-player-ui/commit/c83e95f5e925aae30d5aa150b44b4c0977a0b6cb demonstrates the usageof that. However, it is [not supported](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#browser_compatibility) on all browsers/platforms that we need to cover.

To cover old browsers/platforms like the WebView on Android KitKat we have now have a `FocusVisibilityTracker` that puts a `.bmpui-focus-visible` CCS class on the elements similarly as the native `:focus-visible` CCS pseudo-class would apply. 

### How it works
We're basically tracking `mousedown`, `pointerdown`, `touchstart`, and `keydown` events to determine which input method was last used. Once a UI element gets focus (`focus` event) and the last interaction was a keyboard we add the `.bmpui-focus-visible` class to that element to make the focus visible. Once it loses the focus again (`blur` event) we're removing that CCS class again.

### DIFF
--- a/spec/focusvisibilitytracker.spec.ts
+++ b/spec/focusvisibilitytracker.spec.ts
@@ -19,7 +19,7 @@ describe('FocusVisibilityTracker', () => {
   });
 
   it('should add the focus-visible class on a UI button that gets focus w/o initial interaction', () => {
-    // Tinitially , the last interaction is set to keyboard to cover case of htting TAB from the adress bar which would
+    // Initially, the last interaction is set to keyboard. To cover the case of hitting TAB from the address bar which would
     // not trigger a keydown event.
     uiButton.focus();
 
--- a/spec/focusvisibilitytracker.spec.ts
+++ b/spec/focusvisibilitytracker.spec.ts
@@ -1,0 +1,96 @@
+import { JSDOM } from 'jsdom';
+import { FocusVisibilityTracker } from '../src/ts/focusvisibilitytracker';
+
+describe('FocusVisibilityTracker', () => {
+  const bitmovinUIPrefix = 'bmpui';
+  const uiButtonId = `${bitmovinUIPrefix}-some-button`;
+  const nonUiButtonId = 'some-button';
+  const focusVisibileClassName = '{{PREFIX}}-focus-visible';
+  let document: Document;
+  let uiButton: HTMLElement;
+  let nonUiButton: HTMLElement;
+  let tracker: FocusVisibilityTracker;
+
+  beforeEach(() => {
+    const dom = new JSDOM();
+    document = dom.window.document;
+    uiButton = createAndAddButton(document, uiButtonId);
+    nonUiButton = createAndAddButton(document, nonUiButtonId);
+
+    // make x instance of Element/DOMTokenList checks work
+    global.Element = dom.window.Element;
+    global.DOMTokenList = dom.window.DOMTokenList;
+
+    tracker = new FocusVisibilityTracker(document, bitmovinUIPrefix);
+  });
+
+  it('should add the focus-visible class on a UI button that gets focus w/o initial interaction', () => {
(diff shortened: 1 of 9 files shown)


stonko1994 (approved):

src/ts/main.ts
```diff
@@ -1,3 +1,4 @@
+import 'focus-visible';
```
offnertho, line 1: blank line?

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "files": [
-    "src/ts/main.ts"
+    "src/ts/main.ts",
+    "src/ts/decs.d.ts"
```
offnertho, line 4: just a question, but why is this file called `decs.d.ts` ?

wasp898 (approved):

jmsn:
@stonko1994 @offnertho @wasp898 

Thanks for the reviews so far 🙂 . However, after some internal discussion, we decided to move away from using that polyfill and went for a solution that works in a similar way but is much slimmer and tailored to our specific needs.

spec/focusvisibilitytracker.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  beforeEach(() => {
+    const { window } = new JSDOM();
+    global.document = window.document;
+    uiButton = createAndAddButton(document, uiButtonId);
+    nonUiButton = createAndAddButton(document, nonUiButtonId);
+    tracker = new FocusVisibilityTracker(bitmovinUIPrefix);
+  });
+
+  it('should add the focus-visible class on a UI button that gets focus w/o initial interaction', () => {
```
stonko1994, line 21: I know we have differences in using the `should` word in test names 😄 (Up to you if you change that - No blocker still would prefer not using it) **all occurrences**
```suggestion
  it('adds the focus-visible class on a UI button that gets focus w/o initial interaction', () => {
```

spec/focusvisibilitytracker.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+    // Tinitially , the last interaction is set to keyboard to cover case of htting TAB from the adress bar which would
```
stonko1994, line 22: Typos:
```suggestion
    // Initially, the last interaction is set to keyboard. To cover the case of hitting TAB from the address bar which would
```

stonko1994 (approved):
Manual testing looks good 👍 

_Two minor stuff - approving already_

wasp898 (approved):

felix-hoc (approved):
LGTM.
+1 to the other comment about test naming as `it('should ...` but not a biggie :)
Approved!

spec/focusvisibilitytracker.spec.ts
jmsn: done in https://github.com/bitmovin/bitmovin-player-ui/pull/363/commits/bd031ac5d07276393ed1fc86a0f980df749206e6

PR #366, opened 2021-02-16
Add CI step for publishing to npm

fabiandev:
Publishes releases to npm, e.g.:

git branch/tag            | npm channel
---------------------|---------------
`v3.25.0`                    | `latest`
`v3.25.1`                    | `latest`
`v3.25.0-b.1`             | `beta`
`v3.25.0-rc.1`            | `staging`
`nightly/2021/02/15` | `nightly`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ before_script:
 script:
   - npm test
   - gulp build-prod
+  - bash publish.sh
 cache:
   yarn: true
 env:
--- a/publish.sh
+++ b/publish.sh
@@ -1,0 +1,111 @@
+#!/bin/bash
+set -e
+
+PACKAGE_NAME="bitmovin-player-ui"
+CI_BRANCH=$TRAVIS_BRANCH
+
+if ! [[ "${CI_BRANCH}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+  echo "INFO ${CI_BRANCH} is not a valid version to be published, skipping"
+  exit 0
+fi
+
+CHANNEL=-1
+NPM_TAG=-1
+MAJOR=-1
+MINOR=-1
+POSTIFX=-1
+VERSION=-1
+
+DRY_RUN_CMD=
+if ($DRY_RUN); then
+    DRY_RUN_CMD="--dry-run"
+    echo "INFO performing a dry run"
+fi
+
+if [[ "${CI_BRANCH}" =~ ^v([0-9]+)\.([0-9]+)\.([0-9]+)-?([a-z]*) ]]; then
+    MAJOR=${BASH_REMATCH[1]}
+    MINOR=${BASH_REMATCH[2]}
+    HOTFIX=${BASH_REMATCH[3]}
+    POSTFIX=${BASH_REMATCH[4]}
+    VERSION=${CI_BRANCH:1}
+    case ${POSTFIX} in
+        "b")
+            CHANNEL="beta"
+            NPM_TAG="beta"
+            ;;
+        "rc")
+            CHANNEL="staging"
+            NPM_TAG="staging"
+            ;;
+        "")
+            CHANNEL="stable"
+            NPM_TAG="latest"
+            ;;
+        *)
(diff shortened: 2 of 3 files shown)


publish.sh
```diff
@@ -0,0 +1,101 @@
...
+MAJOR=-1
+MINOR=-1
+POSTIFX=-1
+VERSION=-1
+
+DRY_RUN_CMD=
+if ($DRY_RUN); then
+    DRY_RUN_CMD="--dry-run"
+    echo "INFO performing a dry run"
+fi
```
HalfbyteHeroes, line 20: What is the dry run used for? Assuming it is used for testing purposes, do we really want to commit it?

publish.sh
```diff
@@ -0,0 +1,101 @@
...
+function version_gt() { test "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1"; }
+
+if !$DRY_RUN && version_gt ${NPM_LATEST} ${VERSION}; then
+    # The version we just published is lower than the previously tagged version on npm, so we need to revert the
+    # tag to the previous version to avoid version downgrades (this e.g. avoids that a 7.2.5 hotfix release overwrites
+    # the latest-tagged 7.3.2)
+    echo "INFO reverting "${NPM_TAG}" tag from the just published version ${VERSION} to the greater ${NPM_LATEST}"
+    # It takes a while until the metadata after npm publish is updated so we need to wait to avoid a failed tag update
+    # "npm WARN dist-tag add latest is already set to version ${VERSION}"
+    sleep 10
```
HalfbyteHeroes, line 99: Is there a way to check, whether the metadata has already been updated? Using a timeout seems a bit fragile to me.

HalfbyteHeroes (commented):
I'm not at all familiar with bash scripting, so I can't really comment on the script itself. General workflow looks sensible to me, though 👍

publish.sh
fabiandev: I would like to leave it there, so we can set the `DRY_RUN` environment variable via CI and look at the output in production the first time w/o actually running the command

publish.sh
fabiandev: Probably 😄 We could e.g. handle this via a configurable amount of retries. However, I've based this on existing code from us and this seems to be very reliable, so I'd rather keep this for a follow-up improvement.

.travis.yml
```diff
@@ -6,6 +6,7 @@ before_script:
 script:
   - npm test
   - gulp build-prod
+  - bash publish.sh
```
offnertho, line 9: are there no packages available which takes care of publishing from e.g. gulp. 

Just wondering why we did have to write this on our own.

fabiandev: > I've based this on existing code from us and this seems to be very reliable

That's why ☝️ I'd like to get this step automated with minimal effort. I'm sure there are packages available around this though and we can definitely improve it in the future if needed.

prirania (approved):

publish.sh
HalfbyteHeroes: Ok, fair point!

publish.sh
```diff
@@ -0,0 +1,103 @@
...
+
+echo "INFO npm tag set to ${NPM_TAG}"
+if [[ ${NPM_TAG}  == -1 ]]; then
+    echo "ERROR npm tag ${NPM_TAG} not valid"
+    exit 1
+fi
+
+## Check if this version was already published.
+## If something went wrong during a later build step and we re-run the release
+## after fixing the problem, the npm publish would fail the build.
```
HalfbyteHeroes, line 59: Are those double-`#`s a typo?

publish.sh
HalfbyteHeroes: Ok 👍

publish.sh
fabiandev: I've removed the second one

HalfbyteHeroes (approved):

.travis.yml
fabiandev: @offnertho ok for you to move forward?

PR #367, opened 2021-03-03
[PW-4666] Support mobile v3 error events

mukulmishra18:
## Description

`PlayerEvent.Error` does not exist anymore in the Mobile SDKs V3. Instead of this `PlayerEvent.SourceError` and `PlayerEvent.PlayerError` event is introduced.

### Problem
The UI does not support the new Error events in Mobile SDKs

### Solution
Listen for `PlayerEvent.SourceError` and `PlayerEvent.PlayerError` along with `PlayerEvent.Error` and make sure the message is displayed correctly. Before we had mapping of error code with error message internally, but now it is not needed as the `PlayerErrorEvent` or `SourceErrorEvent` will have the message already. We have to use it directly and apply the customization config, if present.

### DIFF
--- a/spec/components/errormessageoverlay.spec.ts
+++ b/spec/components/errormessageoverlay.spec.ts
@@ -15,14 +15,14 @@ describe('ErrorMessageOverlay', () => {
       uiInstanceManagerMock = MockHelper.getUiInstanceManagerMock();
     });
   
-    it('should add event listener for source loaded event', () => {
+    it('adds an event listener for source loaded event', () => {
       const onSpy = jest.spyOn(playerMock, 'on');
       errorMessageOverlay.configure(playerMock, uiInstanceManagerMock);
 
       expect(onSpy).toHaveBeenCalledWith(playerMock.exports.PlayerEvent.SourceLoaded, expect.any(Function));
     });
 
-    it('should add event listener for error event when not mobile v3', () => {
+    it('adds an event listener for error event when not mobile v3', () => {
       const onSpy = jest.spyOn(playerMock, 'on');
       errorMessageOverlay.configure(playerMock, uiInstanceManagerMock);
 
@@ -40,14 +40,14 @@ describe('ErrorMessageOverlay', () => {
         (playerMock.exports.PlayerEvent as any).PlaylistTransition = MobileV3PlayerEvent.PlaylistTransition;
       });
 
-      it('should add event listener for sourceerror and playererror when mobile v3', () => {
+      it('adds an event listener for sourceerror and playererror when mobile v3', () => {
         errorMessageOverlay.configure(playerMock, uiInstanceManagerMock);
   
         expect(onSpy).toHaveBeenCalledWith(MobileV3PlayerEvent.PlayerError, expect.any(Function));
         expect(onSpy).toHaveBeenCalledWith(MobileV3PlayerEvent.SourceError, expect.any(Function));
       });
   
-      it('should use message from the error event when mobile v3', () => {  
+      it('uses message from the error event when mobile v3', () => {  
         const setTextSpy = jest.spyOn(errorMessageOverlay['errorLabel'], 'setText');
   
         errorMessageOverlay['tvNoiseBackground'] = { start: () => {} } as any;
--- a/spec/components/errormessageoverlay.spec.ts
+++ b/spec/components/errormessageoverlay.spec.ts
@@ -1,0 +1,69 @@
+import { UIInstanceManager } from './../../src/ts/uimanager';
+import { ErrorMessageOverlay } from '../../src/ts/components/errormessageoverlay';
+import { MobileV3PlayerEvent } from '../../src/ts/mobilev3playerapi';
+import { MockHelper, TestingPlayerAPI } from '../helper/MockHelper';
+
+describe('ErrorMessageOverlay', () => {
+  describe('configure', () => {
+    let errorMessageOverlay: ErrorMessageOverlay;
+    let playerMock: TestingPlayerAPI;
+    let uiInstanceManagerMock: UIInstanceManager;
+  
+    beforeEach(() => {
+      errorMessageOverlay = new ErrorMessageOverlay({});
+      playerMock = MockHelper.getPlayerMock();
+      uiInstanceManagerMock = MockHelper.getUiInstanceManagerMock();
+    });
+  
+    it('should add event listener for source loaded event', () => {
+      const onSpy = jest.spyOn(playerMock, 'on');
+      errorMessageOverlay.configure(playerMock, uiInstanceManagerMock);
+
+      expect(onSpy).toHaveBeenCalledWith(playerMock.exports.PlayerEvent.SourceLoaded, expect.any(Function));
+    });
+
+    it('should add event listener for error event when not mobile v3', () => {
+      const onSpy = jest.spyOn(playerMock, 'on');
+      errorMessageOverlay.configure(playerMock, uiInstanceManagerMock);
+
+      expect(onSpy).toHaveBeenCalledWith(playerMock.exports.PlayerEvent.Error, expect.any(Function));
+    });
+
+    describe('mobile v3 handling', () => {
+      let onSpy: jest.SpyInstance;
+
+      beforeEach(() => {
+        onSpy = jest.spyOn(playerMock, 'on');
+
+        (playerMock.exports.PlayerEvent as any).SourceError = MobileV3PlayerEvent.SourceError;
+        (playerMock.exports.PlayerEvent as any).PlayerError = MobileV3PlayerEvent.PlayerError;
+        (playerMock.exports.PlayerEvent as any).PlaylistTransition = MobileV3PlayerEvent.PlaylistTransition;
+      });
+
+      it('should add event listener for sourceerror and playererror when mobile v3', () => {
+        errorMessageOverlay.configure(playerMock, uiInstanceManagerMock);
+  
+        expect(onSpy).toHaveBeenCalledWith(MobileV3PlayerEvent.PlayerError, expect.any(Function));
+        expect(onSpy).toHaveBeenCalledWith(MobileV3PlayerEvent.SourceError, expect.any(Function));
+      });
+  
+      it('should use message from the error event when mobile v3', () => {  
+        const setTextSpy = jest.spyOn(errorMessageOverlay['errorLabel'], 'setText');
+  
+        errorMessageOverlay['tvNoiseBackground'] = { start: () => {} } as any;
+  
+        errorMessageOverlay.configure(playerMock, uiInstanceManagerMock);
+  
+        const playerErrorEvent = {
+          type: MobileV3PlayerEvent.PlayerError,
+          message: 'this is a player error',
+          name: 'playererror',
+        };
+        playerMock.eventEmitter.fireEvent<any>(playerErrorEvent);
+  
+        expect(onSpy).toHaveBeenCalledWith(MobileV3PlayerEvent.PlayerError, expect.any(Function));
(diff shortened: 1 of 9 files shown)


HalfbyteHeroes:
Ping @fabiandev @offnertho @stonko1994 @ravloony 😁

src/ts/components/errormessageoverlay.ts
```diff
@@ -95,38 +99,39 @@ export class ErrorMessageOverlay extends Container<ErrorMessageOverlayConfig> {
     }, this.config);
   }
 
-  configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
+  configure(player: PlayerAPI | MobileV3PlayerAPI, uimanager: UIInstanceManager): void {
```
offnertho, line 102: I'm not super into the UI, but I think we might consider a different approach here. 

Passing in 2 different APIs and handling the logic inside does not feel like a good approach to me. 

Isn't there some mechanic already in the UI where we can apply different facets or something similar @stonko1994 ? 

I think this will not be the only place where we have to change something in the UI because of the different mobile API.

offnertho (changes requested):
Like the fact that you added a lot of unit tests 💪 

I think that we might want to find a more general solution for the problem of different APIs for mobile and web. 

I think it might make sense to detect where we are on, and then instantiate specific components based on that, because I think this problem will come in different places as well.

src/ts/mobilev3playerapi.ts
```diff
@@ -0,0 +1,39 @@
...
+  let everyKeyExists = true;
+
+  for (const key in MobileV3PlayerEvent) {
+    if (MobileV3PlayerEvent.hasOwnProperty(key) && !player.exports.PlayerEvent.hasOwnProperty(key)) {
+      everyKeyExists = false;
+    }
+  }
+
+  return everyKeyExists;
+}
```
mukulmishra18, line 39: Just to highlight this one. We are checking for all `SourceError`, `PlayerError` and `PlaylistTransition` to be present in the `PlayerEvent` for mobile V3. Not sure if we need all of them or making sure that just one of the event present will be enough.

src/ts/components/errormessageoverlay.ts
stonko1994: > I think this will not be the only place where we have to change something in the UI because of the different mobile API.

That's true.

> where we can apply different facets

We have the different UI-variants, yes 🤔 But there is nothing on component level. 

> detect where we are on, and then instantiate specific components based on that

I like the idea however not sure if in the scope of this change now. (Maybe let's discuss this offline in more detail 🤔)

src/ts/components/errormessageoverlay.ts
```diff
@@ -2,8 +2,12 @@ import {ContainerConfig, Container} from './container';
 import {Label, LabelConfig} from './label';
 import {UIInstanceManager} from '../uimanager';
 import {TvNoiseCanvas} from './tvnoisecanvas';
-import {ErrorUtils} from '../errorutils';
+import {ErrorUtils } from '../errorutils';
```
stonko1994, line 5: Either before and after or no space 😄 
```suggestion
import { ErrorUtils } from '../errorutils';
```

src/ts/errorutils.ts
```diff
@@ -67,7 +68,11 @@ export namespace ErrorUtils {
     3100: 'An Advertising module error has occurred. Refer to the attached AdvertisingError.',
   };
 
-  export const defaultErrorMessageTranslator: ErrorMessageTranslator = (error: ErrorEvent) => {
+  export const defaultMobileV3ErrorMessageTranslator = (error: MobileV3PlayerErrorEvent | MobileV3SourceErrorEvent) => {
+    return `${error.message}\n(${error.name})`;
```
stonko1994, line 72: There is no property `name` on the v3 `ErrorEvent`:

IMG_0010 1

src/ts/mobilev3playerapi.ts
```diff
@@ -0,0 +1,39 @@
...
+import { WrappedPlayer } from './uimanager';
+
+export enum MobileV3PlayerEvent {
+  SourceError = 'sourceerror',
+  PlayerError = 'playererror',
+  PlaylistTransition = 'playlisttransition',
+}
+
+export interface MobileV3PlayerErrorEvent extends PlayerEventBase {
+  name: 'onPlayerError';
```
stonko1994, line 11: Why do you need the `name` property here?

src/ts/mobilev3playerapi.ts
```diff
@@ -0,0 +1,39 @@
...
+  code: number;
+  message: string;
+}
+
+export interface MobileV3SourceErrorEvent extends PlayerEventBase {
+  name: 'onSourceError';
```
stonko1994, line 17: Why do you need the `name` property here?

src/ts/mobilev3playerapi.ts
stonko1994: It depends if the web player will add those events as well eventually 🤔

src/ts/uimanager.ts
```diff
@@ -762,7 +762,7 @@ class InternalUIInstanceManager extends UIInstanceManager {
 /**
  * Extended interface of the {@link Player} for use in the UI.
  */
-interface WrappedPlayer extends PlayerAPI {
+export interface WrappedPlayer extends PlayerAPI {
```
stonko1994, line 765: Do we really want to export this? 🤔

spec/components/errormessageoverlay.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    let playerMock: TestingPlayerAPI;
+    let uiInstanceManagerMock: UIInstanceManager;
+  
+    beforeEach(() => {
+      errorMessageOverlay = new ErrorMessageOverlay({});
+      playerMock = MockHelper.getPlayerMock();
+      uiInstanceManagerMock = MockHelper.getUiInstanceManagerMock();
+    });
+  
+    it('should add event listener for source loaded event', () => {
```
stonko1994, line 18: I know we have differences in using the `should` word in test names 😄 (Up to you if you change that - No blocker still would prefer not using it) **all occurrences**

```suggestion
    it('adds an event listener for source loaded event', () => {
```

src/ts/mobilev3playerapi.ts
mukulmishra18: It was defined here https://github.com/bitmovin/player-api-specification/blob/master/ios/events.md
```java
public final class PlayerErrorEvent: PlayerEvent {
    public let name = "onPlayerError"
    public let code: PlayerError.Code
    public let message: String
    public let timestamp: TimeInterval
    public let data: DeficiencyData?
}
```

That's why we used in the `defaultMobileV3ErrorMessageTranslator`

src/ts/uimanager.ts
mukulmishra18: We are using the interface in the type guard as one of the player type, when typed in `uimanager`

src/ts/components/errormessageoverlay.ts
mukulmishra18: I think we can just remove it... we have a type guard anyway in this function that will later check for the `PlayerEvent` and typecast it to `MobileV3PlayerAPI`.

fabiandev (commented):
I think what @offnertho says makes sense, maybe we can find a way to better separate the handling of the different APIs and make it easier to extend and build on it in the future? 🙂

src/ts/components/errormessageoverlay.ts
```diff
@@ -142,4 +147,30 @@ export class ErrorMessageOverlay extends Container<ErrorMessageOverlayConfig> {
...
+        message = customMessage;
+      } else {
+        // The message is a translation function, so we call it
+        message = customMessage(event);
+      }
+    }
+  }
+
+  return message;
+}
```
ravloony, line 176: Maybe extract the "check if it's a function and do something, else do something else" logic out?

src/ts/mobilev3playerapi.ts
```diff
@@ -0,0 +1,39 @@
...
+export function isMobileV3PlayerAPI(player: WrappedPlayer | PlayerAPI | MobileV3PlayerAPI): player is MobileV3PlayerAPI {
+  let everyKeyExists = true;
+
+  for (const key in MobileV3PlayerEvent) {
+    if (MobileV3PlayerEvent.hasOwnProperty(key) && !player.exports.PlayerEvent.hasOwnProperty(key)) {
+      everyKeyExists = false;
+    }
+  }
+
+  return everyKeyExists;
```
ravloony, line 38: Adis says this is his fault :smile: 
```suggestion
  for (const key in MobileV3PlayerEvent) {
    if (MobileV3PlayerEvent.hasOwnProperty(key) && !player.exports.PlayerEvent.hasOwnProperty(key)) {
      return false;
    }
  }

  return true;
```

ravloony (changes requested):
Looks ok, had a couple of nitpicks.

src/ts/components/errormessageoverlay.ts
mukulmishra18: I refactored the function to just have one level of complexity, so we don't need to extract anything 😁

src/ts/mobilev3playerapi.ts
```diff
@@ -0,0 +1,39 @@
+import { PlayerAPI, PlayerEvent, PlayerEventBase, PlayerEventCallback } from 'bitmovin-player';
+import { WrappedPlayer } from './uimanager';
+
+export enum MobileV3PlayerEvent {
```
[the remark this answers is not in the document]
mukulmishra18: > Is it necessary that the UI knows about the specific error types?

Yes, some error type exists in Mobile V3 and some not. For example, `PlayerError` and `SourceError` will exists in Mobile V3 but `Error` will be removed.

> Would it be possible to listen to just any error event?

I mean sure, you can listen to any event but you have to add event handler for specific events for Mobile V3. If this answers your question.

> Will there only ever be SourceError and PlayerError, or does the UI now need to stay in sync w/ the categories (types, listeners, etc.)?

What exactly do you mean with this? For now we have `SourceError`, `PlayerError` and `PlaylistTransition` events in Mobile V3 that needs to be supported. If there is any other, we have to support it as well if required I guess.

mukulmishra18:
> I think what @offnertho says makes sense, maybe we can find a way to better separate the handling of the different APIs and make it easier to extend and build on it in the future? 

Sure, but I would like to have this in a separate PR. This shouldn't be the concern here.

I am not sure how much work it require to achieve this. I think we have to rewrite the whole component for V3, so basically a different one.

src/ts/components/errormessageoverlay.ts
offnertho: Well I don't think we can just remove it. Can u at least explore the option of having a `mobile-errormessageoverlay` and a `weberrormessageoverlay`? I'd be interested in how that could look like. If it is too much effort we still can say out of scope and plan it  properly.

offnertho:
@mukulmishra18 

I do think it is a concern here. It is about how we implement this feature right which we are in full control right now. If it takes longer to do it properly, we can talk about possible options, but we should always target the cleanest possible and most future proof solution. Especially considering more changes coming from mobileV3.

mukulmishra18:
>  If it takes longer to do it properly, we can talk about possible options

Well, I don't have much knowledge of the codebase but I looked into it and I think it is not as simple to do 😄 , that's why I was saying that this shouldn't be the concern here. But I will have a look again 👍

ravloony (approved):
LGTM :+1:

fabiandev (commented):
As I couldn't be part of the sync call, leaving this up to the other reviewers :)

offnertho (approved):

offnertho:
As decided in the call we had:

the changes are fine as they are right now, however, we need a `task force` to think about how we can support more diverging mobile and web apis in the future.

PR #368, opened 2021-03-05
PW-4775 touchscreen error IE/UWP fix

thomhs22:
Fixed the  'Touchscreen is undefined' error on IE/UWP when using mouse and touchscreen laptop


thomhs22:
closing to recreate correctly

PR #369, opened 2021-03-05
PW-4775 touchscreen error IE/UWP fix

thomhs22:
Fixed the 'Touchscreen is undefined' error on IE/UWP when using mouse and touchscreen laptop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Controls' focus highlighting is shown in case of non-keyboard interaction on some browsers/platforms
-- TouchEvent error in seekbar controls when using the mouse on a touchscreen Windows device and IE/UWP
+- `TouchEvent` error in seekbar controls when using the mouse on a touchscreen Windows device and IE/UWP
 
 ## [3.24.0] - 2021-02-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Controls' focus highlighting is shown in case of non-keyboard interaction on some browsers/platforms
+- TouchEvent error in seekbar controls when using the mouse on a touchscreen Windows device and IE/UWP
 
 ## [3.24.0] - 2021-02-16
 
--- a/src/ts/components/seekbar.ts
+++ b/src/ts/components/seekbar.ts
@@ -652,7 +652,7 @@ export class SeekBar extends Component<SeekBarConfig> {
     // and mouseup handlers to the whole document. A seek is triggered when the user lifts the mouse key.
     // A seek mouse gesture is thus basically a click with a long time frame between down and up events.
     seekBar.on('touchstart mousedown', (e: MouseEvent | TouchEvent) => {
-      let isTouchEvent = BrowserUtils.isTouchSupported && e instanceof TouchEvent;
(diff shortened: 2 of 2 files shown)


jmsn (commented):
Could you please elaborate what this `Touchscreen is undefined` error is caused by? I don't quite get it. @thomhs22

thomhs22:
> Could you please elaborate what this `Touchscreen is undefined` error is caused by? I don't quite get it. @thomhs22

hello @jmsn , sure, as specified in ticket: https://bitmovin.atlassian.net/browse/PW-4775, this issue is seen on IE/UWP apps when trying to scrub the seek bar/volume bar, using the mouse, only when it is a Windows Touchscreen Device.
The `TouchEvent` type is not supported by IE (https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/TouchEvent),  so the following instruction fails:

> let isTouchEvent = BrowserUtils.isTouchSupported && e instanceof TouchEvent;

because `TouchEvent `is undefined.
My fix is adding a validation by checking the `constructor name` before doing the `TypeAssertion`

stonko1994:
@thomhs22 please don't use internal issue-links in this repo as it's public and open source

Cagince (approved):

src/ts/components/seekbar.ts
```diff
@@ -652,7 +652,7 @@ export class SeekBar extends Component<SeekBarConfig> {
     // and mouseup handlers to the whole document. A seek is triggered when the user lifts the mouse key.
     // A seek mouse gesture is thus basically a click with a long time frame between down and up events.
     seekBar.on('touchstart mousedown', (e: MouseEvent | TouchEvent) => {
-      let isTouchEvent = BrowserUtils.isTouchSupported && e instanceof TouchEvent;
+      let isTouchEvent = BrowserUtils.isTouchSupported && e.constructor.name === 'Touchscreen' && e instanceof TouchEvent;
```
jmsn, line 655: I'm afraid, I think this check for the constructor is not very predictable/safe and there's an easier way.

I'd propose to check for `window.TouchEvent` instead. And I'd put that in a helper function that we'd then use in both places 🙂

```ts
function isTouchEvent(e: UIEvent): e is TouchEvent {
  return window.TouchEvent && e instanceof TouchEvent;
};
```

jmsn (changes requested):

src/ts/components/seekbar.ts
thomhs22: hi @jmsn unfortunately  the `window.TouchEvent` is also `undefined` on IE, I tried this as well.

jmsn: yes and that's exactly why checking `window.TouchEvent` before `e instanceof TouchEvent` will prevent us from running into the error 🙂

thomhs22: thanks for the suggestion @jmsn I just made the changes.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Controls' focus highlighting is shown in case of non-keyboard interaction on some browsers/platforms
+- TouchEvent error in seekbar controls when using the mouse on a touchscreen Windows device and IE/UWP
```
jmsn, line 11:
```suggestion
- `TouchEvent` error in seekbar controls when using the mouse on a touchscreen Windows device and IE/UWP
```

jmsn (commented):
Nice, one last nitpick and we're good I think

jmsn (approved):

PR #377, opened 2021-04-20
Ad Duration values: round to the nearest integer to compensate for du…

SInCE:
Example: Ad duration is 30.04 seconds, 30 seconds to be expected to be shown in the UI.

Action taken: round ad duration to the closest integer value.


hanneslinder:
Change looks good, but can you please add a `changelog` as well?

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
```
hanneslinder, line 7:
```suggestion
## [develop]
```

We usually use `develop` for unreleased changes :smiley:

src/ts/stringutils.ts
```diff
@@ -106,12 +106,12 @@ export namespace StringUtils {
...
+          time = Math.round(player.getDuration() - player.getCurrentTime());
         }
       } else if (formatString.indexOf('playedTime') > -1) {
-        time = player.getCurrentTime();
+        time = Math.round(player.getCurrentTime());
       } else if (formatString.indexOf('adDuration') > -1) {
-        time = player.getDuration();
+        time = Math.round(player.getDuration());
       }
       return formatNumber(time, formatString);
```
ravloony, line 116: Nit: You could do a single `Math.round` call here instead and save a few bytes.
```suggestion
      return formatNumber(Math.round(time), formatString);
```

ravloony (changes requested):
Just a nit, otherwise looks fine.

src/ts/stringutils.ts
SInCE: 👍

hanneslinder (approved):
LGTM :+1:

Cagince (approved):
LGTM as well

ravloony (approved):
LGTM :+1:

PR #378, opened 2021-05-03
[PW-5093] Add handling for `cueUpdate` events

hanneslinder:
_Waiting until the player PR is finished before we merge this one_ 

This PR adds support for `cueUpdate` events

Changes to the UI: 

- Whenever `cueUpdate` event is triggered, we check if there is a previous cue in the `subtitleManager` that matches the id. 
- If a match is found we create a new subtitle label and replace the current one. 
- Matching by id works because `cueUpdate` events contain the same `start` and `text` properties as the `cueEnter` event, only the `html` content is different.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [develop]
 
 ### Added
-- Support for subtitle `CueUpdate` events
+- Support for `CueUpdate` events that where introduced in player v8.60.0
 
 ## [3.26.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Added
+- Support for subtitle `cueUpdate` events
+
 ## [3.26.0]
 
 ### Fixed
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -69,11 +69,10 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
 
     player.on(player.exports.PlayerEvent.CueUpdate, (event: SubtitleCueEvent) => {
       const label = this.generateLabel(event);
-      const previousLabel = subtitleManager.getPreviousCue(event);
+      const labelToReplace = subtitleManager.cueUpdate(event, label);
 
-      if (previousLabel) {
-        subtitleManager.cueUpdate(event, label);
-        this.subtitleContainerManager.replaceLabel(previousLabel, label);
+      if (labelToReplace) {
+        this.subtitleContainerManager.replaceLabel(labelToReplace, label);
       }
     });
 
@@ -371,12 +370,15 @@ class ActiveSubtitleManager {
     this.addCueToMap(event, label);
   }
 
-  cueUpdate(event: SubtitleCueEvent, label: SubtitleLabel): void {
-    this.addCueToMap(event, label);
-  }
+  cueUpdate(event: SubtitleCueEvent, label: SubtitleLabel): SubtitleLabel | undefined {
+    const labelToReplace = this.popCueFromMap(event);
 
-  getPreviousCue(event: SubtitleCueEvent): SubtitleLabel | undefined {
-    return this.popCueFromMap(event);
+    if (labelToReplace) {
+      this.addCueToMap(event, label);
+      return labelToReplace;
+    }
+
+    return undefined;
   }
 
   private addCueToMap(event: SubtitleCueEvent, label: SubtitleLabel): void {
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -376,7 +376,7 @@ class ActiveSubtitleManager {
   }
 
   getPreviousCue(event: SubtitleCueEvent): SubtitleLabel | undefined {
-    return this.removeCueFromMap(event);
+    return this.popCueFromMap(event);
   }
 
   private addCueToMap(event: SubtitleCueEvent, label: SubtitleLabel): void {
@@ -390,7 +390,7 @@ class ActiveSubtitleManager {
     this.activeSubtitleCueCount++;
   }
 
-  private removeCueFromMap(event: SubtitleCueEvent): SubtitleLabel | undefined {
+  private popCueFromMap(event: SubtitleCueEvent): SubtitleLabel | undefined {
     let id = ActiveSubtitleManager.calculateId(event);
     let activeSubtitleCues = this.activeSubtitleCueMap[id];
 
@@ -442,7 +442,7 @@ class ActiveSubtitleManager {
    * @return {SubtitleLabel|null}
    */
   cueExit(event: SubtitleCueEvent): SubtitleLabel {
-    return this.removeCueFromMap(event);
+    return this.popCueFromMap(event);
   }
 
   /**
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -373,7 +373,7 @@ class ActiveSubtitleManager {
     this.addCueToMap(event, label);
   }
 
-  getPreviousCue(event: SubtitleCueEvent): SubtitleLabel {
+  getPreviousCue(event: SubtitleCueEvent): SubtitleLabel | undefined {
     return this.removeCueFromMap(event);
   }
 
@@ -388,7 +388,7 @@ class ActiveSubtitleManager {
     this.activeSubtitleCueCount++;
   }
 
-  private removeCueFromMap(event: SubtitleCueEvent): SubtitleLabel {
+  private removeCueFromMap(event: SubtitleCueEvent): SubtitleLabel | undefined {
     let id = ActiveSubtitleManager.calculateId(event);
     let activeSubtitleCues = this.activeSubtitleCueMap[id];
 
@@ -405,12 +405,10 @@ class ActiveSubtitleManager {
       this.activeSubtitleCueCount--;
 
       return activeSubtitleCue.label;
-    } else {
-      return null;
     }
   }
 
-  static generateImageTagText(imageData: string): string {
+  static generateImageTagText(imageData: string): string | undefined {
     if (!imageData) {
       return;
     }
@@ -427,13 +425,11 @@ class ActiveSubtitleManager {
    * @param event
    * @return {SubtitleLabel}
    */
-  getCues(event: SubtitleCueEvent): SubtitleLabel[] {
+  getCues(event: SubtitleCueEvent): SubtitleLabel[] | undefined {
     let id = ActiveSubtitleManager.calculateId(event);
     let activeSubtitleCues = this.activeSubtitleCueMap[id];
     if (activeSubtitleCues && activeSubtitleCues.length > 0) {
       return activeSubtitleCues.map((cue) => cue.label);
-    } else {
-      return null;
     }
   }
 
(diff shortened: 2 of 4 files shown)


src/ts/components/subtitleoverlay.ts
```diff
@@ -256,18 +280,20 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
       this.updateComponents();
       this.show();
     }
   }
 
   removePreviewSubtitleLabel(): void {
-    this.previewSubtitleActive = false;
-    this.subtitleContainerManager.removeLabel(this.previewSubtitle);
-    this.updateComponents();
+    if (this.previewSubtitleActive) {
```
NikolaNikushev, line 292: nit: That way you are not the owner of all lines, but only the updated IF clause
```suggestion
    if (!this.previewSubtitleActive) {
        return;
    }
```

src/ts/components/subtitleoverlay.ts
```diff
@@ -339,34 +365,52 @@ class ActiveSubtitleManager {
...
     // Create array for id if it does not exist
     this.activeSubtitleCueMap[id] = this.activeSubtitleCueMap[id] || [];
 
     // Add cue
     this.activeSubtitleCueMap[id].push({ event, label });
     this.activeSubtitleCueCount++;
+  }
 
-    return label;
+  private removeCueFromMap(event: SubtitleCueEvent): SubtitleLabel {
```
NikolaNikushev, line 391: If we apply the FIFO approach, we could then name it `popLastCueFromMap`

As remove suggest we are fully removing and it should not exist anymore.

NikolaNikushev, line 391: As per code talk, we should use `undefined` instead of nulls
+ Missing return type for undefined
```suggestion
  private removeCueFromMap(event: SubtitleCueEvent): SubtitleLabel | undefined {
```

src/ts/components/subtitleoverlay.ts
```diff
@@ -339,34 +365,52 @@ class ActiveSubtitleManager {
...
+       * works fine. Theoretically it can happen that two cues with colliding ids are removed at different times, in
+       * the wrong order. This rare case has yet to be observed. If it ever gets an issue, we can take the unstable
+       * cue end time (which can change between CueEnter and CueExit IN CueUpdate) and use it as an
+       * additional hint to try and remove the correct one of the colliding cues.
+       */
+      let activeSubtitleCue = activeSubtitleCues.shift();
+      this.activeSubtitleCueCount--;
+
+      return activeSubtitleCue.label;
+    } else {
```
NikolaNikushev, line 408: Nit: No need for `else`, as the previous case has a return

src/ts/components/subtitleoverlay.ts
```diff
@@ -339,34 +365,52 @@ class ActiveSubtitleManager {
...
+      return null;
```
NikolaNikushev, line 409: As per code talk we should prefer undefined over nulls
```suggestion
      return undefined;
```

src/ts/components/subtitleoverlay.ts
```diff
@@ -457,22 +483,27 @@ export class SubtitleRegionContainerManager {
...
+      return {
+        regionContainerId: label.vtt.region && label.vtt.region.id ? label.vtt.region.id : 'vtt',
+        regionName: 'vtt',
+      };
+    } else {
+      return {
+        regionContainerId: label.region || 'default',
+        regionName: label.region || 'default',
+      };
+    }
```
NikolaNikushev, line 497: No need for ELSE as we have a return in the previous statement
```suggestion
    }
      return {
        regionContainerId: label.region || 'default',
        regionName: label.region || 'default',
      };
   
```

src/ts/components/subtitleoverlay.ts
```diff
@@ -507,6 +538,13 @@ export class SubtitleRegionContainerManager {
     this.subtitleRegionContainers[regionContainerId].addLabel(label, overlaySize);
   }
 
+  updateLabel(previousLabel: SubtitleLabel, newLabel: SubtitleLabel): void {
+    const { regionContainerId } = this.getRegion(previousLabel);
+
+    this.subtitleRegionContainers[regionContainerId].removeLabel(previousLabel);
```
NikolaNikushev, line 544: Question: here we implement a different logic than the FIFO approach above, is this correct?

NikolaNikushev (changes requested):

src/ts/components/subtitleoverlay.ts
```diff
@@ -54,27 +54,29 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
       this.updateComponents();
     });
 
+    player.on(player.exports.PlayerEvent.CueUpdate, (event: SubtitleCueEvent) => {
+      const label = this.generateLabel(event);
+      const previousLabel = subtitleManager.getPreviousCue(event);
+
+      if (previousLabel) {
+        subtitleManager.cueUpdate(event, label);
+        this.subtitleContainerManager.updateLabel(previousLabel, label);
```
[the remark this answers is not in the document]
hanneslinder: Sorry I am not following, mind to explain what you mean?
What's happening here is that in order to update a cue we must first find the previous one. If one is found we remove it from the DOM, and insert a new one.

src/ts/components/subtitleoverlay.ts
hanneslinder: Thanks for spotting, fixed in https://github.com/bitmovin/bitmovin-player-ui/pull/378/commits/6b9d35304a117fb79d93fe0fad303c7e013b989d

src/ts/components/subtitleoverlay.ts
hanneslinder: I agree, fixed in https://github.com/bitmovin/bitmovin-player-ui/pull/378/commits/6b9d35304a117fb79d93fe0fad303c7e013b989d

src/ts/components/subtitleoverlay.ts
hanneslinder: Fixed in 6b9d353

src/ts/components/subtitleoverlay.ts
hanneslinder: Fixed in 6b9d353

src/ts/components/subtitleoverlay.ts
```diff
@@ -339,34 +365,52 @@ class ActiveSubtitleManager {
...
-      text: event.html || ActiveSubtitleManager.generateImageTagText(event.image) || event.text,
-      vtt: event.vtt,
-      region: event.region,
-      regionStyle: event.regionStyle,
-    });
+  cueUpdate(event: SubtitleCueEvent, label: SubtitleLabel): void {
+    this.addCueToMap(event, label);
+  }
+
+  getPreviousCue(event: SubtitleCueEvent): SubtitleLabel {
```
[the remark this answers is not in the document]
hanneslinder: As discussed offline I would rather keep it like that as removal of cues from the internal map is of no concern to the outside world. Having to manage the cue map from outside would only increase complexity and invite bugs without any real gain.

src/ts/components/subtitleoverlay.ts
```diff
@@ -54,27 +54,29 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
-      this.subtitleContainerManager.addLabel(labelToAdd, this.getDomElement().size());
+      this.subtitleContainerManager.addLabel(label, this.getDomElement().size());
       this.updateComponents();
     });
 
+    player.on(player.exports.PlayerEvent.CueUpdate, (event: SubtitleCueEvent) => {
+      const label = this.generateLabel(event);
+      const previousLabel = subtitleManager.getPreviousCue(event);
+
+      if (previousLabel) {
```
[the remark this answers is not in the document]
hanneslinder: I disagree, this should be kept as implementation detail inside the class.

src/ts/components/subtitleoverlay.ts
hanneslinder: Changed in https://github.com/bitmovin/bitmovin-player-ui/pull/378/commits/144549952d0db2a91607a93623db00df517d7dcb

src/ts/components/subtitleoverlay.ts
hanneslinder: Sure thing, I've renamed the method in https://github.com/bitmovin/bitmovin-player-ui/pull/378/commits/79883890938961fab7227b2b191bc2c8e77dcc38

src/ts/components/subtitleoverlay.ts
hanneslinder: Unless I am missing something the FIFO approach should still be valid here.

NikolaNikushev (approved):
LGTM

felix-hoc (commented):
Nice and clean changes 💪🏼 
I'm just not 100% certain on the flow to remove the previous cue.

src/ts/components/subtitleoverlay.ts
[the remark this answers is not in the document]
hanneslinder: Yeah I agree that it's a bit hard to follow and I am open to suggestions on how to simplify this one :smiley: I didn't want to refactor the whole `ActiveSubtitleManager` so I tried to make it work with what was already there. If required I can still rework this part of the code tho, just let me know your thoughts.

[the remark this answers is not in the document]
hanneslinder: Thanks @felix-hoc for your suggestion. I've applied it now and it indeed looks a bit cleaner that way :smiley:

felix-hoc (approved):
LGTM 👍🏼 Can confirm it works with the latest player build from the linked PR :)

stonko1994 (changes requested):
- Please try to not put any internal links in the PR description as this repo is open source
- CEA captions seem to be broken.

You can test it using one of those buttons:
Screenshot 2021-05-04 at 08 55 27


# Expected behaviour (dev):
_I have a black background on the subtitles configured via the UI to make it visible_
## CEA-608 subtitle test pattern (60 secs)
Screenshot 2021-05-04 at 08 57 43

## CEA-608 subtitle test pattern 2
Screenshot 2021-05-04 at 08 57 59

# Current behaviour on this branch
## CEA-608 subtitle test pattern (60 secs)
- No output at all
## CEA-608 subtitle test pattern 2
- No output at all

src/ts/components/subtitleoverlay.ts
```diff
@@ -339,34 +366,53 @@ class ActiveSubtitleManager {
...
+       */
+      let activeSubtitleCue = activeSubtitleCues.shift();
+      this.activeSubtitleCueCount--;
+
+      return activeSubtitleCue.label;
+    }
   }
 
-  private static generateImageTagText(imageData: string): string {
+  static generateImageTagText(imageData: string): string | undefined {
```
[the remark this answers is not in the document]
hanneslinder: Never seen optional chaining used for a return type, is this valid?

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Added
+- Support for subtitle `CueUpdate` events
```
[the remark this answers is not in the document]
hanneslinder: Updated the changelog to include the player version here https://github.com/bitmovin/bitmovin-player-ui/pull/378/commits/2b239ffe5253e59012ca5814329551e29d1171be

stonko1994 (commented):
The CEA rendering looks wrong 🤔 

Screenshot 2021-05-04 at 10 41 43
Screenshot 2021-05-04 at 10 41 07

src/ts/components/subtitleoverlay.ts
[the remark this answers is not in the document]
stonko1994: https://github.com/microsoft/TypeScript/issues/13184

stonko1994 (approved):
Manual testing looks good now 👍

jmsn:
I guess this should have gone to `develop` instead of `master` 😄

PR #379, opened 2021-05-06
Merge `master` to `develop`

jmsn:
https://github.com/bitmovin/bitmovin-player-ui/pull/378 was merged into `master` but we need it in `develop` as the release branch is created from there.

Side note: We should really enable some branch protection for `master`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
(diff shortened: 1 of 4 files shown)


hanneslinder (approved):
Whoops, that one is on me :neutral_face:

jmsn:
Discussed that we'll instead revert in `master` and create a new PR going into `develop` for https://github.com/bitmovin/bitmovin-player-ui/pull/378

PR #387, opened 2021-05-19
Fix a  glitch with the seekbar

iktumi:
On some media if you drag and drop the marker on the seekbar while the content is not loaded, the marker will jump back to previous know position insetead of setting itself imeediatily on the current seeked position.


offnertho:
Sorry for the PR closing @iktumi. 

We are currently renaming the master branch to main. Could you pls create another PR, if still relevant, based on main?

PR #388, opened 2021-06-14
[PW-4867] Fix VTT position end alignment

mukulmishra18:
When the align property of the VTT cue is set to `end` or `right`, we set the CSS `right` property to the VTT position value. This is wrong because this moves the cue to a totally different position.

I think the whole understanding of the VTT cue positioning is wrong. We think that the cues should be positioned to the VTT cue `position` offset from the `start`, `end`, `left`, or `right` side of the video container. But this is not true, the cue should be positioned to the offset in the writing direction of the VTT.

When position align is set to `end` or `right` the cue text should be present on the right side of the cue box.
When the position align is set to `start` or `left`, the cue text should be present on the left side of the cue box.

This PR fixes the cue position when the position align is set to `end` of `right`.

This is how the cue should look like when the position is 80% and based on position align, it should be aligned at the left or right side of the cue box.

Untitled Diagram (1)

### DIFF
--- a/src/ts/vttutils.ts
+++ b/src/ts/vttutils.ts
@@ -128,7 +128,7 @@ const setVttPositionAlign = (cueContainerDom: DOM, vtt: VTTProperties, direction
         break;
       case 'line-right':
         cueContainerDom.css(direction, 'auto');
-        cueContainerDom.css(DirectionPair.get(direction), `${vtt.position}%`);
+        cueContainerDom.css(DirectionPair.get(direction), `${100 - vtt.position}%`);
         cueContainerDom.css('justify-content', 'flex-end');
         break;
       default:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/) 
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- VTT cue positioning when position alignment is `end` or `right`
+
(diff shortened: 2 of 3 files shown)


stonko1994 (approved):

src/ts/vttutils.ts
```diff
@@ -128,7 +128,7 @@ const setVttPositionAlign = (cueContainerDom: DOM, vtt: VTTProperties, direction
         break;
       case 'line-right':
         cueContainerDom.css(direction, 'auto');
-        cueContainerDom.css(DirectionPair.get(direction), `${vtt.position}%`);
+        cueContainerDom.css(DirectionPair.get(direction), `${100 - vtt.position}%`);
```
bitmovin-kenny, line 131: This basically turns the `line-right` into `line-left` as far as i can tell.
If I use `position:0%,line-left` the content will be aligned on the left border (0% from flex-start)
if I use `position:0%,line-right` the content will also be aligned on the left border since it will get converted to 100% from flex-end

this can't be right, can it?

bitmovin-kenny (changes requested):

src/ts/vttutils.ts
mukulmishra18: Yes, this is correct and the content should at the left border when you use **position: 0%** because it is **0% in the writing direction**, that is from the left side of the container.

The `line-left` and `line-right` does **not define the offset of the text/cue box**, rather in which direction the box will grow or where to put the text so that it will match the offset provided by position. However you want to understand.

TeslaAdis (approved):

bitmovin-kenny (approved):

PR #392, opened 2021-06-25
Replace npm `prepublish` with `prepare`

felix-hoc:
The `prepublish` script got deprecated, and does not get executed anymore when running `npm publish` on newer npm versions.
Instead, we can use `prepare`, which has the same behavior (runs on `install`, `ci`, `publish`, and more)

For reference:
- https://github.com/npm/npm/issues/10074
- https://docs.npmjs.com/cli/v7/using-npm/scripts

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "types": "./dist/js/framework/main.d.ts",
   "scripts": {
     "test": "npx jest",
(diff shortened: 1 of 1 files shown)


fabiandev (approved):

bitmovin-kenny (approved):
Thanks for linking the docs 👍

PR #398, opened 2021-07-27
[PW-5266] Update UI for new seeking behavior

wasp898:
As you know, we have changed a bit the behavior of the `getCurrentTime` API. So now we return the seek target immediately, and we won't wait for the seeking to be complete. This created a small inconsistency in the UI since the playback time label still only gets refreshed after the Seeked event. Also, a glitch in the buffer level can be observed during the transition from Seek to Seeked: suddenly the buffer is huge and then it gets reset 👀 

**Changes:**
- update the playback time label on Seek event
- do not update the buffer level during seekings since it currently causes a glitch due to the sudden changes that happen to the buffer level during the transition

### DIFF
--- a/src/ts/components/playbacktimelabel.ts
+++ b/src/ts/components/playbacktimelabel.ts
@@ -133,7 +133,7 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
     };
 
     player.on(player.exports.PlayerEvent.TimeChanged, playbackTimeHandler);
-    player.on(player.exports.PlayerEvent.Seeked, playbackTimeHandler);
+    player.on(player.exports.PlayerEvent.Seek, playbackTimeHandler);
 
     player.on(player.exports.PlayerEvent.TimeShift, updateLiveTimeshiftState);
     player.on(player.exports.PlayerEvent.TimeShifted, updateLiveTimeshiftState);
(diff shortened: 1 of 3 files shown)


hanneslinder (approved):
Looks good :+1: 
A changelog entry might be good tho

wasp898:
> Looks good 👍
> A changelog entry might be good tho

Oh right, my bad 😅 
I also changed the base branch as it would have merged to master directly.

hanneslinder:
> I also changed the base branch as it would have merged to master directly.

I totally missed that :see_no_evil: Good you caught it because in one of the last releases I merged directly to `master` and reverting that caused lots of pain

felix-hoc (approved):
LGTM 👍🏼

src/ts/components/playbacktimelabel.ts
```diff
@@ -133,7 +133,7 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
     };
 
     player.on(player.exports.PlayerEvent.TimeChanged, playbackTimeHandler);
-    player.on(player.exports.PlayerEvent.Seeked, playbackTimeHandler);
+    player.on(player.exports.PlayerEvent.Seek, playbackTimeHandler);
```
TeslaAdis, line 136: Would it hurt to have both? Due to possible `time` adjustment in the player. Not sure how often `timechanged` event would not come right after `seeked` 🤷‍♂️

TeslaAdis (approved):

src/ts/components/playbacktimelabel.ts
wasp898: Makes sense. 👍

PR #402, opened 2021-08-04
Fix empty lines in subtitles

bitmovin-kenny:
We accidentally added empty lines to Subtitles. From the following subtitle
```
00:00:00.000 --> 00:00:05.000
We are in for a ride.
```
we would get the following overlay:
image

this extra empty line got added because of our support for multiple subtitles cues by the following CSS rule:
```css
.bmpui-ui-subtitle-overlay .bmpui-ui-subtitle-label:nth-child(1n-1):after {
    content: "\A";
    white-space: pre-line;
}
```
This pseudo element should put each subtitle cue into a separate line, but when WebVTT styling is applied, the height of the pseudo element became one line instead of 0, due to `display: inline-flex`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,6 +140,8 @@ local.properties
 # Code Recommenders
 .recommenders/
 
+### VsCode ###
+.vscode
 
 ### Windows ###
 # Windows thumbnail cache files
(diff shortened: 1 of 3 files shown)


TeslaAdis (approved):
🤷‍♂️

NikolaNikushev (approved):

stonko1994 (approved):

PR #409, opened 2021-09-29
[PW-5720] Reset styles on subtitle overlay content

mimcz:
It may happen that TTML elements rendered inside of `bmpui-ui-subtitle-overlay` conflict with other style definitions on the target application resulting in broken or unexpected subtitle styling. This aims to perform some basic reset on elements nested inside of `bmpui-ui-subtitle-overlay` to try to prevent this.

### DIFF
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -55,3 +55,10 @@
     transition: bottom $animation-duration-short ease-in;
   }
 }
+
+.#{$prefix}-ui-uicontainer .#{$prefix}-ui-subtitle-overlay * {
+  // This aims to prevent possibly conflicting style definitions inherited
+  // from target applications whuch can break subtitle styling. It's still possible
+  // to override this with selector of higher priority score.
+  all: unset;
+}
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -3,6 +3,7 @@
 
 .#{$prefix}-ui-uicontainer .#{$prefix}-ui-subtitle-overlay {
   @extend %ui-container;
+
   @include hidden;
 
   bottom: 0;
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -24,7 +24,7 @@
   }
 
   p {
-    // It may happen that we sender <p> inside of an <span> and the `all: unset;` reseting above sets
+    // It may happen that we render <p> inside of an <span> and the `all: unset;` reseting above sets
     // p to inherit the inline display instead of its default display block so this sets it back.
     display: block;
   }
--- a/src/scss/skin-modern/components/_subtitleoverlay.scss
+++ b/src/scss/skin-modern/components/_subtitleoverlay.scss
@@ -1,64 +1,71 @@
 @import '../variables';
 @import '../mixins';
 
-.#{$prefix}-ui-subtitle-overlay {
-  @extend %ui-container;
+.#{$prefix}-ui-uicontainer {
+  .#{$prefix}-ui-subtitle-overlay {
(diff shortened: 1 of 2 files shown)


NikolaNikushev (approved):

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -1,7 +1,7 @@
 @import '../variables';
 @import '../mixins';
 
-.#{$prefix}-ui-subtitle-overlay {
+.#{$prefix}-ui-uicontainer .#{$prefix}-ui-subtitle-overlay {
```
stonko1994, line 4: Why do we need to scope this to the `-ui-uicontainer` now? 🤔 Where else should the `-ui-subtitle-overlay` be? 🤔

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -16,6 +16,19 @@
   top: 0;
   transition: bottom $animation-duration-short ease-out;
 
+  * {
+    // This aims to prevent possibly conflicting style definitions inherited
+    // from target applications which can break subtitles styling. It's still possible
+    // to override this with selector of higher priority score.
+    all: unset;
```
stonko1994, line 23: AFAIK IE doesn't support this? 🤔 Do we care about IE or is there a different possibility for IE? 🤔

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -16,6 +16,19 @@
   top: 0;
   transition: bottom $animation-duration-short ease-out;
 
+  * {
```
stonko1994, line 19: Just to double-check (haven't tested it): This doesn't interfere with our subtitle styling which can be done by a user through the `SettingsPanel`?

stonko1994 (commented):
I do get why this is needed. However, would it make sense to maybe apply this on the root container of the UI and not just the subtitles? 🤔 E.g. other elements of the UI could also be affected?

src/scss/skin-modern/components/_subtitleoverlay.scss
mimcz: Yes settings in styling panel works. Just the inline styles that are defined from TTML has higher priority, but I guess that is expected as its inline in closest level and it did behave like that even before this change.

src/scss/skin-modern/components/_subtitleoverlay.scss
mimcz: Yep true, we dediced not to care for IE in this case. It would require quite a lot of lines just to cover the reset. And its also just a safety check.

src/scss/skin-modern/components/_subtitleoverlay.scss
mimcz: It adds more prio score for these style definitions if there is `*` selector that would have higher priority.

mimcz:
> I do get why this is needed. However, would it make sense to maybe apply this on the root container of the UI and not just the subtitles? 🤔 E.g. other elements of the UI could also be affected?

@stonko1994 I initially wanted to apply it to the root container. At first sight it looked ok, but it broke the interaction as it reseted also some pointer events so I scoped it only to the subtitles overlay.

src/scss/skin-modern/components/_subtitleoverlay.scss
stonko1994: Hmm I see 🤔 But why would we only need it on the `subtitle-overlay` and not on all other components as well? 🤔

src/scss/skin-modern/components/_subtitleoverlay.scss
stonko1994: Ok please add a comment stating that this doesn't work on IE to the existing comment.

src/scss/skin-modern/components/_subtitleoverlay.scss
stonko1994: Gotch 👍

PR #412, opened 2021-11-03
Add seek preview as a configurable option

huysbs:
We are using Google DAI for server side ads insertion and we are implementing an ad snapback: if the user scrubs past an ad breakpoint, we force them to watch the ad by issuing a seek command to the beginning of the midroll.

The seek preview feature conflicts with the snapback logic because as it does the preview it seems to issue a seek event which makes our snapback thinks that the user has seeked past the breakpoint while they are still holding the mouse button and dragging around.

Disabling the seek preview fixes this bug.

The intention is **not** to hide the seek thumbnail but the actual video preview. As the user drags the seek button over the timeline, the video content is being updated which triggers other player events (seek for example).


src/ts/uiconfig.ts
```diff
@@ -85,4 +85,9 @@ export interface UIConfig {
    * For an example have a look at {@link ErrorMessageOverlayConfig.messages}
    */
   errorMessages?: ErrorMessageMap | ErrorMessageTranslator;
+  /**
+   * Toggles the seek preview feature.
+   * Default: true
+   */
+  enableSeekPreview?: boolean;
```
stonko1994, line 92: Any specific reason to but the option on the `UIConfig` over the `SeekBarConfig` 🤔

stonko1994 (changes requested):
Sorry for the late review.

- Can you please add a PR description to describe the problem and the change in this PR? Also, screenshots of how the UI looks when the option is disabled would be helpful.
- Is the intention only hiding the thumbnails or the whole `SeekBarLabel` (including the time) as well?
- I run the playback locally and changing the option doesn't seem to have any effect as it is right now 🤔 
- Also please resolve the merge conflicts.

huysbs:
> * Can you please add a PR description to describe the problem and the change in this PR? Also, screenshots of how the UI looks when the option is disabled would be helpful.
> * Is the intention only hiding the thumbnails or the whole `SeekBarLabel` (including the time) as well?
> * I run the playback locally and changing the option doesn't seem to have any effect as it is right now 🤔
> * Also please resolve the merge conflicts.

I've updated the description. A screenshot would not be useful in this case. When you disable the seekPreview what happens is that the actual video will not update as you drag the seek button around. The Seek Preview is not the thumbnail above the mouse but it's preview the actual video stream.

stonko1994:
Thanks for adding the description 👍 The target of this PR is clear to me now.

I'm still undecided if the new config option should go to the `UIConfig` or to the `SeekBarConfig`. The reasoning is that the `Seekbar` actually is reused e.g. for the `VolumeSlider`. There we also have an `onSeekPreview` with an `update...WhileScrubbing` which is not configurable by this property now.
One could argue that this is unrelated but not the best design but still having a config option on every `SeekBar` directly if the value should be updated while scrubbing or after releasing would IMO the cleaner approach.

For convenience, we could keep the option on the `UIConfig`. See `seekbarSnappingRange` as an example:
https://github.com/bitmovin/bitmovin-player-ui/blob/3d25d3304acaac0b4d14bb4d97d204ef40793929/src/ts/components/seekbar.ts#L420-L422

PS: Please add a changelog entry and resolve the merge conflict.

quochuy:
@stonko1994 not sure if I needed to bump the version number in package.json or if its automatically done by your publishing process but I did it. If it needs to be reverted, let me know.

stonko1994:
> @stonko1994 not sure if I needed to bump the version number in package.json or if its automatically done by your publishing process but I did it. If it needs to be reverted, let me know.

Yes please revert the version bump. That will happen in our release process. Apart from that, looks good to me 👍

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.35.0] - 2022-02-17
+
+### Added
+- seek preview is now configurable via `isSeekPreviewEnabled` from SeekBarConfig or UIConfig
```
felix-hoc, line 10: Slight nitpick to aviod confusion about the config options. We should mention the respective property names
```suggestion
- Seek preview is now configurable via `UIConfig.enableSeekPreview` or `SeekBarConfig.isSeekPreviewEnabled`
```

felix-hoc (changes requested):
Thanks a lot for your contribution! The code changes look good to me, just one suggestion for the changelog entry.

Also, some organizational stuff: Please change the base branch to `develop` and make sure the changelog entry is in the correct section. It should be in `## [develop]`, without a date. This will be updated automatically on release. As an example, see this PR: https://github.com/bitmovin/bitmovin-player-ui/pull/425/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR6

quochuy:
@felix-hoc I've did the suggested changes. Also renamed SeekBarConfig.isSeekPreviewEnabled to SeekBarConfig.enableSeekPreview it makes more sense since it's a config name not a state (which it initially was).

felix-hoc:
Great, thanks 🙌🏼 this looks g2g from my side.
Anything from you @stonko1994?

felix-hoc (approved):

stonko1994 (approved):
Unit tests would be nice if easily possible but not a hard requirement.

PR #413, opened 2021-12-14
Fix jumping behaviour of the seekbar scrubber while still seeking

NikolaNikushev:
As already identified by @iktumi  on https://github.com/bitmovin/bitmovin-player-ui/pull/387, the scrubber will bounce forward and backward while seeking. 

Why:
This happens due to the smooth transition of the scrubber connected to the current time of the player.
Every time the scrubber is moved, the relevant seek information is updated with the current time, which then moves the scrubber. 
Also when the player fires the appropriate events like 'Seeked' the position will be updated, which will overwrite the newly dragged position. 

Fix:
Do not update the scrubber position if the user is currently seeking, but still allow other operations such as dragging and position updates to update the scrubber.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [develop]
 
 ### Fixed
-- Dragging the scrubber in the seekbar jumping between locations during a seek operation
+- The scrubber could jump to an old position during a seek operation when it was dragged.
 
 ## [3.31.0] - 2021-10-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- Dragging the scrubber in the seekbar jumping between locations during a seek operation
+
 ## [3.31.0] - 2021-10-12
 
 ### Added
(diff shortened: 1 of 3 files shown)


src/ts/components/seekbar.ts
```diff
@@ -249,7 +251,9 @@ export class SeekBar extends Component<SeekBarConfig> {
         // paused nor playing. Playback updates are handled in the Timeout below.
         if (this.config.smoothPlaybackPositionUpdateIntervalMs === SeekBar.SMOOTH_PLAYBACK_POSITION_UPDATE_DISABLED
           || forceUpdate || player.isPaused() || (player.isPaused() === player.isPlaying())) {
-          this.setPlaybackPosition(playbackPositionPercentage);
+          if (!this.isSeeking()) {
```
bitmovin-kenny, line 254: this can be combined with the if above

src/ts/components/seekbar.ts
```diff
@@ -220,7 +220,9 @@ export class SeekBar extends Component<SeekBarConfig> {
           this.setPlaybackPosition(100);
         } else {
           let playbackPositionPercentage = 100 - (100 / player.getMaxTimeShift() * player.getTimeShift());
-          this.setPlaybackPosition(playbackPositionPercentage);
+          if (!this.isSeeking()) {
```
bitmovin-kenny, line 223: i guess the calculation of the `playbackPositionPercentage` can also be moved into this if, since it's unused outside?

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [develop]
+
+### Fixed
+- Dragging the scrubber in the seekbar jumping between locations during a seek operation
```
bitmovin-kenny, line 10:
```suggestion
- The scrubber could jump to an old position during a seek operation when it was dragged.
```

bitmovin-kenny (changes requested):
LGTM overall, just some minor nitpicks

src/ts/components/seekbar.ts
NikolaNikushev: I wanted to keep it separate, as there are already enough If statements above.

I will move the above to a separate variable and not have a nested if.

bitmovin-kenny (approved):
LGTM, but your build is failing

TeslaAdis (approved):
LGTM, makes sense.

PR #414, opened 2021-12-15
Refactor position handler 

NikolaNikushev:
Problem:

`playbackPositionHandler` has a lot of repeating logic, which is cluttered and does not allow easy testing. I am separating the calculations into standalone functions to allow cleaner reading of the live and vod case handling for the playback position.

### DIFF
--- a/src/ts/components/seekbar.ts
+++ b/src/ts/components/seekbar.ts
@@ -14,6 +14,7 @@ import { SeekBarType, SeekBarController } from './seekbarcontroller';
 import { i18n } from '../localization/i18n';
 import { BrowserUtils } from '../browserutils';
 import { TimelineMarkersHandler } from './timelinemarkershandler';
+import {getMinBufferLevel} from "./seekbarbufferlevel";
 
 /**
  * Configuration interface for the {@link SeekBar} component.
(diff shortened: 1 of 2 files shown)


src/ts/components/seekbar.ts
```diff
@@ -207,46 +207,62 @@ export class SeekBar extends Component<SeekBarConfig> {
     let isUserSeeking = false;
     let isPlayerSeeking = false;
 
+    const getPlaybackPositionPercentage = (): number => {
```
offnertho, line 210: I guess it is a preference thing, 

but why do you use `const doStuff = () => ` instead of `private doStuff function() {..}`?

src/ts/components/seekbar.ts
```diff
@@ -207,46 +207,62 @@ export class SeekBar extends Component<SeekBarConfig> {
...
+      if (player.isLive()) {
```
offnertho, line 211: I guess we could pass this in as a parameter. 

Or split it up even more and have 3 functions: 

```
getLivePlaybackPositionPercentage() {return getPlaybackPositionPercentage(true)}
getVodPlaybackPositionPercentage() {return getPlaybackPositionPercentage(false)}
getPlaybackPositionPercentage(isLive: boolean) {
... the stuff
}
```

src/ts/components/seekbar.ts
```diff
@@ -207,46 +207,62 @@ export class SeekBar extends Component<SeekBarConfig> {
...
+
+      const videoBufferLength = player.getVideoBufferLength();
+      const audioBufferLength = player.getAudioBufferLength();
+      // Calculate the buffer length which is the smaller length of the audio and video buffers. If one of these
+      // buffers is not available, we set it's value to MAX_VALUE to make sure that the other real value is taken
+      // as the buffer length.
+      let bufferLength = Math.min(
+          videoBufferLength != null ? videoBufferLength : Number.MAX_VALUE,
+          audioBufferLength != null ? audioBufferLength : Number.MAX_VALUE);
+      // If both buffer lengths are missing, we set the buffer length to zero
```
offnertho, line 235: this could be a function as well

offnertho (changes requested):
Nice refactoring. 

Some small remarks. but looking good.

src/ts/components/seekbar.ts
NikolaNikushev: Updated here https://github.com/bitmovin/bitmovin-player-ui/pull/414/commits/d0781dd012a4a1c057f7d1d476bd8d9a0164b893

src/ts/components/seekbar.ts
NikolaNikushev: Updated here https://github.com/bitmovin/bitmovin-player-ui/pull/414/commits/d0781dd012a4a1c057f7d1d476bd8d9a0164b893

src/ts/components/seekbar.ts
NikolaNikushev: For this case, I do not think this is necessary.

A bigger refactoring needs to be done here and move more of the `playbackPositionHandling` into a separate file, but currently, that is out of scope for this PR.

src/ts/components/seekbar.ts
offnertho: Sry i meant this whole block, could be a function `getMinBufferLevel`

src/ts/components/seekbar.ts
offnertho: 👍

src/ts/components/seekbar.ts
NikolaNikushev: https://github.com/bitmovin/bitmovin-player-ui/pull/414/commits/49b96bf065c026f15e48fb846086e98b2ff6d2a7

src/ts/components/seekbar.ts
```diff
@@ -14,6 +14,7 @@ import { SeekBarType, SeekBarController } from './seekbarcontroller';
 import { i18n } from '../localization/i18n';
 import { BrowserUtils } from '../browserutils';
 import { TimelineMarkersHandler } from './timelinemarkershandler';
+import {getMinBufferLevel} from './seekbarbufferlevel';
```
offnertho, line 17: Formatting?

offnertho (approved):

src/ts/components/seekbar.ts
NikolaNikushev: https://github.com/bitmovin/bitmovin-player-ui/pull/414/commits/3d3f6a61ed921c16b6408362b3bee0827f151e72 👍

PR #415, opened 2021-12-16
[Fix] Jumping scrubber when a segment is downloaded and the scrubber is dragged

NikolaNikushev:
Problem:
When using the scrubber on a touch device or dragging it with the mouse, there is a chance that the scrubber will bounce back to the previous position. 
This happens due to the segment request finished updating the playback location and the user is also is dragging the scrubber. Both operations happen in parallel but there is no seeking happening yet, since dragging does not start a seek operation until the user drops the seek bar.

Also it is possible that the `seeked` event fired from the player is not in sync with setting the seeking, since they are tracked in 2 different event callbacks.

Resolve:
- Keep track of the scrubbing action to not trigger a position update if there was a segment request finished. 
- Move both `seeked` handling callbacks into one callback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - The scrubber could jump to an old position during a seek operation when it was dragged.
+- The Seekbar scrubber could jump to an old position on touch devices when a new buffer is available during a seek operation.
 
 ## [3.31.0] - 2021-10-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - The scrubber could jump to an old position during a seek operation when it was dragged.
-- The Seekbar scrubber could jump to an old position on touch devices when a new buffer is available during a seek operation.
+- The Seekbar scrubber could jump to an old position on touch devices when the buffer updates during a seek operation.
 
 ## [3.31.0] - 2021-10-12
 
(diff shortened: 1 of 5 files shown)


spec/components/seekbar.spec.ts
```diff
@@ -81,5 +81,86 @@ describe('SeekBar', () => {
...
+      expect(setPlaybackPositionSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it('will move after a successful segment request download', function () {
+      playerMock.eventEmitter.fireSegmentRequestFinished();
+      expect(setPlaybackPositionSpy).toHaveBeenCalledTimes(1);
+    });
+
+    describe('vod event tracking', () => {
+
```
stonko1994, line 96: Please remove empty line

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - The scrubber could jump to an old position during a seek operation when it was dragged.
+- The Seekbar scrubber could jump to an old position on touch devices when a new buffer is available during a seek operation.
```
stonko1994, line 11: Nit:
```suggestion
- The Seekbar scrubber could jump to an old position on touch devices when the buffer updates during a seek operation.
```

spec/components/seekbar.spec.ts
```diff
@@ -81,5 +81,86 @@ describe('SeekBar', () => {
 
           expect(setPlaybackPositionSpy).toHaveBeenCalledTimes(timesCalled);
         })
+
+    it('will be moved after a successful seeked event', function () {
```
stonko1994, line 85: Any specific reason to use `function () {` over `() => {` 🤔

stonko1994 (approved):
Code looks good 👍 Just minor suggestions

TeslaAdis (approved):
LGTM!

PR #419, opened 2021-12-20
[PW-6220] Provide ability to override ListBox aria-labels

kishorens:
Problem: By default, we are using label/text as `aria-label` for list-item when Listbox is used. When complex HTML is used as a label customer wants to have an option to provide a different value for `aria-label`

Fix: provided an option to pass optional `aria-label` when  `additem` is called for adding new items to the Listbox

### DIFF
--- a/src/ts/components/listselector.ts
+++ b/src/ts/components/listselector.ts
@@ -109,7 +109,7 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
    * @param ariaLabel custom aria label for the listItem
    */
   addItem(key: string, label: LocalizableText, sortedInsert = false, ariaLabel = '') {
-    const listItem = { key: key, label: i18n.performLocalization(label), ...(ariaLabel && {ariaLabel: ariaLabel})};
+    const listItem = { key: key, label: i18n.performLocalization(label), ...(ariaLabel && { ariaLabel })};
 
     // Apply filter function
     if (this.config.filter && !this.config.filter(listItem)) {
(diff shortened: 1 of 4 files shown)


TeslaAdis (changes requested):
Overall looks good.
Missing changelog entry.

Can we also have a unit test when `ariaLabel` is not provided?

src/ts/components/listselector.ts
```diff
@@ -104,9 +106,10 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
    * @param key the key of the item to add
    * @param label the (human-readable) label of the item to add
    * @param sortedInsert whether the item should be added respecting the order of keys
+   * @param ariaLabel custom aria label for the listItem
    */
-  addItem(key: string, label: LocalizableText, sortedInsert = false) {
-    const listItem = { key: key, label: i18n.performLocalization(label) };
+  addItem(key: string, label: LocalizableText, sortedInsert = false, ariaLabel = '') {
+    const listItem = { key: key, label: i18n.performLocalization(label), ...(ariaLabel && {ariaLabel: ariaLabel})};
```
TeslaAdis, line 112:
```suggestion
    const listItem = { key: key, label: i18n.performLocalization(label), ...(ariaLabel && { ariaLabel })};
```

kishorens:
> Can we also have a unit test when ariaLabel is not provided?

@TeslaAdis  The existing unit tests already cover the cases where aria labels are not provided. 
```
  it('adds new items to the end of the list', () => {
      listSelector.addItem('A', 'itemA');
      listSelector.addItem('C', 'itemC');
      listSelector.addItem('B', 'itemB');

TeslaAdis (approved):

PR #423, opened 2022-01-05
[PW-6221]Added change to update aria-label on playbackbutton toggle

kishorens:
Problem: The `aria-label` is not updated when the playback button is toggled.

Fix: Added change to update the aria-label on playback button toggle

### DIFF
--- a/src/ts/components/playbacktogglebutton.ts
+++ b/src/ts/components/playbacktogglebutton.ts
@@ -39,8 +39,10 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
 
       if (player.isPlaying() || this.isPlayInitiated) {
         this.on();
+        this.getDomElement().attr('aria-label', i18n.performLocalization(i18n.getLocalizer('pause')));
       } else {
         this.off();
+        this.getDomElement().attr('aria-label', i18n.performLocalization(i18n.getLocalizer('play')));
       }
     };
 
--- a/src/ts/components/playbacktogglebutton.ts
+++ b/src/ts/components/playbacktogglebutton.ts
@@ -39,11 +39,15 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
 
       if (player.isPlaying() || this.isPlayInitiated) {
         this.on();
-        this.getDomElement().attr('aria-label', i18n.performLocalization(i18n.getLocalizer('pause')));
+        updateAriaLabel('pause');
       } else {
         this.off();
-        this.getDomElement().attr('aria-label', i18n.performLocalization(i18n.getLocalizer('play')));
+        updateAriaLabel('play');
       }
+    };
+
+    const updateAriaLabel = (label: string) => {
+      this.getDomElement().attr('aria-label', i18n.performLocalization(i18n.getLocalizer(label)));
     };
 
     // Call handler upon these events
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Updating the markers on live streams causing unhandled exception after player is destroyed
+- Incorrect aria-label on playbackbutton toggle
 
 ## [3.32.0] - 2021-12-21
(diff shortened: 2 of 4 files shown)


src/ts/components/playbacktogglebutton.ts
```diff
@@ -39,8 +39,10 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
 
       if (player.isPlaying() || this.isPlayInitiated) {
         this.on();
+        this.getDomElement().attr('aria-label', i18n.performLocalization(i18n.getLocalizer('pause')));
       } else {
         this.off();
+        this.getDomElement().attr('aria-label', i18n.performLocalization(i18n.getLocalizer('play')));
```
ravloony, line 45: Suggestion: You could maybe split these two virtually identical lines to a private method?

ravloony (changes requested):
I just had a small nitpick

src/ts/components/playbacktogglebutton.ts
```diff
@@ -39,11 +39,17 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
...
+        updateAriaLabel('pause');
       } else {
         this.off();
+        updateAriaLabel('play');
       }
     };
 
+    const updateAriaLabel = (label: string) => {
+      this.getDomElement().attr('aria-label', i18n.performLocalization(i18n.getLocalizer(label)));
+    };
```
stonko1994, line 51: Every component in our UI should in theory support the `ariaLabel`. Therefore this 'update' possibility should be moved to the `component.ts`.

Also, I think the `updateAriaLabel` function shouldn't hide the dependency to the localization feature. I would prefer that this function takes a `LocalizableText` as an input parameter.

stonko1994 (commented):
- Please don't use internal links/issue numbers in this repo as it's open-source (PR title)
- I had a quick look at the `component.ts` and saw that there is an `ariaLabel` inside the `ComponentConfig` though it's not used/picked up when building the DOM element. This seems like a bug to me. IMO at DOM element creation the `ariaLabel` should be applied.
  - Can you please verify this and create an internal issue for this if needed?

src/ts/components/playbacktogglebutton.ts
ravloony: @stonko1994 I had a stab at this here: https://github.com/bitmovin/bitmovin-player-ui/compare/feature/fix-pause-aria-label...feature/fix-pause-aria-label-2

stonko1994: I had a look at it but I'm not super happy with the changes to the `ToggleButtonConfig` 🤔 

I can see a problem that the `ToggleButtonConfig` extends from the `ButtonConfig` which again extends the `ComponentConfig`. That means that the `ToggleButtonConfig` now has 3 `areaLabel` related properties:

```ts
interface ToggleButtonConfig extends ButtonConfig {
  ariaLabel?: LocalizableText; // From the `ComponentConfig`
  onAriaLabel?: LocalizableText;
  offAriaLabel?: LocalizableText;
}
```

IMO that is now super confusing what to use when someone creates a custom component or initializes a `ToggleButton` with a custom configuration 🤔 

_+1: The mentioned bug from the review comment is still there_

ravloony: Fully agree, I wasn't sure what to do though, so I just left it to see what you'd say.

I can see two options:

```
interface ToggleButtonConfig extends ButtonConfig {
  ariaLabel?: LocalizableText; // From the `ComponentConfig`
  offAriaLabel?: LocalizableText;
}
```
and doing some funky type magic to replace the ariaLabel in the inherited class with something like

```
ariaLabel: { on?: LocalizableText, off?: LocalizableText }
```

stonko1994: Hm I'm not sure either 🤔 Both not ideal.

Do we really need the configuration option? What about manually managing it for the `ToggleButton` 🤔

ravloony: Discussion happened on Slack. All points resolved in #431

PR #425, opened 2022-01-14
Fix timeshift markers causing unhandled exception timeout

NikolaNikushev:
## Problem
When you launch a stream and destroy the player before the 'Play' event is sent out, the console will report an unhandled exception to update the markers every 1 second. This is caused due to the `configureLivePausedTimeshiftUpdater` scheduling the timeout every 1s, but we never clear this timeout if the player was destroyed.

## Fix
Listen for player destroyed and clear the `pausedTimeshiftUpdater`


wasp898 (approved):
Nice catch!

PR #426, opened 2022-01-17
Use error message from `ErrorEvent` when one is available

wasp898:
We have started attaching more context to error events in bitdash. This includes a message property as well (introduced in 8.74).

Change: use the message from the event if there is one, fallback to the default message.

Before:
Schermata 2022-01-17 alle 14 12 01

After:
Schermata 2022-01-17 alle 14 11 39

### DIFF
--- a/src/ts/errorutils.ts
+++ b/src/ts/errorutils.ts
@@ -73,7 +73,7 @@ export namespace ErrorUtils {
   };
 
   export const defaultWebErrorMessageTranslator: ErrorMessageTranslator = (error: ErrorEvent) => {
-    const errorMessage = ErrorUtils.defaultErrorMessages[error.code];
+    const errorMessage = error.message ?? ErrorUtils.defaultErrorMessages[error.code];
 
     if (errorMessage) {
       // Use the error message text if there is one
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [develop]
 
+### Changed
+- `PlayerError` events that provide a message are displayed instead of default messages
+
 ### Fixed
 - Updating the markers on live streams causing unhandled exception after player is destroyed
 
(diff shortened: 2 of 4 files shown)


src/ts/errorutils.ts
```diff
@@ -73,7 +73,7 @@ export namespace ErrorUtils {
   };
 
   export const defaultWebErrorMessageTranslator: ErrorMessageTranslator = (error: ErrorEvent) => {
-    const errorMessage = ErrorUtils.defaultErrorMessages[error.code];
+    const errorMessage = error.message ?? ErrorUtils.defaultErrorMessages[error.code];
```
stonko1994, line 76: Have you checked with the mobile teams if their `error.message` property should be used here as well?

felix-hoc, line 76: Are we sure that all our error messages are formatted properly? I think this is more important when displaying it in the UI than when logging to the console.

And also that they don't include any internal knowledge that shouldn't be public OR information that is not relevant to the end user.

wasp898: The handling for that is in a separate function, this should be Web only I believe.

felix-hoc (commented):
I'm just wondering about the content getting displayed...

Other than that it LGTM from a player-web perspective.

src/ts/errorutils.ts
wasp898: These error messages are brand new, so nothing unexpected should be exposed through these. Whether it's worth exposing them through the UI is a good question, we have a ticket to do it but I'll rediscuss with the team.

stonko1994: true that 🤔 missed that 👍

wasp898: You made an excellent point and we also decided that it's not worth it, as the message won't be more interesting to the user at all: quite the opposite. Thanks a lot for the input, we did not think this through hard enough 😅

wasp898:
Upon further reflection after @felix-hoc comments, we have decided that it's not worth exposing the error messages after all, given that the information that they provide is not useful to users in most of the cases.
The generic messages that exist already are fine as they are right now.

PR #431, opened 2022-01-31
Fix pause aria label

ravloony:
Problem: The aria-label is not updated when the playback button is toggled.

Fix:
- Added aria capabilities to the component parent class
- Added logic to the ToggleButton class to switch the aria label when button is toggled.

This PR replaces https://github.com/bitmovin/bitmovin-player-ui/pull/423


src/ts/components/playbacktogglebutton.ts
```diff
@@ -17,8 +17,9 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
 
     this.config = this.mergeConfig(config, {
       cssClass: 'ui-playbacktogglebutton',
-      text: i18n.getLocalizer('playPause'),
-      ariaLabel: i18n.getLocalizer('play'),
+      text: i18n.getLocalizer('play'),
+      onAriaLabel: i18n.getLocalizer('pause'),
+      offAriaLabel: i18n.getLocalizer('play'),
```
felix-hoc, line 22: We should also update `HugePlaybackToggleButton` accordingly - either replace `ariaLabel` with the two new properties, or just remove `ariaLabel` and rely on the `PlaybackToggleButton` super class.

felix-hoc (dismissed):
LGTM, but `HugePlaybackToggleButton` should also be updated.

src/ts/components/playbacktogglebutton.ts
ravloony: Done in 6aa57a2e

stonko1994 (approved):

PR #432, opened 2022-01-31
Add unit tests for aria label toggle

felix-hoc:
Adds unit tests for https://github.com/bitmovin/bitmovin-player-ui/pull/431

When a toggle button changes state through calling `on()` or `off()`, the `aria-label` should be updated accordingly.

### DIFF
--- a/spec/components/playbacktogglebutton.spec.ts
+++ b/spec/components/playbacktogglebutton.spec.ts
@@ -38,11 +38,12 @@ describe('ToggleButton', () => {
       const offLocalizer = () => 'off';
 
       const config: ToggleButtonConfig = {
-        onAriaLabel: offLocalizer,
+        offAriaLabel: offLocalizer,
       };
 
       toggleButton = new ToggleButton(config);
       toggleButton.on();
+      toggleButton.off();
 
       expect(mockDomElement.attr).toHaveBeenCalledWith('aria-label', 'off');
     });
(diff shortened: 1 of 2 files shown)


ravloony (approved):
Thanks!

spec/components/playbacktogglebutton.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    uiInstanceManagerMock = MockHelper.getUiInstanceManagerMock();
+
+    // Setup DOM Mock
+    mockDomElement = MockHelper.generateDOMMock();
+    jest.spyOn(ToggleButton.prototype, 'getDomElement').mockReturnValue(mockDomElement);
+  });
+
+  describe('aria label', () => {
+    it('should set on label when toggle state is on', () => {
+      const onLocalizer = () => 'on';
```
stonko1994, line 25: Can we add a second test using a string instead of the function? 🤔 To also cover the non-localized use-case.

spec/components/playbacktogglebutton.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+      toggleButton.on();
+
+      expect(mockDomElement.attr).toHaveBeenCalledWith('aria-label', 'on');
+    });
+
+    it('should set off label when toggle state is off', () => {
+      const offLocalizer = () => 'off';
+
+      const config: ToggleButtonConfig = {
+        onAriaLabel: offLocalizer,
```
stonko1994, line 41: I think you wanted to test the `offAriaLabel` here 🤔 

```suggestion
        offAriaLabel: offLocalizer,
```

spec/components/playbacktogglebutton.spec.ts
ravloony: Done in b1d4f036

spec/components/playbacktogglebutton.spec.ts
ravloony: Fixed in 8e967a21

stonko1994 (approved):

PR #435, opened 2022-02-10
Move changelog entry to correct section

jmsn:
The changelog entry for https://github.com/bitmovin/bitmovin-player-ui/pull/431 went into the section of an already released version instead of `develop`. 

To double-check, as per the `v3.33.0` tag https://github.com/bitmovin/bitmovin-player-ui/blob/v3.33.0/CHANGELOG.md it was released without the changes of said PR.


ravloony (approved):
Thanks! :see_no_evil:

PR #451, opened 2022-03-30
Line align "end" not correct for multiple line cues

bitsmith1010:
## Summary

Vtt cues with style
`00:00:00.000 --> 00:00:09.000 line:83%,end`
with multiple lines don't render cue box position correctly.

## Detail

Such cues would use css **margin-\*** properties to apply line alignment,(*) but the method uses the cue line height in the calculation. First commits in this branch corrected this with a line count factor, but it was still not accurate. One problem is that values in pixels are only calculated at a future step (e.g. in the user agent browser).

One solution is to use relative sizes and offset proportions and compensate vtt **lineAlign** in the CSS properties that implement vtt **line** (**top**, **bottom**, **left**, **right**). This is the approach in this branch.

There remains the problem of line align "center" style which would need to calculate the cue height, because at the moment I didn't find a way to express this in CSS. While this last WebVTT incompliance remains, in this branch a relative size of 5% for cue line replaces the 28 _px_ that was used before. This is standard and will scale to any resolution.

---

### Notes

(*) WebVtt line alignment: https://www.w3.org/TR/webvtt1/#webvtt-cue-line-alignment

### DIFF
--- a/spec/vttutils.spec.ts
+++ b/spec/vttutils.spec.ts
@@ -203,8 +203,8 @@ describe('Vtt Utils', () => {
 
               VttUtils.setVttCueBoxStyles(mockRegionContainer, subtitleOverLaySize);
 
-              expect(spyCss).toHaveBeenCalledTimes(13);
-              expect(spyCss).toHaveBeenNthCalledWith(10, 'bottom', '50%');
+              expect(spyCss).toHaveBeenCalledTimes(11);
+              expect(spyCss).toHaveBeenNthCalledWith(8, 'bottom', '50%');
             });
           });
         });
@@ -221,7 +221,7 @@ describe('Vtt Utils', () => {
 
           expect(spyCss).toHaveBeenCalledTimes(11);
           expect(spyCss).toHaveBeenNthCalledWith(6, 'writing-mode', 'vertical-lr');
-          expect(spyCss).toHaveBeenNthCalledWith(7, 'right', '0');
+          expect(spyCss).toHaveBeenNthCalledWith(8, 'left', '0');
         });
 
         describe('Line positioning', () => {
@@ -248,8 +248,8 @@ describe('Vtt Utils', () => {
 
             VttUtils.setVttCueBoxStyles(mockRegionContainer, subtitleOverLaySize);
 
-            expect(spyCss).toHaveBeenCalledTimes(12);
-            expect(spyCss).toHaveBeenNthCalledWith(9, 'right', '50%');
+            expect(spyCss).toHaveBeenCalledTimes(11);
+            expect(spyCss).toHaveBeenNthCalledWith(8, 'left', '50%');
           });
 
           it('should set positive line positioning', () => {
@@ -264,8 +264,8 @@ describe('Vtt Utils', () => {
 
             VttUtils.setVttCueBoxStyles(mockRegionContainer, subtitleOverLaySize);
 
-            expect(spyCss).toHaveBeenCalledTimes(12);
-            expect(spyCss).toHaveBeenNthCalledWith(9, 'right', '19.047619047619047%');
+            expect(spyCss).toHaveBeenCalledTimes(11);
+            expect(spyCss).toHaveBeenNthCalledWith(8, 'left', '19.047619047619047%');
           });
 
           it('should set negative line positioning', () => {
@@ -280,8 +280,8 @@ describe('Vtt Utils', () => {
 
             VttUtils.setVttCueBoxStyles(mockRegionContainer, subtitleOverLaySize);
 
-            expect(spyCss).toHaveBeenCalledTimes(12);
-            expect(spyCss).toHaveBeenNthCalledWith(9, 'right', '80.95238095238095%');
+            expect(spyCss).toHaveBeenCalledTimes(11);
+            expect(spyCss).toHaveBeenNthCalledWith(8, 'left', '80.95238095238095%');
           });
 
           describe('Line alignment', () => {
@@ -298,7 +298,7 @@ describe('Vtt Utils', () => {
 
               VttUtils.setVttCueBoxStyles(mockRegionContainer, subtitleOverLaySize);
 
-              expect(spyCss).toHaveBeenCalledTimes(12);
+              expect(spyCss).toHaveBeenCalledTimes(11);
             });
 
             it('should do center line alignment', () => {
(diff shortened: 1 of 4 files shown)


spec/vttutils.spec.ts
```diff
@@ -349,7 +349,7 @@ describe('Vtt Utils', () => {
 
           expect(spyCss).toHaveBeenCalledTimes(11);
           expect(spyCss).toHaveBeenNthCalledWith(6, 'writing-mode', 'vertical-rl');
-          expect(spyCss).toHaveBeenNthCalledWith(7, 'left', '0');
+          expect(spyCss).toHaveBeenNthCalledWith(8, 'right', '0');
```
bitsmith1010, line 352: here there is a new css mutation between the "writing-mode" and position, I didn't examine this

src/ts/vttutils.ts
```diff
@@ -4,6 +4,11 @@ import { DOM, Size } from './dom';
 
 // Our default height of a line
 const lineHeight = 28;
```
bitsmith1010, line 6: I think that this is still used in other parts of this vtt library

src/ts/vttutils.ts
```diff
@@ -60,7 +72,12 @@ const setVttLine = (
...
 ) => {
-  if (vtt.line === 'auto') {
+  const overlayReferenceEdge = DirectionPair.get(direction);
+  if (vtt.line === 'auto' && vtt.vertical) {
+    cueContainerDom.css(overlayReferenceEdge, '0');
+    return;
+  }
+  if (vtt.line === 'auto' && !vtt.vertical) {
     return;
   }
```
bitsmith1010, line 82: probably this can be done more organized

src/ts/vttutils.ts
```diff
@@ -77,35 +94,55 @@ const setVttLine = (
     relativeLinePosition = (100 * absoluteLinePosition) / subtitleOverLaySize.height;
   }
 
-  cueContainerDom.css(direction, `${relativeLinePosition}%`);
-  setVttLineAlign(cueContainerDom, vtt, direction);
+  if (vtt.lineAlign !== 'end')
+    cueContainerDom.css(
+      overlayReferenceEdge, `${relativeLinePosition}%`);
```
bitsmith1010, line 99: since the offset passes to `setVttLineAlign()` there is only need to set this for start alignment. Otherwise this css property is set 2 times for center alignment while 1 time is sufficient.

src/ts/vttutils.ts
```diff
@@ -77,35 +94,55 @@ const setVttLine = (
...
+//   Direction.Right, Direction.Left
+const setCueBoxPositionForVerticalWriting = (
+  cueContainerDom: DOM,
+  direction: Direction,
+  vtt: VTTProperties,
+  subtitleOverlaySize: Size,
+) => {
+    const writingMode = direction === Direction.Right ?
+      'vertical-lr' : 'vertical-rl';
+    const opositeEdge = direction;
```
bitsmith1010, line 139: this is not used I think

src/ts/vttutils.ts
```diff
@@ -138,6 +175,30 @@ const setVttPositionAlign = (cueContainerDom: DOM, vtt: VTTProperties, direction
...
+const countLines = (innerHtml: string) =>
+  innerHtml.split('<br />').length;
+
+// this is only a rough approximation based in a standard line height
+const setCssForCenterLineAlign = (
+  cueContainerDom: DOM,
+  direction: Direction,
+  relativeCueBoxPosition: number) => {
+  const overlayReferenceEdge = DirectionPair.get(direction);
+  const opositeEdge = DirectionPair.get(direction);
```
bitsmith1010, line 187: not used I think

src/ts/vttutils.ts
```diff
@@ -138,6 +175,30 @@ const setVttPositionAlign = (cueContainerDom: DOM, vtt: VTTProperties, direction
...
+  const centerOffset = lineHeightPercent * lineCount / 2;
+  const offset = relativeCueBoxPosition - centerOffset;
+  cueContainerDom.css(overlayReferenceEdge, `${offset}%`);
+};
+
+const setCssForEndLineAlign = (
+  cueContainerDom: DOM,
+  direction: Direction,
+  offset: number) => {
+      const overlayReferenceEdge = DirectionPair.get(direction);
```
bitsmith1010, line 197: not used

src/ts/vttutils.ts
```diff
@@ -146,8 +207,11 @@ export namespace VttUtils {
     const vtt = cueContainer.vtt;
     const cueContainerDom = cueContainer.getDomElement();
 
+
     setDefaultVttStyles(cueContainerDom, vtt);
-    setVttWritingDirection(cueContainerDom, vtt, subtitleOverlaySize);
+
+    lineCount = countLines(cueContainer.getText());
+    setVttWritingDirectionAndCueBoxPositioning(cueContainerDom, vtt, subtitleOverlaySize);
```
bitsmith1010, line 214: possibly to separate this to `setWritingDirection()` and `setCueBoxPositioning()`

bitsmith1010 (commented):
I see some minor changes, these can be possibly for a future merge.

src/ts/vttutils.ts
```diff
@@ -77,35 +94,54 @@ const setVttLine = (
...
-    setVttLine(cueContainerDom, vtt, Direction.Right, subtitleOverlaySize);
-  } else if (vtt.vertical === 'rl') {
-    cueContainerDom.css('writing-mode', 'vertical-rl');
-    cueContainerDom.css(Direction.Left, '0');
-    cueContainerDom.css(Direction.Top, '0');
-    setVttLine(cueContainerDom, vtt, Direction.Left, subtitleOverlaySize);
+    setVttLine(cueContainerDom, vtt, Direction.Bottom, subtitleOverlaySize);
+    break;
+  // "lr" is "left to right" so this is vertical growing right
+  case 'lr':
```
mimcz, line 118: Maybe lets use enum for these values

src/ts/vttutils.ts
```diff
@@ -4,6 +4,11 @@ import { DOM, Size } from './dom';
 
 // Our default height of a line
 const lineHeight = 28;
+
+// Default relative line height
+const lineHeightPercent = 5;
+let lineCount: number;
```
mimcz, line 10: Here we may use the default value

mimcz (approved):
Nice👌

PR #459, opened 2022-06-28
Add menu captions to the TV UI

kishorens:
Jira - https://bitmovin.atlassian.net/browse/SOL-3687

- [x] `CHANGELOG` entry


src/ts/components/menucaption.ts
```diff
@@ -0,0 +1,26 @@
+import {Label, LabelConfig} from './label';
+import {UIInstanceManager} from '../uimanager';
```
dweinber, line 2:
```suggestion
import { Label, LabelConfig } from './label';
import { UIInstanceManager } from '../uimanager';
```

src/ts/components/settingstogglebutton.ts
```diff
@@ -5,6 +5,7 @@ import {Component, ComponentConfig} from './component';
 import {ArrayUtils} from '../arrayutils';
 import { PlayerAPI } from 'bitmovin-player';
 import { i18n } from '../localization/i18n';
+import {Menucaption} from './menucaption';
```
dweinber, line 8:
```suggestion
import { Menucaption } from './menucaption';
```

src/ts/uifactory.ts
```diff
@@ -47,6 +47,7 @@ import { PlayerAPI } from 'bitmovin-player';
 import { i18n } from './localization/i18n';
 import { SubtitleListBox } from './components/subtitlelistbox';
 import { AudioTrackListBox } from './main';
+import {Menucaption} from './components/menucaption';
```
dweinber, line 50:
```suggestion
import { Menucaption } from './components/menucaption';
```

src/ts/uifactory.ts
```diff
@@ -456,17 +459,26 @@ export namespace UIFactory {
...
+                subtitleMenuCaption,
+                audioMenuCaption,
+              ],
+              cssClasses: ['ui-menucaption'],
+            }),
             new Container({
               components: [
                 new MetadataLabel({ content: MetadataLabelContent.Title }),
-                new SettingsToggleButton({
+                 new SettingsToggleButton({
```
dweinber, line 472: Please remove this space, it breaks the 2-spaces indentation used in this project

```suggestion
                new SettingsToggleButton({
```

dweinber (changes requested):
Thanks for working on this! The code looks good (apart from a few nitpicks about code style / white spaces).

I'm afraid we'll need to tweak the functionality and design though:
1. When the MenuCaption gets displayed, the SettingsToggleButtons get pushed down. I don't think we should do that, the buttons must not move.
2. The MenuCaption should be displayed when the SettingsToggleButton has the focus (e.g. when navigating via TAB there on desktop) as well as when the button is clicked (= menu is opened). The first part is not in place yet.

dweinber:
I've made a few changes to this PR:
- It uses the existing `Label` component instead of creating a new one
- It is shown when the SettingsToggleButton is focused (e.g. hovered or selected (but not activated) via spatial navigation) and hidden if not focused
- Position is now below the buttons, so that hide/show doesn't trigger layout changes.

@felix-hoc would appreciate if you could review this one.

Screenshots:
Screenshot 2024-06-28 at 15 23 22

Screenshot 2024-06-28 at 15 23 16

Screenshot 2024-06-28 at 15 23 26

Screenshot 2024-06-28 at 15 23 30

@bit-jacoba let me know if you're happy with that.

dweinber:
(Need to check why the CI failed 🤔)

dweinber:
(CI failure was just a tslint error, which is fixed now)

felix-hoc (commented):
Implementation looks OK from my side, just not sure about the design/concept. It feels a bit "unfinished". Perhaps we could at least center the labels underneath the buttons? But the `Audio Track` one might be a bit too long.

Alternatively, what if we just had the label displayed when the menu is opened? So as a header to the select box.

dweinber:
I changed the approach quite a bit as I realized that all the buttons already have span/label elements included. There's now a new (optional) config option for _all_ buttons (including `ToggleButton`, `SettingsToggleButton` etc) to show the label on hover/focus.

However, I'm still not 100% happy with this, as long texts could be cut off for buttons on the right or left side, but I don't have a better idea.

This is how it looks like now:

Screenshot 2024-07-10 at 12 45 38

Screenshot 2024-07-10 at 12 45 21

Screenshot 2024-07-10 at 12 45 32

src/scss/skin-modern/_skin-tv.scss
```diff
@@ -16,7 +16,8 @@
     }
   }
 
-  .#{$prefix}-ui-titlebar-top {
+  .#{$prefix}-ui-titlebar-top,
+  .#{$prefix}-ui-menucaption {
```
felix-hoc: I guess we don't need this anymore, right?
```suggestion
  .#{$prefix}-ui-titlebar-top {
```

src/ts/components/button.ts
```diff
@@ -22,12 +23,18 @@ export interface ButtonConfig extends ComponentConfig {
    * Default: false
    */
   acceptsTouchWithUiHidden?: boolean;
+
+  /**
+   *
```
felix-hoc: Please add a short description

felix-hoc (commented):
Nice, IMO this approach is better 👍🏼 

Could we also get rid of the blue glow in the background of the text?

src/scss/skin-modern/_skin-tv.scss
dweinber: Good catch, done in 5efe823e

src/ts/components/button.ts
dweinber: I did add that but didn't push 🤦 it's there now in 604c91ff

dweinber:
> Could we also get rid of the blue glow in the background of the text?

I really tried but didn't manage. Happy to receive any suggestions or help 😅

PR #461, opened 2022-08-18
set time format after ready event

bitsmith1010:
### The problem

It was found that in certain circumstances the time label would not count the hours, only the minutes and seconds that remain.

### The cause

The time label has a format that is defined when it's initialized. If the player duration when it's initialized < 1 hour, this format will be _mm:ss_ so the hour quantity would always be ignored.

```
        Player                                       Player UI
---------------------------------------------------------------
player not ready yet                                    |
player duration = 0                                     |
             |                                 UI time label initializes
             |                                 set time format, read duration
             |                                  < 1 hour => time format = mm:ss   
             |                                    display 00:00
player ready,                                            |
player duration = 01:30:00           =>            display 30:00
```

### DIFF
--- a/spec/components/playbacktimelabel.spec.ts
+++ b/spec/components/playbacktimelabel.spec.ts
@@ -76,7 +76,6 @@ describe('PlaybackTimeLabel', () => {
 
       jest.spyOn(playerMock, 'getDuration').mockReturnValue(100);
       playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
-      playerMock.eventEmitter.fireReadyEvent();
       expect(playbackTimeLabel.getText()).toEqual('01:40');
     });
 
--- a/spec/components/playbacktimelabel.spec.ts
+++ b/spec/components/playbacktimelabel.spec.ts
@@ -62,10 +62,13 @@ describe('PlaybackTimeLabel', () => {
       jest.spyOn(playerMock, 'getCurrentTime').mockReturnValue(30);
 
       playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
+      playerMock.eventEmitter.fireEvent<PlayerEventBase>(
+        {type: PlayerEvent.Ready, timestamp: Date.now()}
+      );
       expect(playbackTimeLabel.getText()).toEqual('01:10');
     });
 
-    it.only('displays the total time mm:ss', () => {
+    it('displays the total time mm:ss', () => {
       jest.spyOn(playerMock, 'isLive').mockReturnValue(false);
       playbackTimeLabel = new PlaybackTimeLabel({ timeLabelMode: PlaybackTimeLabelMode.TotalTime });
 
@@ -81,7 +84,7 @@ describe('PlaybackTimeLabel', () => {
       expect(playbackTimeLabel.getText()).toEqual('01:40');
     });
 
-    it.only('displays the total time hh:mm:ss', () => {
+    it('displays the total time hh:mm:ss', () => {
       jest.spyOn(playerMock, 'isLive').mockReturnValue(false);
       playbackTimeLabel = new PlaybackTimeLabel({ timeLabelMode: PlaybackTimeLabelMode.TotalTime });
 
--- a/spec/components/playbacktimelabel.spec.ts
+++ b/spec/components/playbacktimelabel.spec.ts
@@ -1,6 +1,7 @@
 import { MockHelper, TestingPlayerAPI } from '../helper/MockHelper';
 import { UIInstanceManager } from '../../src/ts/uimanager';
 import { PlaybackTimeLabel, PlaybackTimeLabelMode } from '../../src/ts/components/playbacktimelabel';
+import { PlayerEvent, PlayerEventBase} from 'bitmovin-player';
 
 const liveEdgeActiveCssClassName = 'ui-playbacktimelabel-live-edge';
 
@@ -64,7 +65,8 @@ describe('PlaybackTimeLabel', () => {
       expect(playbackTimeLabel.getText()).toEqual('01:10');
     });
 
-    it('displays the total time', () => {
+    it.only('displays the total time mm:ss', () => {
+      jest.spyOn(playerMock, 'isLive').mockReturnValue(false);
       playbackTimeLabel = new PlaybackTimeLabel({ timeLabelMode: PlaybackTimeLabelMode.TotalTime });
 
       // Setup DOM Mock
@@ -73,7 +75,29 @@ describe('PlaybackTimeLabel', () => {
 
       jest.spyOn(playerMock, 'getDuration').mockReturnValue(100);
       playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
+      playerMock.eventEmitter.fireEvent<PlayerEventBase>(
+        {type: PlayerEvent.Ready, timestamp: Date.now()}
+      );
       expect(playbackTimeLabel.getText()).toEqual('01:40');
+    });
+
+    it.only('displays the total time hh:mm:ss', () => {
+      jest.spyOn(playerMock, 'isLive').mockReturnValue(false);
+      playbackTimeLabel = new PlaybackTimeLabel({ timeLabelMode: PlaybackTimeLabelMode.TotalTime });
+
+      // Setup DOM Mock
+      const mockDomElement = MockHelper.generateDOMMock();
+      jest.spyOn(playbackTimeLabel, 'getDomElement').mockReturnValue(mockDomElement);
+
+      jest.spyOn(playerMock, 'getDuration').mockReturnValue(0);
+      expect(playbackTimeLabel.getText()).toEqual(undefined);
+
+      jest.spyOn(playerMock, 'getDuration').mockReturnValue(3600);
+      playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
+      playerMock.eventEmitter.fireEvent<PlayerEventBase>(
+        {type: PlayerEvent.Ready, timestamp: Date.now()}
+      );
+      expect(playbackTimeLabel.getText()).toEqual('01:00:00');
     });
   });
 
--- a/spec/components/playbacktimelabel.spec.ts
+++ b/spec/components/playbacktimelabel.spec.ts
@@ -62,9 +62,7 @@ describe('PlaybackTimeLabel', () => {
       jest.spyOn(playerMock, 'getCurrentTime').mockReturnValue(30);
 
       playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
-      playerMock.eventEmitter.fireEvent<PlayerEventBase>(
-        {type: PlayerEvent.Ready, timestamp: Date.now()}
-      );
+      playerMock.eventEmitter.fireReadyEvent();
       expect(playbackTimeLabel.getText()).toEqual('01:10');
     });
 
@@ -78,9 +76,7 @@ describe('PlaybackTimeLabel', () => {
 
       jest.spyOn(playerMock, 'getDuration').mockReturnValue(100);
       playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
-      playerMock.eventEmitter.fireEvent<PlayerEventBase>(
-        {type: PlayerEvent.Ready, timestamp: Date.now()}
-      );
+      playerMock.eventEmitter.fireReadyEvent();
       expect(playbackTimeLabel.getText()).toEqual('01:40');
     });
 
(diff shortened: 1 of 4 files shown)


bitsmith1010:
the review can begin, it will be pending to manual test on Android tomorrow

src/ts/components/playbacktimelabel.ts
```diff
@@ -132,7 +134,15 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
       }
     };
 
+    let playbackReadyHandler = () => {
+      // Set time format depending on source duration
+      this.timeFormat = Math.abs(player.isLive() ? player.getMaxTimeShift() : player.getDuration()) >= 3600 ?
```
bitsmith1010, line 139: It seems to me that a format string could be used to change between various string formats. In the latest release, the only formats are _mm:ss_ and _hh:mm:ss_, but the first doesn't map to the variable's range (there could be time quantities > 1 hour). It's not clear to me then in what use case this would apply...

Is the use case that the user of the library could select _mm:ss_ even if time is > 1 hour where the Player logically should trigger to display the correct value (_hh:mm:ss_)?

I would ask what exactly this time format should do - 
- to remove the hours field when = 0?
- to allow the user to select a time format? (hh:mm:ss, total minutes:seconds, total seconds, etc)
- to ignore the hour value?

bitsmith1010: one note on the last comment

> Is the use case that the user of the library could select mm:ss even if time is > 1 hour where the Player logically should trigger to display the correct value (hh:mm:ss)?

If the use case is that the user of the library controls this format, then would it be inconsistent if the Bitmovin UI would change this format when it's value is out of (the format's) range? The following program trajectory could occur:

- The user uses `PlaybackTimeLabel.setTimeFormat()` to set _mm:ss_
- The duration changes and is > 1h
- The component observes this and corrects the format to _hh:mm:ss_

So the user would not be able to control the format if the component would correct the duration during the program trajectory.

src/ts/components/playbacktimelabel.ts
```diff
@@ -132,7 +134,15 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
...
+      StringUtils.FORMAT_HHMMSS : StringUtils.FORMAT_MMSS;
+      playbackTimeHandler();
+    };
+
     player.on(player.exports.PlayerEvent.TimeChanged, playbackTimeHandler);
+    player.on(player.exports.PlayerEvent.Ready, playbackReadyHandler);
```
wasp898, line 145: The only problem that I'm seeing is that imho duration should be available after `SourceLoaded`. If that's not the case, it could be a bug on the Player SDK (I guess Android in this case). If duration changes afterwards, there should be a `DurationChanged` event. If this also does not happen on Android, maybe we should argue that they should introduce such an event, since it seems to be relevant.

If we still decide to treat this as a UI issue, I'm not convinced setting the duration after the `Ready` event is the best approch. `Ready` indicates that content is playable, but duration should be already available much earlier.

wasp898: @matamegger @strangesource are you guys sure all this is handled correctly on the Android Player side? Meaning that everything is by design? Are we sure this should be addressed on the UI side?

@stonko1994 do you mind sharing your opinion as well?

Thanks.

bitsmith1010: in fact, I can't even play this in the Chromecast receiver, I tried the SDK version 3. Also note that in the jira item the version number is 8.80, this would be a Web Player... 

at the same time, in the jira incident to use the Android samples...

bitsmith1010: but even without to test it, it's clear that the "mm:ss" could be selected before the duration is known, so if we know what event to await we can merge this and close the item.

If we don't know in what setup the client used, the manual test can be done later once we have this information.

strangesource: Hi @bitsmith1010, @wasp898 👋,

as already [stated in the Android issue](https://bitmovin.atlassian.net/browse/PA-1661?focusedCommentId=235624): When we start casting a stream, we have to wait for the duration to be transmitted from the cast receiver. This takes some time and can result in an invalid duration being available when the Bitmovin Web UI starts up. 

I see your point that the duration should be available after the loaded event. Unfortunately this is currently not the case when the cast device is already connected before loading the source into the player as the loaded event is solely emitted for local source loading. 

Having said that, I do see a discrepancy on the UI side here where the **duration is updated but the format is not** resulting in the "cropped" duration. 
This caused my reasoning that the issue could/should be handled in the UI.

stonko1994: I would agree with both sides here 🤔!

As I understand it, there seems to be an issue on the Android SDK side where the duration is not available where it should be, and the change of it is not indicated via a `DurationChanged` event.

On the other hand, I also agree that if the UI already updates the duration on other occasions (in this case, based on `TimeChanged` events), why not also update the time format?

spec/components/playbacktimelabel.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { MockHelper, TestingPlayerAPI } from '../helper/MockHelper';
 import { UIInstanceManager } from '../../src/ts/uimanager';
 import { PlaybackTimeLabel, PlaybackTimeLabelMode } from '../../src/ts/components/playbacktimelabel';
+import { PlayerEvent, PlayerEventBase} from 'bitmovin-player';
```
stonko1994, line 4:
```suggestion
import { PlayerEvent, PlayerEventBase } from 'bitmovin-player';
```

spec/components/playbacktimelabel.spec.ts
```diff
@@ -61,10 +62,14 @@ describe('PlaybackTimeLabel', () => {
       jest.spyOn(playerMock, 'getCurrentTime').mockReturnValue(30);
 
       playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
+      playerMock.eventEmitter.fireEvent<PlayerEventBase>(
+        {type: PlayerEvent.Ready, timestamp: Date.now()}
+      );
```
stonko1994, line 67: Please add the event emitting of the player mock to the `PlayerEventEmitter.ts` as we do for all other events. **(All occurrences)**

spec/components/playbacktimelabel.spec.ts
```diff
@@ -61,10 +62,14 @@ describe('PlaybackTimeLabel', () => {
...
       expect(playbackTimeLabel.getText()).toEqual('01:10');
     });
 
-    it('displays the total time', () => {
+    it('displays the total time mm:ss', () => {
```
stonko1994, line 71: Could we add the use-case into the test name for easier understanding?

```suggestion
    it('displays the total time mm:ss if duration is lower than 1 hour', () => {
```

spec/components/playbacktimelabel.spec.ts
```diff
@@ -73,8 +78,30 @@ describe('PlaybackTimeLabel', () => {
 
       jest.spyOn(playerMock, 'getDuration').mockReturnValue(100);
       playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
+      playerMock.eventEmitter.fireEvent<PlayerEventBase>(
+        {type: PlayerEvent.Ready, timestamp: Date.now()}
+      );
       expect(playbackTimeLabel.getText()).toEqual('01:40');
     });
+
+    it('displays the total time hh:mm:ss', () => {
```
stonko1994, line 87:
```suggestion
    it('displays the total time hh:mm:ss if duration is greater than 1 hour', () => {
```

spec/components/playbacktimelabel.spec.ts
```diff
@@ -73,8 +78,30 @@ describe('PlaybackTimeLabel', () => {
...
+      playbackTimeLabel = new PlaybackTimeLabel({ timeLabelMode: PlaybackTimeLabelMode.TotalTime });
+
+      // Setup DOM Mock
+      const mockDomElement = MockHelper.generateDOMMock();
+      jest.spyOn(playbackTimeLabel, 'getDomElement').mockReturnValue(mockDomElement);
+
+      jest.spyOn(playerMock, 'getDuration').mockReturnValue(0);
+      expect(playbackTimeLabel.getText()).toEqual(undefined);
+
+      jest.spyOn(playerMock, 'getDuration').mockReturnValue(3600);
```
stonko1994, line 98: I would create a separate test for this where you explicitly test the switch from 0 to something greater than 1 hour based on the given event.

IMO to really test the use-case we are talking about you would need to emit a SourceLoaded event between changing the duration

src/ts/components/playbacktimelabel.ts
```diff
@@ -132,7 +134,15 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
       }
     };
 
+    let playbackReadyHandler = () => {
+      // Set time format depending on source duration
+      this.timeFormat = Math.abs(player.isLive() ? player.getMaxTimeShift() : player.getDuration()) >= 3600 ?
+      StringUtils.FORMAT_HHMMSS : StringUtils.FORMAT_MMSS;
```
stonko1994, line 140:
```suggestion
      this.timeFormat = Math.abs(player.isLive() ? player.getMaxTimeShift() : player.getDuration()) >= 3600 ?
        StringUtils.FORMAT_HHMMSS : StringUtils.FORMAT_MMSS;
```

stonko1994, line 140: Could we extract this into a `updateTimeFormatBasedOnDuration` function and maybe also call it from the `playbackTimeHandler`? 🤔 Then we would be fully safe to always show the correct format (especially as I don't see a `DurationChanged` handler)

stonko1994 (changes requested):

src/ts/components/playbacktimelabel.ts
wasp898: Okay @strangesource, thanks.

Then my only suggestion to @bitsmith1010 is that we update the format both on initialization (as it was) and after `Ready` event, and not only on `Ready`. The suggestion from David is also valid.

src/ts/components/playbacktimelabel.ts
```diff
@@ -132,7 +134,15 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
       }
     };
 
+    let playbackReadyHandler = () => {
```
wasp898, line 137: Possible name. Also we need to call this function by default on initialization.
```suggestion
    let updatePlaybackTimeFormat = () => {
```

spec/components/playbacktimelabel.spec.ts
bitsmith1010: at the end we used `Ready`

spec/components/playbacktimelabel.spec.ts
```diff
@@ -73,8 +76,39 @@ describe('PlaybackTimeLabel', () => {
 
       jest.spyOn(playerMock, 'getDuration').mockReturnValue(100);
       playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
+      playerMock.eventEmitter.fireReadyEvent();
```
wasp898, line 79: We don't need this anymore.
```suggestion
```

wasp898 (approved):
Just one small thing left. Other than that LGTM now 👍

spec/components/playbacktimelabel.spec.ts
```diff
@@ -61,10 +62,12 @@ describe('PlaybackTimeLabel', () => {
       jest.spyOn(playerMock, 'getCurrentTime').mockReturnValue(30);
 
       playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
+      playerMock.eventEmitter.fireReadyEvent();
```
wasp898, line 65:
```suggestion
```

bitsmith1010:
manual test results:

functions good, time format is correct with source > 1hr with Chromecast and Android Player

src/ts/components/playbacktimelabel.ts
```diff
@@ -132,7 +134,15 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
       }
     };
 
+    let updateTimeFormatBasedOnDuration = () => {
+      // Set time format depending on source duration
+      this.timeFormat = Math.abs(player.isLive() ? player.getMaxTimeShift() : player.getDuration()) >= 3600 ?
+      StringUtils.FORMAT_HHMMSS : StringUtils.FORMAT_MMSS;
```
stonko1994, line 140: Little formating for better readability:

```suggestion
      this.timeFormat = Math.abs(player.isLive() ? player.getMaxTimeShift() : player.getDuration()) >= 3600 ?
        StringUtils.FORMAT_HHMMSS : StringUtils.FORMAT_MMSS;
```

stonko1994 (approved):
One minor formating issue but approving already 👍

PR #468, opened 2022-10-13
Implement spatial navigation

HalfbyteHeroes:
This PR implements spatial navigation based on our `Component`s & `Container`s.  The main components are explained below.

### `NavigationGroup`
Takes the container as the first parameter and elements that are navigatable inside that container.
 It is responsible for tracking active elements, dispatching focus, and blurring on elements changing. It hides the container on back action. It provides public `onNavigation` and `onAction` method properties which can be overwritten and be used as an event handler for listening to internal changes. Returning `true` in those methods will not execute default navigation and action handler.

### `RootNavigationGroup`
Extends `NavigationGroup` and implements additional logic for actions to hide/show UI.

### `SpatialNavigation`
Takes `RootNavigationGroup` as the first parameter, and the rest are other `NavigationGroups`. It is responsible for tracking all of the navigation groups, and for triggering `enable`/`disable` on them based on the container `onShow` and `onHide` subscribers. Listens to keyboard events and triggers handle navigation and handle action on the active group.
 

### `SeekBarHandler`
Overwrites `onNavigation` and `onAction` on the `RootNavigationGroup` and implements scrubbing with the preview tooltip. Pressing enter will seek to the scrubbed position, pressing back will cancel. Scrub speed increases over time as long as the left/right buttons are pressed, and will reset once they are released.
 

### `NavigationAlgorithm`
This is the algorithm used to select the next element when attempting to navigate away from the current one. Its based on angle and distance. It calculates angles between elements, rules out all of the elements that have angles greater than 45 degrees, and picks the closest one based on distance and angle as the next element.
 

### Testing
To test it, run `gulp serve` and select `modernTVUi` from UI drop down in the testing page. Then, focus the player and use the arrow keys as well as `Esc` and `Enter` to navigate the elements. (`ModernTVUi` factory is not final)


src/ts/spatialnavigation/navigationalgorithm.ts
```diff
@@ -0,0 +1,78 @@
...
+    x: point2.x - point1.x,
+    y: point2.y - point1.y,
+  });
+
+  const angleCos = dotProduct(directionVector, elementVector) / (length(directionVector) * length(elementVector));
+
+  return Math.acos(angleCos) * 180 / Math.PI;
+}
+
+export function getElementInDirection(
```
felix-hoc, line 57: Functions in this file could use some comments. Readability isn't great when someone isn't that familiar with vector maths.
E.g., what's the unit of the angle returned by `calculateAngle`? deg or rad?
what are we sorting elements by?
...etc.

src/ts/spatialnavigation/navigationgroup.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  public focusElement(element: HTMLElement): void {
+    this.blurActiveElement();
+    this.activeElement = element;
+    this.activeElement.focus();
+  }
+
+  public blurActiveElement(): void {
+    this.activeElement?.blur();
+    this.activeElement = undefined;
```
felix-hoc, line 54: why is the `activeElement` unset here? Seems a bit confusing.
Maybe just blur it here and unset it in `disable()` instead?

src/ts/spatialnavigation/navigationgroup.ts
```diff
@@ -0,0 +1,141 @@
...
+    this.components = components;
+    this.eventSubscriber = new NodeEventSubscriber();
+  }
+
+  /**
+   * Adds onNavigation override
+   * @param direction {Direction}
+   * @param target {HTMLElement}
+   *
+   * @returns Boolean - true means it should stop propagation, and false means it should be handled by default handler
```
felix-hoc, line 29: Assigning such meaning to a callback return value seems a bit arbitrary. `return true;` or `return false;` has no implicit meaning in a user-implemented function.
Have you considered to instead pass a `preventDefault()` function instead?

src/ts/spatialnavigation/spatialnavigation.ts
```diff
@@ -0,0 +1,170 @@
...
+  LEFT = 'left',
+  RIGHT = 'right',
+}
+
+export enum Action {
+  SELECT = 'select',
+  BACK = 'back',
+}
+
+const KeyMap: {[key: number]: Direction | Action} = {
```
felix-hoc, line 18: where does that keymap come from? Which devices can we expect it to work with? (maybe add a comment)

src/ts/spatialnavigation/seekbarhandler.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  public release(): void {
+    this.eventSubscriber.release();
+    this.rootNavigationGroup.onAction = undefined;
+    this.rootNavigationGroup.onNavigation = undefined;
+  }
+}
+
+function isSeekBarWrapper(element: HTMLElement): boolean {
+  return Array.from(element.classList).findIndex(className => className.includes('-ui-seekbar')) > -1;
```
felix-hoc, line 145: this would also match `bmpui-ui-seekbar-label`. Isn't that a problem?

src/ts/spatialnavigation/spatialnavigation.ts
```diff
@@ -0,0 +1,170 @@
...
+      group.disable();
+      this.activeNavigationGroups.splice(groupIndex, 1);
+      this.updateEnabledNavigationGroup();
+    }
+  };
+
+  private subscribeToNavigationGroupVisibilityChanges(): void {
+    this.navigationGroups.forEach(group => {
+      const onShowHandler = this.onShow.bind(this, group);
+      const onHideHandler = this.onHide.bind(this, group);
```
felix-hoc, line 94: Do we _need_ to bind these? They are declared as arrow functions, that should retain the scope right?

felix-hoc (changes requested):
Implementation seems to be fine, had a few comments.

In general, I would welcome a bit more code doc to make it easier for external users to get started (e.g., describing the purpose of the added classes, and also individual functions).

Did not test or look at unit tests yet.

src/ts/spatialnavigation/navigationgroup.ts
TeslaAdis: Makes sense, addressed in [f97b46b](https://github.com/bitmovin/bitmovin-player-ui/pull/468/commits/f97b46b10c03d2fcae542155352f265bb13a6276)

src/ts/spatialnavigation/navigationgroup.ts
HalfbyteHeroes: Agreed. Fixed in be6e4b57 - we're now using `preventDefault`.

src/ts/spatialnavigation/spatialnavigation.ts
HalfbyteHeroes: No, we don't 😄 Changed in da004f28 - using error functions instead now.

src/ts/spatialnavigation/seekbarhandler.ts
HalfbyteHeroes: Fair point. Using a more strict check in 06f9e8eb.

TeslaAdis: Using regex now - addressed in [06f9e8e](https://github.com/bitmovin/bitmovin-player-ui/pull/468/commits/06f9e8eb512a641311758adea1ce73111097ea65)

src/ts/spatialnavigation/spatialnavigation.ts
HalfbyteHeroes: Restructured the key maps in c52e166b 👍

TeslaAdis: We have refactored it to have different key groups for different devices - will give a bit more insight.
Addressed in [c52e166](https://github.com/bitmovin/bitmovin-player-ui/pull/468/commits/c52e166b1316399acc965b1f8c306f9e44967912)

src/ts/browserutils.ts
```diff
@@ -39,6 +39,38 @@ export class BrowserUtils {
...
   }
 
+  static get isHisense(): boolean {
+    if (!this.windowExists()) {
+      return false;
+    }
+    return navigator && navigator.userAgent && /Hisense/.test(navigator.userAgent);
+  }
+
+  static get isPlayStation(): boolean {
```
felix-hoc, line 49: nit: we're only testing for `Playstation 5` in the UA, so the function name should reflect that.

src/ts/spatialnavigation/keymap.ts
```diff
@@ -0,0 +1,129 @@
...
+    // D-pad Down
+    40: Direction.DOWN,
+    // D-pad Left
+    37: Direction.LEFT,
+    // D-pad Right
+    39: Direction.RIGHT,
+    // D-pad OK
+    13: Action.SELECT,
+    // Back
+    461: Action.BACK,
```
felix-hoc, line 36: I like this file a lot, very nicely documented 🤩  and easily extendable 👌🏼

felix-hoc (approved):
Very nice 💪🏼 
Thanks for adding the code doc, and good calls on choosing where it was needed. I think that adds a lot of value.

Additional question: Should the spatial navigation UI be mentioned in the readme?

src/ts/components/uicontainer.ts
```diff
@@ -49,6 +49,9 @@ export class UIContainer extends Container<UIContainerConfig> {
   private userInteractionEventSource: DOM;
   private userInteractionEvents: { name: string, handler: EventListenerOrEventListenerObject }[];
 
+  public hideUi: () => void = () => {};
+  public showUi: () => void = () => {};
```
stonko1994, line 53: I had a closer look at what it means to make those two methods public.

_IMO the usage of this feels weird._

### Why?

In order to create a UI it's very handy to just use our provided Factories (same as we do in `index.html`)

Create the UI:
```
    // Add UI
    uiManager = bitmovin.playerui.UIFactory.buildDefaultUI(player, uiConfig);
```

So now a user has this `uiManager` at hand. One would expect to be able to fully toggle the UI visible. So where would you look for this? I would assume on the `uiManager` as this is the thing to interact with.

Though there is nothing :travolta:

Instead, you need to know the internals of our UI and how it's setup (multiple UI variants inside one `uimanager`)

Ok how to access those UIContainer instances now 🤔  let's search for them 💪 The only thing you find will be the `uiManager.getUiVariants()` so you would need to do `uiManager.getUiVariants().forEach((variant) => variant.ui.hideUi())` which is super inconvenient.

---

I would propose to now expose them now or add something on the `uiManager` to enable a full visibility toggle hiding all the `uiVariant` complexity.

src/ts/spatialnavigation/keymap.ts
```diff
@@ -0,0 +1,129 @@
...
+  // Enter
+  13: Action.SELECT,
+  // Escape
+  27: Action.BACK,
+};
+
+/**
+ * Returns the matching key map for the current platform.
+ */
+export function getKeyMapForPlatform(): KeyMap {
```
stonko1994, line 115: Would it make sense to make this somehow configurable to allow customizing and adding new platforms as well? 🤔

src/ts/spatialnavigation/navigationalgorithm.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  return elements
+    // don't take the current element into account
+    .filter(elem => elem !== activeElement)
+    // get the angle between, and distance to any other element from the current element
+    .map(elem => {
+      const elemVector = getElementVector(elem);
+      const dist = distance(activeElemVector, elemVector);
+      const angle = calculateAngle(activeElemVector, elemVector, direction);
+      return [angle, dist, elem] as const;
```
stonko1994, line 118: Nit: would it make sense to use an object with named parameters here? 🤔 I think that would allow to use the name of the property in `.shift()?.[2];` instead of `2`

stonko1994 (commented):
Is there a way to manually test this without any of the mentioned platforms?

---

The old Safari User using the new UI with mouse and keyboard sees this -> (screenshot) Not sure if this is designed to be used or if this can be ignored

Screenshot 2022-10-19 at 12 37 31

src/ts/browserutils.ts
HalfbyteHeroes: Changed the check in d6111066 to check for PlayStation in general, not PlayStation 5 in particular.

src/ts/spatialnavigation/keymap.ts
TeslaAdis: It is open source, so anyone could extend this if needed. Don't think we need to allow configuring this from the outside.

src/ts/spatialnavigation/navigationalgorithm.ts
TeslaAdis: Addressed in [e3e404c](https://github.com/bitmovin/bitmovin-player-ui/pull/468/commits/e3e404ce668e068d91d4800e0d6c707b433ebddf)

PR #472, opened 2022-11-07
PW-8039 Fix/spatialnavigation/bounding client rect

mcarriga:
On tested Tizen and Webos systems, the HTMLElement.`getBoundingClientRect()` returns a **ClientRect** object instead of the expected **DOMRect** object. The **ClientRect** object does NOT contain `x` and `y` properties but instead both **ClientRect** and **DOMRect** do contain `left` and `top` properties.

~~Changed navigationalgorithm.`getElementVector` to use `left` and `top`~~
~~Changed seekbarhandler.`initializeCursorPosition` to use `left` and `top`~~

Changed navigationalgorithm.`getElementVector` to use `left` and `top` in instances where `x` and `y` are not available
Changed seekbarhandler.`initializeCursorPosition` to use `left` and `top` in instances where `x` and `y` are not available

### DIFF
--- a/src/ts/spatialnavigation/navigationalgorithm.ts
+++ b/src/ts/spatialnavigation/navigationalgorithm.ts
@@ -57,22 +57,11 @@ function distance(a: Vector, b: Vector): number {
  * @param element The element to get the center of
  */
 function getElementVector(element: HTMLElement): Vector {
-  const boundingRect = element.getBoundingClientRect();
-
-  let _x: number;
-  let _y: number;
-
-  if (boundingRect.hasOwnProperty('x') && boundingRect.hasOwnProperty('y')) {
-    _x = boundingRect.x;
-    _y = boundingRect.y;
-  } else {
-    _x = boundingRect.left;
-    _y = boundingRect.top;
-  }
+  const boundingRect = getBoundingRectFromElement(element);
 
   return {
-    x: _x + boundingRect.width / 2,
-    y: _y + boundingRect.height / 2,
+    x: boundingRect.x + boundingRect.width / 2,
+    y: boundingRect.y + boundingRect.height / 2,
   };
 }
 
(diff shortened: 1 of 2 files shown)


src/ts/spatialnavigation/navigationalgorithm.ts
```diff
@@ -59,9 +59,20 @@ function distance(a: Vector, b: Vector): number {
...
+  let _x: number;
+  let _y: number;
+
+  if (boundingRect.hasOwnProperty('x') && boundingRect.hasOwnProperty('y')) {
+    _x = boundingRect.x;
+    _y = boundingRect.y;
+  } else {
+    _x = boundingRect.left;
+    _y = boundingRect.top;
+  }
```
HalfbyteHeroes, line 71: Seeing that we use that code in 2 places, can we move it into a separate function that we can use in both of them?

src/ts/spatialnavigation/navigationalgorithm.ts
```diff
@@ -59,9 +59,20 @@ function distance(a: Vector, b: Vector): number {
 function getElementVector(element: HTMLElement): Vector {
   const boundingRect = element.getBoundingClientRect();
 
+  let _x: number;
+  let _y: number;
+
+  if (boundingRect.hasOwnProperty('x') && boundingRect.hasOwnProperty('y')) {
```
HalfbyteHeroes, line 65: I'd suggest we use
```suggestion
  if (typeof boundingRect.x === 'number' && typeof boundingRect.y === 'number') {
```
instead of `hasOwnProperty` to prevent any issues with inheritance.

src/ts/spatialnavigation/navigationalgorithm.ts
mcarriga: agreed - added exported function `getBoundingRectFromElement` in latest commit

src/ts/spatialnavigation/navigationalgorithm.ts
mcarriga: Makes sense - especially considering jsdom has not fully implemented the DOMRect class. Updated in latest commit

TeslaAdis (approved):

PR #476, opened 2022-11-18
Improve track list spatial navigation

HalfbyteHeroes:
The current spatial navigation implementation did not close list/track navigation groups if the user clicked `Left` or `Right` buttons of the remote or when a list item/track was selected.

This PR therefore introduces a `ListNavigationGroup` that automatically closes the navigation group when user navigates left or right (or up and down depending on the `ListOrientation`) and when a list item was selected.


mcarriga (approved):
looks great! tested on Tizen and WebOS successfully!

PR #477, opened 2022-11-18
Enable usage of ErrorMessageOverlay without Player Errors

dweinber:
Added a `display(errorMessage: string): void` method to the `ErrorMessageOverlay` to enable its usage also for application errors without player errors while keeping the same look and feel.


stonko1994 (approved):
LGTM 🙂 👍

PR #480, opened 2022-11-22
Add new ReplayButton component

dweinber:
Add new ReplayButton component that can be used in controlbar and works during playback

### DIFF
--- a/src/scss/skin-modern/components/_replaybutton.scss
+++ b/src/scss/skin-modern/components/_replaybutton.scss
@@ -1,0 +1,12 @@
+@import '../variables';
+@import '../mixins';
+
+.#{$prefix}-ui-replaybutton {
+  @extend %ui-button;
+
+  background-image: svg('assets/skin-modern/images/replay.svg');
+
+  &:hover {
+    @include svg-icon-shadow;
+  }
+}
--- a/src/scss/skin-modern/components/_replaybutton.scss
+++ b/src/scss/skin-modern/components/_replaybutton.scss
@@ -4,7 +4,7 @@
 .#{$prefix}-ui-replaybutton {
   @extend %ui-button;
 
-  background-image: svg('assets/skin-modern/images/prev.svg');
+  background-image: svg('assets/skin-modern/images/replay-nocircle.svg');
 
   &:hover {
     @include svg-icon-shadow;
--- a/src/scss/skin-modern/components/_replaybutton.scss
+++ b/src/scss/skin-modern/components/_replaybutton.scss
@@ -4,7 +4,7 @@
 .#{$prefix}-ui-replaybutton {
   @extend %ui-button;
 
-  background-image: svg('assets/skin-modern/images/replay.svg');
+  background-image: svg('assets/skin-modern/images/prev.svg');
 
   &:hover {
     @include svg-icon-shadow;
--- a/src/ts/components/replaybutton.ts
+++ b/src/ts/components/replaybutton.ts
@@ -33,7 +33,7 @@ export class ReplayButton extends Button<ButtonConfig> {
       } else {
         this.show();
       }
-    })
+    });
 
     this.onClick.subscribe(() => {
       if (!player.hasEnded()) {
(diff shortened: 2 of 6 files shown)


src/ts/components/replaybutton.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
+    super.configure(player, uimanager);
+
+    this.onClick.subscribe(() => {
+      if (!player.hasEnded()) {
+        player.seek(0);
+      }
+      player.play('ui');
```
offnertho, line 27: so we would call `play` either way? So if we are currently paused and the user hits the button we would play again? 

Just wondering if this was considered from an UX perspective.

src/scss/skin-modern/components/_replaybutton.scss
```diff
@@ -0,0 +1,12 @@
+@import '../variables';
+@import '../mixins';
+
+.#{$prefix}-ui-replaybutton {
+  @extend %ui-button;
+
+  background-image: svg('assets/skin-modern/images/prev.svg');
```
offnertho, line 7: I guess this is the same Image as our large replay icon at the end of playback?

dweinber: No it's not, and on purpose. The large replay icon put in the small controlbar looks very strange as 
a) it's hard to see due to the circle around it and so on, and 
b) due to it's design also doesn't fit to other icons like the playbacktogglebutton

dweinber: If you think we should use something more similar to the hugereplaybutton I'm happy to talk to our designers for that.

src/ts/components/replaybutton.ts
dweinber: Fair point, I missed this case indeed.

src/scss/skin-modern/components/_replaybutton.scss
offnertho: I was just asking because I did not see this `svg` added in the PR. 

But I trust your (and design) judgement on this one. 

As long as the points you mentioned are fulfilled it is fine.

src/ts/main.ts
```diff
@@ -79,6 +79,7 @@ export { SettingsPanelPageBackButton } from './components/settingspanelpagebackb
 export { SettingsPanelPageOpenButton } from './components/settingspanelpageopenbutton';
 export { SubtitleSettingsPanelPage } from './components/subtitlesettings/subtitlesettingspanelpage';
 export { SettingsPanelItem } from './components/settingspanelitem';
+export { ReplayButton } from './components/replaybutton';
```
offnertho, line 82: general question, 

so it would be possible not to `enable` this button, or is it enabled per default? 

Is there some config, or is this done by code?

dweinber: If you mean our default UI: it's not included at all at this point. We merely add and expose the component so that customized UIs can use it easily.

src/ts/components/replaybutton.ts
dweinber: Updated in 5d8b0c47

src/scss/skin-modern/components/_replaybutton.scss
dweinber: Added new icon in cb24b06c.
Preview: 
image

offnertho: Thanks! Looks fitting! 

I personally would not place it beside the play-button to avoid hitting the replay when wanting to pause the content, but I guess that is up to the customization then. 

Icon looks fine.

src/ts/components/replaybutton.ts
offnertho: Thanks! 

Just wondering, what happens in a live stream? Should it be possible for this component to even show?

src/scss/skin-modern/components/_replaybutton.scss
dweinber: Yes, placement of the button is up to the customer of course, this is just an example to show in the PR how it looks like.

offnertho (approved):
LGTM.

src/ts/components/replaybutton.ts
dweinber: Another very good question! I think there are cases where it's useful for live streams as well, but that's a slightly more complex topic, let's save that for a future PR. Button is now hidden during live streams: 0e5bae25

PR #495, opened 2022-12-01
Add SmartTV UI with Spatial Navigation

dweinber:
- Adds support for Spatial Navigation inlcuding tests (reviewed as part of https://github.com/bitmovin/bitmovin-player-ui/pull/468, https://github.com/bitmovin/bitmovin-player-ui/pull/472, https://github.com/bitmovin/bitmovin-player-ui/pull/475, https://github.com/bitmovin/bitmovin-player-ui/pull/476) - thanks to @mcarriga @TeslaAdis @HalfbyteHeroes @mukulmishra18 (and whoever I forgot here)
- Adds a default SmartTV UI variant, which is the main thing left to review, i.e.
  - `src/scss/skin-modern/_skin-tv.scss`
  - `src/scss/skin-modern/_skin.scss`
  - `src/ts/uifactory.ts`

- Remove several SASS linting warnings


src/scss/skin-modern/_skin-tv.scss
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  .#{$prefix}-ui-titlebar-top {
+    margin-bottom: 1vh;
+
+    > .#{$prefix}-container-wrapper {
+      width: 100%;
+      font-size: $large-size;
+
+      display: inline-flex;
```
felix-hoc, line 26: There are a bunch of warnings reported by GitHub Actions - I assume linting errors because of wrong sort order. Would be nice to get them fixed.

felix-hoc (commented):
Nice work, all! 🤩 

I had a look at the mentioned files and noticed some linting issues, other than that LGTM 💪🏼 

I don't have a device available so I just had a quick test on desktop, but have not observed any issues. I trust this has been properly tested on TVs anyway.

src/scss/skin-modern/_skin-tv.scss
dweinber: There always have been quite some SASS linting WARNINGS for a long time, GH Actions just makes some of them more prominent now 😉 
I fixed quite a lot of them in the last few commits, a few are still there (they aren't as easy to fix - if we resolve the nesting ones in a straight forward way, we seem to exchange that warning with a nesting depth warning. Might need more care and testing). I'd like to keep that out of this PR if possible.

felix-hoc: Ok, I thought since we're adding a new file here, it would make sense to already do it "properly" :)
Thanks for fixing them!

We can also think about adjusting linting rules if we feel like they don't make sense for us.

felix-hoc (approved):

src/scss/skin-modern/_skin-tv.scss
dweinber: Totally get it and usually agree.

I'm not in the position to recommend changing linting rules, but maybe you can discuss internally and let me know the outcome, happy to invest some time later on if I can.

PR #499, opened 2022-12-01
Upgrade dependencies

dweinber:
After merging all the dependabot PR's, the build got broken.
Fixed by:
- Manually upgrading `typescript` to 4.3
- Fixing a newly discovered "type error" in a unit test (it is like that on purpose)
- Manually installing jest-environment-jsdom
- Adding browser API polyfill for JSDOM
- Upgrading nodejs from 10 to 16


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '10'
+  - '16'
```
felix-hoc, line 3: please also add an `.nvmrc` file that contains this version :)
GitHub Actions can then directly [read the configured version from there](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file), same as users' `nvm` installation.
Not sure about travis.

setup-jest.js
```diff
@@ -0,0 +1 @@
+require('@inrupt/jest-jsdom-polyfills');
```
felix-hoc, line 1: please make this into a TypeScript file and use `import`.

spec/errorutils.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { MobileV3PlayerErrorEvent } from '../src/ts/mobilev3playerapi';
 import { ErrorUtils } from '../src/ts/errorutils';
 import defaultMobileV3ErrorMessageTranslator = ErrorUtils.defaultMobileV3ErrorMessageTranslator;
-import { ErrorEvent } from 'bitmovin-player';
+import { ErrorEvent, PlayerEvent } from 'bitmovin-player';
```
felix-hoc, line 4: why do we need this `PlayerEvent` import? I don't see it used anywhere.

felix-hoc (changes requested):
There is a problem with rendering the control buttons:
image

The elements themselves are present and work, they are just not visible. This does not happen with the current `develop` branch.

Also - these are potentially pretty significant changes (esp. switching to Node 16). Please add a changelog entry so uses are aware :)

.travis.yml
dweinber: This will be part of https://github.com/bitmovin/bitmovin-player-ui/pull/500 :) I can add here too if you like though

setup-jest.js
dweinber: Done in 7d1cffa8

spec/errorutils.spec.ts
dweinber: We don't, unused import. This was a left over from my tries to properly fix this 😅 
Fixed in 21bd209d

.travis.yml
dweinber: Added in 3a62a045

dweinber:
Thanks for the review. I added Changelog entries for the NodeJS and Typescript upgrades in 72c0d78e.

I'm still looking into the icon issue (good catch btw, thanks!), looks like the processing/importing of SVG images changes in the tools, leading to invalid CSS values for the SVG images.

dweinber:
SVG icons work again after downgrading `postcss-svg` npm package in 5f29c3ca.

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
```
felix-hoc, line 7: this should be
```suggestion
## [develop]
```
(required for our release process)

package.json
```diff
@@ -18,39 +18,46 @@
...
-    "gulp-postcss": "^6.2.0",
+    "gulp-postcss": "^9.0.1",
     "gulp-rename": "^1.4.0",
     "gulp-replace": "^0.6.1",
     "gulp-sass": "^5.1.0",
     "gulp-sass-lint": "^1.4.0",
     "gulp-sourcemaps": "^2.6.4",
     "gulp-tslint": "^8.1.3",
-    "gulp-typescript": "^3.2.4",
+    "gulp-typescript": "^6.0.0-alpha.1",
```
felix-hoc, line 46: Oof, should we really use an `alpha` version here? 😅

CHANGELOG.md
dweinber: I was just following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) as the top of the file recommends 😉

package.json
dweinber: Dependabot thought it's a great idea. Who am I to not trust it? 😅

dweinber: `6.0.0-alpha.1` was published 3 years ago and is the last publish of `gulp-typescript` as it seems: https://www.npmjs.com/package/gulp-typescript
I can try downgrading to 5.x if you feel safer with that?

dweinber: Done in 7f709a9b

felix-hoc: dependabot is not always to be trusted 😅 
yeah if it doesn't make a difference, I'd stay with a stable release :)

dweinber: Apparently 😅 
Agreed, downgraded to the latest stable 5.x

package.json
```diff
@@ -18,39 +18,46 @@
...
     "merge2": "^1.2.2",
     "postcss-svg": "^1.0.6",
     "sass": "^1.49.8",
     "stream-combiner2": "^1.1.1",
-    "ts-jest": "^24.0.2",
+    "ts-jest": "^29.0.3",
     "tsify": "^2.0.2",
     "tslint": "^5.11.0",
-    "typescript": "^3.9.6",
+    "typescript": "4.3",
```
felix-hoc, line 57: nit: Please also use a proper semver here. It should be `^4.3.5`
(and run `npm install` to apply it to the `package-lock.json` file as well)

felix-hoc (approved):
just one more nitpick, otherwise LGTM. Nice upgrades! ✌🏼 

I had a quick test with a custom build, but it would be good to do some more extensive testing, if you haven't done so already 🙏🏼

package.json
dweinber: Makes sense, done in ad046897.

dweinber:
I did quite some testing after you raised the icon problem, but can't of course guarantee that there are no issues

PR #500, opened 2022-12-01
Move to GitHub Actions

dweinber:
Moving from Travis CI to Github Actions to speed builds up and get rid of external dependency.

From push to finished in ~1m:
image


.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
+name: CI
+
+on: 
+  push:
+    branches:
+      - '**'
```
offnertho, line 6: What does this do? 

Can't you just leave it out as it is the default I guess?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+
+  # Allows you to run this workflow manually from the Actions tab
```
offnertho, line 8: sure we can leave the comment, but I guess once you are familiar with GH actions this becomes kind of obsolete.

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+          node-version-file: .nvmrc
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Test
+        run: npm test
+
+      - name: Build
+        run: gulp build-prod
```
offnertho, line 33: You can also `reuse` other workflows: 

https://docs.github.com/de/actions/using-workflows/reusing-workflows

Because this is copied. So what would be my suggestion is to use `upload` and `download` artifact to pass over the build artifact to the release artifact.

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Publish
+        run: bash publish.sh
+        env:
+          NPM_DRY_RUN: false
```
offnertho, line 38: if this is hardcoded, why do we provide the default value?

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: npm ci
+
+      - name: Test
+        run: npm test
+
+      - name: Build
+        run: gulp build-prod
+
+      - name: Publish
+        run: bash publish.sh
```
offnertho, line 36: I guess we don't need the `bash` right?

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
+name: Release
+
+on: 
+  push:
+    tags:
+      - '!player/'
+      - 'v*'
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
offnertho, line 10: If you start it manually (which is good IMHO) what version does it release?

offnertho (changes requested):
Generally quite nice and simple. 

Personally I'd prefer to move away from `bash` scripts and use node scripts as they can be executed quite conveniently in gh-actions. 

Appart from that looks quite good already.

.github/workflows/ci.yml
dweinber: Agreed, removed in 419033e5

dweinber:
I agree, nodejs scripts would be much nicer than bash. However, the focus of this PR is to move from TravisCI to Github Actions. Therefore, I don't plan to touch any of the existing scripts and move the workflow over with as little change as possible. I think such improvements should be done in separate PR(s).

.github/workflows/release.yml
dweinber: Because the `NPM_DRY_RUN` environment variable is used in the `publish.sh` script, and that env variable with the hardcoded value is also configured in TravisCI.

.github/workflows/release.yml
dweinber: We could get rid of that and add the `shell` type in a separate parameter as shown in https://docs.github.com/en/actions/learn-github-actions/essential-features-of-github-actions#adding-scripts-to-your-workflow:

```yaml
jobs:
  example-job:
    steps:
      - name: Run build script
        run: ./.github/scripts/build.sh
        shell: bash
```
Not sure this is a huge improvement to the initial code, but changed it in 6bc9b515.

.github/workflows/release.yml
dweinber: This was a copy & paste mistake. As you aren't able to create manual releases (without a tag) today, I removed this option as well in 419033e5

.github/workflows/ci.yml
dweinber: The default for [push](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push) is:
> Runs your workflow when you push a commit or tag.

As we have a separate workflow for tags, I added the `branches` section. It accepts standard glob patterns, which this is. But you might be right, it could work with `branches` but without any list entries (but it's essentially equivalent).

dweinber: Just noticed that it's not the case, GH Action isn't triggered for this branch anymore with `branches` but no glob:
 
image

.github/workflows/release.yml
dweinber: Agreed, updated (over several commits)

dweinber:
Code is now finally ready for re-review @offnertho.
I still need to add the NPM_AUTH_TOKEN as secret to the repo, will do this next week (but no code change needed after that).

dweinber:
`NPM_AUTH_TOKEN` was added as secret.

.github/workflows/release.yml
offnertho: How can you enable it?

offnertho (approved):
Looks good. 

So this could already be used as is?

.github/workflows/ci.yml
dweinber: Yes, but then it is also triggered for tags. As tags are covered by a separate workflow, I'd suggest leaving the `branches` in. Or why do you see this as such an issue?

.github/workflows/release.yml
dweinber: It's meant for local testing of the script, as far as I can tell. Again, I didn't write, add, or change that script, I just ported it over.

dweinber:
> So this could already be used as is?
Yes.

.github/workflows/ci.yml
dweinber: Discussed offline, fine to keep as is.

PR #503, opened 2022-12-05
Reduce CI build time

dweinber:
Reduce CI & Release workflows run time:
- CI workflow by ~30 sec by 
  - storing only the built files as the artifact (thanks @offnertho for the idea)
  - putting the artifact files into a tarball
- Release workflow by ~15-20 sec by
  - checkout out the repo freshly instead of downloading all files as the artifact
  - downloading only the built files as tarball artifact


felix-hoc (approved):
Makes sense 👍🏼

PR #521, opened 2023-01-31
Add configuration to enter fullscreen on initial play

dweinber:
Inspired by https://community.bitmovin.com/t/setting-full-screen-with-the-configuration-in-javascript/1658/6 this PR adds a configuration to `PlaybackToggleButton`, `HugePlaybackToggleButton` and `PlaybackToggleOverlay` to enter fullscreen on the first attempt to initiate playback via one of those components.

The default value is set to `false`, so no behavior is changed automatically when users update the UI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
+
+## [unreleased]
+
+### Added
+- Configuration option to enter fullscreen when clicking `PlaybackToggleButton` or `HugePlaybackToggleButton` to initiate the initial playback
+
 ## [3.42.0]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [unreleased]
 
 ### Added
-- Configuration option to enter fullscreen when clicking `PlaybackToggleButton` or `HugePlaybackToggleButton` to initiate the initial playback
+- `UIConfig.enterFullscreenOnInitialPlayback` to enter fullscreen when clicking `PlaybackToggleButton` or `HugePlaybackToggleButton` to initiate the initial playback
 
 ## [3.42.0]
 
--- a/src/ts/components/playbacktogglebutton.ts
+++ b/src/ts/components/playbacktogglebutton.ts
@@ -38,6 +38,7 @@ export class PlaybackToggleButton extends ToggleButton<PlaybackToggleButtonConfi
     super.configure(player, uimanager);
 
     let isSeeking = false;
+    let firstPlay = true;
 
     // Handler to update button state based on player state
     let playbackStateHandler = () => {
@@ -57,6 +58,7 @@ export class PlaybackToggleButton extends ToggleButton<PlaybackToggleButtonConfi
     // Call handler upon these events
     player.on(player.exports.PlayerEvent.Play, (e) => {
       this.isPlayInitiated = true;
+      firstPlay = false;
       playbackStateHandler();
     });
 
@@ -82,6 +84,7 @@ export class PlaybackToggleButton extends ToggleButton<PlaybackToggleButtonConfi
     player.on(player.exports.PlayerEvent.Warning, (event: WarningEvent) => {
       if (event.code === player.exports.WarningCode.PLAYBACK_COULD_NOT_BE_STARTED) {
         this.isPlayInitiated = false;
+        firstPlay = true;
         this.off();
       }
     });
@@ -115,6 +118,10 @@ export class PlaybackToggleButton extends ToggleButton<PlaybackToggleButtonConfi
           player.pause('ui');
         } else {
           player.play('ui');
+
+          if (firstPlay && this.config.enterFullscreenOnInitialPlayback) {
+            player.setViewMode(player.exports.ViewMode.Fullscreen);
+          }
         }
       });
     }
--- a/src/ts/components/playbacktogglebutton.ts
+++ b/src/ts/components/playbacktogglebutton.ts
@@ -4,15 +4,24 @@ import {PlayerUtils} from '../playerutils';
 import { PlayerAPI, WarningEvent } from 'bitmovin-player';
 import { i18n } from '../localization/i18n';
 
+export interface PlaybackToggleButtonConfig extends ToggleButtonConfig {
+  /**
+   * Specify whether the player should be set to enter fullscreen by clicking on the playback toggle button
+   * when initiating the initial playback.
+   * Default is false.
+   */
+  enterFullscreenOnInitialPlayback?: boolean;
+}
+
 /**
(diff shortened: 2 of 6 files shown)


src/ts/uifactory.ts
```diff
@@ -308,7 +308,7 @@ export namespace UIFactory {
       components: [
         new BufferingOverlay(),
         new AdClickOverlay(),
-        new PlaybackToggleOverlay(),
+        new PlaybackToggleOverlay({ enterFullscreenOnInitialPlayback: false }),
```
offnertho, line 311: this is duplicated a couple of times... 

can we extract this into a function to be reusable?

src/ts/uifactory.ts
```diff
@@ -129,7 +129,7 @@ export namespace UIFactory {
         }),
         new Container({
           components: [
-            new PlaybackToggleButton(),
+            new PlaybackToggleButton({ enterFullscreenOnInitialPlayback: false }),
```
offnertho, line 132: why are you not using the config here?

offnertho (changes requested):
Overall makes sense,

Just 2 questions/remarks.

src/ts/uifactory.ts
dweinber: Not sure I understand your question. I'm adding the component config here to the component? 🤔

dweinber:
@offnertho Please have a look again, I removed the setting of the individual component configurations in the `uifactory.ts` and instead support now (in addition) the use the `UIConfig.enterFullscreenOnInitialPlayback`. This makes it more powerful as users can set that through the player configuration (`playerConfig.ui.enterFullscreenOnInitialPlayback = true`) without the need of externally managing the UI instance.

src/ts/components/playbacktogglebutton.ts
```diff
@@ -48,6 +63,7 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
     // Call handler upon these events
     player.on(player.exports.PlayerEvent.Play, (e) => {
       this.isPlayInitiated = true;
```
offnertho, line 65: could we re-use this flag as opposed to introduce a new one?

src/ts/components/playbacktogglebutton.ts
```diff
@@ -28,7 +37,13 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
   configure(player: PlayerAPI, uimanager: UIInstanceManager, handleClickEvent: boolean = true): void {
     super.configure(player, uimanager);
 
+    // Set enterFullscreenOnInitialPlayback if set in the uimanager config
+    if (typeof uimanager.getConfig().enterFullscreenOnInitialPlayback === 'boolean') {
+      this.config.enterFullscreenOnInitialPlayback = uimanager.getConfig().enterFullscreenOnInitialPlayback;
```
offnertho, line 42: do we need the extra config now, or can we just always use the `uimanager.getConfig().enterFullscreenOnInitialPlayback`

src/ts/components/playbacktogglebutton.ts
```diff
@@ -73,6 +89,7 @@ export class PlaybackToggleButton extends ToggleButton<ToggleButtonConfig> {
     player.on(player.exports.PlayerEvent.Warning, (event: WarningEvent) => {
       if (event.code === player.exports.WarningCode.PLAYBACK_COULD_NOT_BE_STARTED) {
         this.isPlayInitiated = false;
+        firstPlay = true;
```
offnertho, line 92: why do we need to re-set here? isn't the player in fullscreen already?

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
+
+## [unreleased]
+
+### Added
+- `UIConfig.enterFullscreenOnInitialPlayback` to enter fullscreen when clicking `PlaybackToggleButton` or `HugePlaybackToggleButton` to initiate the initial playback
```
offnertho, line 10: what happens if the user has autoplay enabled, the flag `enterFullscreenOninitalPlayback` set to true, then pauses and plays again. Will it enter fullscreen?

src/ts/components/playbacktogglebutton.ts
dweinber: With the extra config, it can still be configured per component if a customer creates their own (forked or externally configured/managed UI). So not strictly necessary but provides more flexibility.

CHANGELOG.md
dweinber: No, because the `firstPlay` flag is changed in the `Play` event listener: https://github.com/bitmovin/bitmovin-player-ui/blob/d3f7714a4224cdd05955d2277f9cbb76ff7cc844/src/ts/components/playbacktogglebutton.ts#L64-L66

src/ts/components/playbacktogglebutton.ts
dweinber: No we can't. This flag is e.g. reset on `Paused` or `Playing`, which would lead to undesired side effects. For example, the user exits fullscreen, pauses the player (the flag is reset) and when resuming again, fullscreen would be triggered again.

src/ts/components/playbacktogglebutton.ts
dweinber: This warning should only happen with autoplay. By having this here, we enable the enter fullscreen when the user _actually_, manually starts the initial playback (because autoplay was blocked).

If I remove this, the `firstPlay` would be set to `false` in the `Play` event which is also triggered with autoplay.

I see that both options are somewhat reasonable. Keep in mind that in order to remove this also from the `hugeplaybacktogglebutton.ts`, we'd need to introduce yet another flag because the `firstPlay` there was in place before and is also used for other stuff, so we can't just remove that there.

CHANGELOG.md
offnertho: 👍

src/ts/components/playbacktogglebutton.ts
offnertho: 👍

src/ts/components/playbacktogglebutton.ts
offnertho: Fine as is.

src/ts/uiconfig.ts
```diff
@@ -90,4 +90,10 @@ export interface UIConfig {
    * Default: true
    */
   enableSeekPreview?: boolean;
+  /**
+   * Specifies if the player should be set to fullscreen by clicking on the `PlaybackToggleButton` or
+   * `HugePlaybackToggleButton` when attempting the initial playback start.
```
offnertho, line 95: Pls also mention the overlay

offnertho: [nit] maybe `Specifies if the player should enter fullscreen by...`

offnertho (approved):
LGTM.

src/ts/uiconfig.ts
dweinber: Updated in 94bbd241eb4118f9432bda2c2320fa4999116aa8

PR #522, opened 2023-02-02
fix: remove circular dep on AudioTrackListBox

eddiemonge:
`main` was referencing `uifactory`, which then imported back from `main`.

Changing to remove this circular dependency


dweinber (approved):
Makes sense! Thanks for the contribution.

PR #538, opened 2023-03-10
Force overflowing subtitles back into container

Andr3wid:
Certain constellations of VTT properties (see bottom of description for example) can cause the subtitle labels to overflow their container in the UI. Even though the UI handling in this case is spec-conform and expected, it would be convenience-feature to force such subtitles back into the respective container so they are fully visible. 

Could be helpful if users of the UI don't control the source/format of their VTT files or the results of the VTT parser.

For that matter, a config option `forceCroppedSubtitlesIntoViewContainer` is introduced with this PR. It is not `true` by default, as it breaks spec conformity by overriding certain VTT properties.

**Display with** `forceCroppedSubtitlesIntoViewContainer = false`
Screenshot 2023-03-10 at 09 07 12

**Display with** `forceCroppedSubtitlesIntoViewContainer = true`
Screenshot 2023-03-10 at 09 07 26

**Sample VTT file**
```text
WEBVTT

1
00:01.000 --> 00:30.000 position:80%,line-left line:40% align:start
right side cropped A B C D E F G H I K L M N O P Q R S T V X Y Z

2
00:01.000 --> 00:30.000 position:20%,line-right line:60% align:end
A B C D E F G H I K L M N O P Q R S T V X Y Z left side cropped

3
00:01.000 --> 00:30.000 line:5%,end
-----line 1-----
-----line 2-----
top side cropped
-----line 4-----

4
00:01.000 --> 00:30.000 line:95%,start
-----line 1-----
bottom side cropped
-----line 3-----
-----line 4-----
```

### DIFF
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -10,6 +10,13 @@ import { i18n } from '../localization/i18n';
 import { VttUtils } from '../vttutils';
 import { VTTProperties } from 'bitmovin-player/types/subtitles/vtt/API';
 
+interface SubtitleCropDetectionResult {
+  top: boolean;
+  right: boolean;
+  bottom: boolean;
+  left: boolean;
+}
+
 /**
  * Overlays the player to display subtitles.
  */
@@ -128,6 +135,49 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
     this.configureCea608Captions(player, uimanager);
     // Init
     subtitleClearHandler();
+  }
+
+  detectCroppedSubtitleLabel(
+    labelElement: HTMLElement,
+  ): SubtitleCropDetectionResult {
+    const parent = this.getDomElement().get(0);
+
+    const childRect = labelElement.getBoundingClientRect();
+    const parentRect = parent.getBoundingClientRect();
+
+    return {
+      top: childRect.top < parentRect.top,
+      right: childRect.right > parentRect.right,
+      bottom: childRect.bottom > parentRect.bottom,
+      left: childRect.left < parentRect.left,
+    };
+  }
+
+  handleSubtitleCropping(label: SubtitleLabel) {
+    const labelDomElement = label.getDomElement();
+    const cropDetection = this.detectCroppedSubtitleLabel(
+      labelDomElement.get(0),
+    );
+
+    if (cropDetection.top) {
+      labelDomElement.css('top', '0');
+      labelDomElement.css('bottom', 'unset');
+    }
+
+    if (cropDetection.right) {
+      labelDomElement.css('right', '0');
+      labelDomElement.css('left', 'unset');
+    }
+
+    if (cropDetection.bottom) {
+      labelDomElement.css('bottom', '0');
+      labelDomElement.css('top', 'unset');
+    }
+
+    if (cropDetection.left) {
+      labelDomElement.css('left', '0');
+      labelDomElement.css('right', 'unset');
+    }
   }
 
   generateLabel(event: SubtitleCueEvent): SubtitleLabel {
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -74,6 +74,10 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
 
       this.subtitleContainerManager.addLabel(label, this.getDomElement().size());
       this.updateComponents();
+
+      if (uimanager.getConfig().forceSubtitlesIntoViewContainer) {
+        this.handleSubtitleCropping(label);
+      }
     });
 
     player.on(player.exports.PlayerEvent.CueUpdate, (event: SubtitleCueEvent) => {
@@ -84,6 +88,10 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
 
       if (labelToReplace) {
         this.subtitleContainerManager.replaceLabel(labelToReplace, label);
+      }
+
+      if (uimanager.getConfig().forceSubtitlesIntoViewContainer) {
+        this.handleSubtitleCropping(label);
       }
     });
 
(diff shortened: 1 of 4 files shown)


src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -71,3 +71,23 @@
...
+}
+
+.#{$prefix}-right-cropped-subtitle {
+  right: 0 !important;
+  left: auto !important;
+}
+
+.#{$prefix}-bottom-cropped-subtitle {
+  bottom: 0 !important;
+  top: auto !important;
```
Andr3wid, line 87: Check if we could `unset` this or omit it completely.

src/ts/uiconfig.ts
```diff
@@ -96,4 +96,9 @@ export interface UIConfig {
    * Default: false
    */
   enterFullscreenOnInitialPlayback?: boolean;
+
+  /**
+   * Forces subtitle-labels back into their respective container if they overflow and are therefore cropped.
+   */
+  forceCroppedSubtitlesIntoViewContainer?: boolean;
```
Andr3wid, line 103: Check if needs to be added to some public doc.

src/scss/skin-modern/components/_subtitleoverlay.scss
```diff
@@ -71,3 +71,23 @@
     }
   }
 }
+
+.#{$prefix}-top-cropped-subtitle {
+  top: 0 !important;
```
Andr3wid, line 76: Make this alphabetically ordered

Andr3wid: Obsolete - using inline css instead of classes now:
[Use inline css instead of classes for cropped captions](https://github.com/bitmovin/bitmovin-player-ui/pull/538/commits/56a168378e3bc33a86282816f5ef17d73c4a3653)

src/scss/skin-modern/components/_subtitleoverlay.scss
Andr3wid: Using `unset` instead of `auto` as a safety measure.

src/ts/components/subtitleoverlay.ts
```diff
@@ -130,6 +145,49 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+
+  handleSubtitleCropping(label: SubtitleLabel) {
+    const labelDomElement = label.getDomElement();
+    const cropDetection = this.detectCroppedSubtitleLabel(
+      labelDomElement.get(0),
+    );
+
+    if (cropDetection.top) {
+      labelDomElement.css('top', '0');
+      labelDomElement.css('bottom', 'unset');
```
felix-hoc, line 172: Instead of setting this explicitly to `unset` - could we just remove it from the inline styles? Otherwise I'm not sure if external styling would still work, as the inline `unset` would override everything.

I guess we'd need another function in our DOM.ts but it should be easy enough.

felix-hoc (commented):
nice, just 1 more comment regaring `unset`

src/ts/components/subtitleoverlay.ts
Andr3wid: Good call! Implemented it as suggested:

[Remove inline css instead of overriding for caption overflows](https://github.com/bitmovin/bitmovin-player-ui/pull/538/commits/bc723697fed4937db120a334b86175ffe65c3f39)

src/ts/dom.ts
```diff
@@ -562,6 +562,14 @@ export class DOM {
     }
   }
 
+  /**
+   * Removes an inline CSS property if it exists
+   * @param propertyName name of the property to remove
+   */
+  removeInlineCss(propertyName: string): void {
```
felix-hoc, line 569: nit: this could just be `removeCss` - the fact that it gets called on a component already implies that it is about inline css :)

src/ts/dom.ts
```diff
@@ -562,6 +562,14 @@ export class DOM {
...
+    this.setCss(propertyName, null);
```
felix-hoc, line 570: Doesn't this leave an empty entry in the `style` attribute? Like `style="background:null"`?

There is also a function for explicitly this purpose: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/removeProperty

felix-hoc (commented):
just some nitpicks

src/ts/dom.ts
Andr3wid: Rename: [Rename function to removeCss](https://github.com/bitmovin/bitmovin-player-ui/pull/538/commits/7b8b96eb9b08aea85acd17ced6010cf8f34ac633)

src/ts/uiconfig.ts
```diff
@@ -96,4 +96,8 @@ export interface UIConfig {
    * Default: false
    */
   enterFullscreenOnInitialPlayback?: boolean;
+  /**
+  * Forces subtitle-labels back into their respective container if they overflow and are therefore cropped.
+  */
+  forceSubtitlesIntoViewContainer?: boolean;
```
stonko1994, line 102: There seems to be an indentation problem:

```suggestion
  /**
   * Forces subtitle-labels back into their respective container if they overflow and are therefore cropped.
   */
  forceSubtitlesIntoViewContainer?: boolean;
```

stonko1994 (commented):
Looks good to me 💪 Also manually testing works fine.

Can you please add a Changelog entry as well 🙂

Andr3wid:
Changelog updated:
[Add changelog entry](https://github.com/bitmovin/bitmovin-player-ui/pull/538/commits/cb143bdc461fcf5d67273152ba4a9c992f4906eb)

felix-hoc (approved):
please fix the changelog section, otherwise LGTM 🙌🏼

stonko1994 (approved):

PR #561, opened 2023-07-17
Fix volume restoration when changing it via the API

felix-hoc:
When we call the player API directly, such as `player.setVolume(20)` in the browser console, the volume data was not stored in UI. Thus, when trying to restore volume data, (e.g. when unmuting) the value did not get applied.

### Changes
- Store volume on VolumeChanged event
- Unit test for storing the volume


ravloony (approved):
LGTM

PR #563, opened 2023-07-20
Fix settings panel closing issue in the small screen UI (mobile devices)

kishorens:
Problem: The settings panel requires a double touch to close when the control bar on mobile devices auto-hides (small screen UI). 

Fix: Added change to cancel the auto-hide of control bar when the settings panel is active on mobile devices. Once the user interacts with the Settings panel and closes(x) it, the auto-hide will work as usual.

### DIFF
--- a/src/ts/components/controlbar.ts
+++ b/src/ts/components/controlbar.ts
@@ -81,7 +81,7 @@ export class ControlBar extends Container<ControlBarConfig> {
 
     uimanager.onPreviewControlsHide.subscribe((sender, args) => {
       // Cancel the hide event if hovered child components block hiding or if the settings panel is active on mobile.
-      args.cancel = (hoverStackCount > 0 || isSettingsPanelShown);
+      args.cancel = args.cancel || (hoverStackCount > 0 || isSettingsPanelShown);
     });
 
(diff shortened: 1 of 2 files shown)


src/ts/components/controlbar.ts
```diff
@@ -58,13 +61,27 @@ export class ControlBar extends Container<ControlBarConfig> {
...
+
     uimanager.onControlsShow.subscribe(() => {
       this.show();
     });
 
     uimanager.onPreviewControlsHide.subscribe((sender, args) => {
-      // Cancel the hide event if hovered child components block hiding
-      args.cancel = (hoverStackCount > 0);
+      // Cancel the hide event if hovered child components block hiding or if the settings panel is active on mobile.
+      args.cancel = (hoverStackCount > 0 || isSettingsPanelShown);
```
oberhamsi, line 84: mabye do

` args.cancel = args.cancel || (hoverStackCount > 0 || isSettingsPanelShown);`

otherwise components subscribed to this earlier can't cancel.

kishorens: Agree, thanks for the suggestion.

felix-hoc (approved):
Looks good 👍🏼

PR #564, opened 2023-07-21
Fix watch task

felix-hoc:
The `gulp watch` task is not working.

This PR fixes it and aligns the flow with the `serve` task.


bitmovin-kenny (approved):
Didn't work before, works now 👍

PR #572, opened 2023-08-30
Displaying emissons with Eco Mode activated

mariacrdribeiro:
This feature enables users to visualize the number of emissions saved through the activation of the eco mode

Added Features:

- Creating a container for the Eco Mode
- Calculating and Displaying the saved emissons


src/scss/skin-modern/components/_ecomodetogglebutton.scss
```diff
@@ -14,12 +14,14 @@
...
     background-position: 20px center;
     background-size: 45% auto;
+    margin-left: 2%;
   }
 
   &.#{$prefix}-off {
     background-image: url('../../assets/skin-modern/images/toggleOff.svg');
     background-position: 20px center;
     background-size: 45% auto;
+    // margin-left: 2%;
```
jmsn, line 24: Should this be removed?

src/scss/skin-modern/components/_label.scss
```diff
@@ -9,6 +9,10 @@
   white-space: nowrap;
 }
 
-.#{$prefix}-ui-label {
+
+.#{$prefix}-ui-label-savedEnergy {
```
jmsn, line 13: We are changing the existing style for `.#{$prefix}-ui-label` so the style for existing labels style changes. Should we not instead define a new one?

src/ts/components/ecomodecontainer.ts
```diff
@@ -0,0 +1,107 @@
...
+      if (this.ecoModeSavedEnergyItem.isActive()) {
+        this.energySaved(currentEnergyKwh, maxEnergyKwh, this.energySavedLabel);
+      } else {
+        this.savedEnergy = 0;
+        this.energySavedLabel.setText(this.savedEnergy.toFixed(4) + ' gCO2/kWh');
+      }
+    });
+  }
+
+  energySaved(currentEnergyConsuption: number, maxEnergyConsuption: number, energySavedLabel: Label<LabelConfig>) {
```
jmsn, line 74: Can we have a better name for this function, like `updateSavedEnergy`?

src/ts/components/ecomodecontainer.ts
```diff
@@ -0,0 +1,107 @@
...
+    this.currentEnergyEmission = currentEnergyConsuption * 475; // 475 is the average country intensity of all countries in gCO2/kWh
+    this.maxEnergyEmisson = maxEnergyConsuption * 475;
+
+    if (!isNaN(this.currentEnergyEmission) && !isNaN(this.maxEnergyEmisson)) {
+      this.savedEnergy += this.maxEnergyEmisson - this.currentEnergyEmission;
```
jmsn, line 79: `savedEnergy` is actually saved emissions here, so the name is wrong

src/ts/components/ecomodecontainer.ts
```diff
@@ -0,0 +1,107 @@
...
+      energySavedLabel.setText(this.savedEnergy.toFixed(4) + ' gCO2/kWh');
```
jmsn, line 80: Same here `energySavedLabel` is displaying the saved emissions, so the name should reflect that

src/ts/components/ecomodecontainer.ts
```diff
@@ -0,0 +1,107 @@
...
+
+      const currentEnergyKwh = this.calculateEnergyConsumption(frameRate, height, width, bitrate);
+
+      const maxEnergyKwh = this.calculateEnergyConsumption(frameRate, maxHeight, maxWidth, maxBitrate);
+
+      if (this.ecoModeSavedEnergyItem.isActive()) {
+        this.energySaved(currentEnergyKwh, maxEnergyKwh, this.energySavedLabel);
+      } else {
+        this.savedEnergy = 0;
+        this.energySavedLabel.setText(this.savedEnergy.toFixed(4) + ' gCO2/kWh');
```
jmsn, line 69: Question: So disabling the eco mode resets the saved energy and when enabling it again it would start from 0, right? We should think if that is the correct behavior.

src/ts/components/ecomodecontainer.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const energyConsumptionW =
+      fpsWeight * fps +
+      pixeldWeight * height * width +
+      (birateWeight + bitrateInternetWeight) * (bitrate / 1000) +
+      videoCodec +
+      constantOffset +
+      internetConnectionOffset;
+
+    const energyConsumptionKwh = energyConsumptionW / 3.6e6;
```
jmsn, line 103: Can we add an explanatory comment for this caclulation, please?

src/ts/components/ecomodecontainer.ts
```diff
@@ -0,0 +1,107 @@
...
+      const maxHeight = player.getAvailableVideoQualities()[maxQualityAvailable].height;
+      const maxBitrate = player.getAvailableVideoQualities()[maxQualityAvailable].bitrate;
+      const maxWidth = player.getAvailableVideoQualities()[maxQualityAvailable].width;
+
+      const currentEnergyKwh = this.calculateEnergyConsumption(frameRate, height, width, bitrate);
+
+      const maxEnergyKwh = this.calculateEnergyConsumption(frameRate, maxHeight, maxWidth, maxBitrate);
+
+      if (this.ecoModeSavedEnergyItem.isActive()) {
+        this.energySaved(currentEnergyKwh, maxEnergyKwh, this.energySavedLabel);
```
jmsn, line 66: We should discuss this. I _think_ instead of using the energy consumption in kWh we might need to take the energy consumption per second and multiply it by the segment duration?

PR #573, opened 2023-09-13
Only remove subtitles which do not encompass the seek target

bitmovin-kenny:
When seeking, we removed all active subtitles, even when we seeking inside the range of the active subtitle. In that case the UI would not display the Cue because the player will only fire a `CueExit` + `CueEnter` when we played outside of the Cue.

So in this PR we check the playback range of all active cues after a Seek / Timeshift operation has been completed and only remove cues which do not enclose the new current time.

### DIFF
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -121,11 +121,18 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
       this.updateComponents();
     };
 
+    const clearInactiveSubtitles = () => {
+      const removedActiveSubtitles = subtitleManager.clearInactiveCues(player.getCurrentTime());
+      removedActiveSubtitles.forEach(toRemove => {
+        this.subtitleContainerManager.removeLabel(toRemove.label);
+      });
+    };
+
     player.on(player.exports.PlayerEvent.AudioChanged, subtitleClearHandler);
     player.on(player.exports.PlayerEvent.SubtitleEnabled, subtitleClearHandler);
     player.on(player.exports.PlayerEvent.SubtitleDisabled, subtitleClearHandler);
-    player.on(player.exports.PlayerEvent.Seek, subtitleClearHandler);
-    player.on(player.exports.PlayerEvent.TimeShift, subtitleClearHandler);
+    player.on(player.exports.PlayerEvent.Seeked, clearInactiveSubtitles);
+    player.on(player.exports.PlayerEvent.TimeShifted, clearInactiveSubtitles);
     player.on(player.exports.PlayerEvent.PlaybackFinished, subtitleClearHandler);
     player.on(player.exports.PlayerEvent.SourceUnloaded, subtitleClearHandler);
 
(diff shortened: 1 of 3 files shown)


src/ts/components/subtitleoverlay.ts
```diff
@@ -121,11 +121,18 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
       this.updateComponents();
     };
 
+    const clearInactiveSubtitles = () => {
+      const removedActiveSubtitles = subtitleManager.clearInactiveCues(player.getCurrentTime());
```
jmsn, line 125: NIT: For clarity, I would choose a name that reflects that these are actually `cues`

src/ts/components/subtitleoverlay.ts
```diff
@@ -470,6 +477,24 @@ class ActiveSubtitleManager {
     }
   }
 
+  /**
+   * Removes all active cues which don't enclose the given time
+   * @param time the time for which subtitles should remain
+   */
+  public clearInactiveCues(time: number): ActiveSubtitleCue[] {
+    const removedEvents: ActiveSubtitleCue[] = [];
```
jmsn, line 485: NIT: Ar these cues or events? 😅

src/ts/components/subtitleoverlay.ts
```diff
@@ -470,6 +477,24 @@ class ActiveSubtitleManager {
...
+    Object.keys(this.activeSubtitleCueMap).forEach(key => {
+      const activeCues = this.activeSubtitleCueMap[key];
+      activeCues.forEach(element => {
```
jmsn, line 488: Also here use `cue`? 🙂

src/ts/components/subtitleoverlay.ts
```diff
@@ -470,6 +477,24 @@ class ActiveSubtitleManager {
...
+        if (time < element.event.start || time > element.event.end) {
+          this.popCueFromMap(element.event);
+          removedEvents.push(element);
+        }
+      });
+    });
+    return removedEvents;
+  }
```
jmsn, line 496: So we are preserving only the cue that lies at the given time. Can it happen that there are cues in the future (i.e. that lie after `time`) that should not be cleared? 🤔

jmsn (commented):
LGTM

Just few NITs and a question 🙂

src/ts/components/subtitleoverlay.ts
bitmovin-kenny: It's actually a label and an event:
```ts
interface ActiveSubtitleCue {
  event: SubtitleCueEvent;
  label: SubtitleLabel;
}
```

src/ts/components/subtitleoverlay.ts
bitmovin-kenny: Done in https://github.com/bitmovin/bitmovin-player-ui/pull/573/commits/bf0e75d23a90bb38d251c8c6ece7b18e82d07737

src/ts/components/subtitleoverlay.ts
bitmovin-kenny: I can't think of a reason for there being active cues which should linger until they become active again. I would expect the player to emit `CueEnter` events when it reaches that range. Then we would create an active cue again. If we didn't clear it before, there would then be two cues maybe?

jmsn: Thanks 👍 , I thought so, but wanted to ask if I understood it correctly. 🙂

jmsn (approved):

src/ts/components/subtitleoverlay.ts
bitmovin-kenny: renamed in https://github.com/bitmovin/bitmovin-player-ui/pull/573/commits/10c0774097ff2b736a84630fdbe233fdba8bc81c

PR #579, opened 2023-09-21
[PoC] Spatial Navigation - Arrow key continues closest to mouse position

bitmovin-kenny:
When using the magic remote of LG TVs, which behave like a mouse we need a way to switch from the remote back to buttons for this we

- remove focus from an element when the mouse leaves it
- when an arrow key is pressed the navigation continues from the last position the mouse was at.


HalfbyteHeroes:
This PoC was fully implemented in https://github.com/bitmovin/bitmovin-player-ui/pull/580.

Issue #569, opened 2023-08-01
overlapping issue with seekbar thumbnails


payamtrack:
Hi,

I have increased the size of seekbar thumbnails by using the CSS rules below:
```
.bmpui-ui-seekbar-label {
    width: 30%;
}

.bmpui-ui-seekbar-label .bmpui-seekbar-label-inner > .bmpui-container-wrapper .bmpui-seekbar-thumbnail {
    width: 100% !important;
}
```
However, this causes them to get cut off due to overlapping with the player's edges when hovering over the seek bar up to a specific point at the beginning and end. I am not sure how to address this issue. 

![image](https://github.com/bitmovin/bitmovin-player-ui/assets/135003403/cd4976b9-9f09-4f0f-9fea-9812a9431c74)

As illustrated below, I think the thumbnail container should stop following the mouse pointer when it reaches to the edges, and beyond that point, only the caret-down icon should move with the pointer:

![image](https://github.com/bitmovin/bitmovin-player-ui/assets/135003403/00f52caa-085a-436a-9cdc-297d539ed6fe)

Thanks,

jamesdavidholding:
Hi Payamtrack,

We will be closing the ‘Issues’ tab for this Repo this week. If this issue still affects you, please can I ask that you re-raise it through the Bitmovin Support Dashboard, where we will happily pick it up. 

This change improves our response times to you, as the support team is not monitoring GitHub repro.  

Many thanks,

James - Support

PR #584, opened 2023-11-13
new seekbar option to prevent label overlap

oberhamsi:
potential fix for https://github.com/bitmovin/bitmovin-player-ui/issues/569

### DIFF
--- a/src/scss/skin-modern/components/_seekbarlabel.scss
+++ b/src/scss/skin-modern/components/_seekbarlabel.scss
@@ -76,7 +76,7 @@
   &[style*="transform"] {
     .#{$prefix}-seekbar-label-inner {
       &::after {
-        content: unset;
+        transform: translateX(calc(1px * var(--bmpui-seekbar-label-overflow-offset)));
       }
     }
   }
--- a/src/scss/skin-modern/components/_seekbarlabel.scss
+++ b/src/scss/skin-modern/components/_seekbarlabel.scss
@@ -71,9 +71,9 @@
       }
     }
   }
-  // disable position arrow if seekbar label was
-  // moved due to overflow
-  &[style*="transform"] {
+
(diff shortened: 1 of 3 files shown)


felix-hoc:
Thanks for the submission! :) I tested it a bit - unfortunately, the thumbnails seem to jump around wildly when getting close to the edges, and also the arrow disappears: 
[Screencast.webm](https://github.com/bitmovin/bitmovin-player-ui/assets/16101788/e3439daa-ba99-45d2-bbf0-262576e190ab)

I think we need to find a different solution.

oberhamsi:
wouldn't call it "wildly" ;) it moves inside so it doesn't overlap. any suggestions on a direction to take on this?

felix-hoc:
Yeah I get that :) From a user perspective, it would be nice if the thumbnail would simply stop moving out of the frame, so it stays "attached" to the left/right edge when the mouse moves further out.

I had a quick look but also couldn't come up with a better solution so far :/ we'll need to properly look into it.

oberhamsi:
i think i addressed both your concerns - the label now stays at the edge (with a bit of margin) instead of jumping away from it and the arrow keeps indicating the visible position

[Screencast from 2024-02-28 13:09:31.webm](https://github.com/bitmovin/bitmovin-player-ui/assets/53384/173b6715-4d4d-41ba-8572-ae7cc1ade040)

oberhamsi:
any news on this? if you are worried about performance, i could throttle the overflow-prevention.

felix-hoc:
Sorry for the delay! We have spent some time on it and fixed it here: https://github.com/bitmovin/bitmovin-player-ui/pull/661

We took inspiration from your PR and made some adjustments to hopefully reduce the performance impact. We also made this the default behavior.

oberhamsi:
i tested it with v3.76 and works great for us, cheers!

PR #585, opened 2023-11-13
Fix an exception when receiving a `SourceLoaded` event during scrubbing.

stonko1994:
### Problem
When the `playbackPositionHandler` inside the `Seekbar` is called without an event (i.e., with `null`), such as during the reception of a `SourceLoaded` event, the application crashes. This leads to an undefined UI state where the `seekbar` is partially updated to the player state, and the play button may be missing.

### Changes
Adding the missing `null` check for the optional event parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## [Unreleased]
+## [development]
 
 ### Fixed
 - Crash when receiving a `SourceLoaded` event during scrubbing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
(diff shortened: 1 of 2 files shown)


src/ts/components/seekbar.ts
```diff
@@ -254,7 +254,7 @@ export class SeekBar extends Component<SeekBarConfig> {
       // At the same time when the user is scrubbing, we also move the position of the seekbar to display a preview during scrubbing.
       // When the user is scrubbing we do not record this as a user seek operation, as the user has yet to finish their seek,
       // but we should not move the playback position to not create a jumping behaviour.
-      if (scrubbing && event.type === player.exports.PlayerEvent.SegmentRequestFinished && playbackPositionPercentage !== this.playbackPositionPercentage) {
+      if (scrubbing && event && event.type === player.exports.PlayerEvent.SegmentRequestFinished && playbackPositionPercentage !== this.playbackPositionPercentage) {
```
stonko1994, line 257: The event is optional and even `null` per default. However, it's not checked if it's `null` (This is the only usage of this property here)

wasp898 (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
```
wasp898, line 7: Actually I think you need `development` for the CLI to work.
```suggestion
## [development]
```

PR #588, opened 2023-11-22
Prefix css animations

dweinber:
CSS Animations (`@keyframe`) were not prefixed with the default `bmpui` prefix which is used for all CSS classes and names. This could lead to clashes in applications and websites that use the same name. Especially names like `fade-in` or `fade-out` have a high probability.

This PR applies the prefix also for those animations.


offnertho (commented):
Might there be other places where we `forgot` the prefix?

dweinber:
I haven't seen one in a quick look, and a customer specifically reported issues with the animations.

offnertho (approved):

PR #589, opened 2023-12-04
Use develop instead of Unreleased in changelog

mukulmishra18:
In changelog file `Untracked` is used instead of `develop` for changes merged in develop branch.

This PR fixes that.


felix-hoc (approved):

felix-hoc:
cc @dweinber , as this was added in https://github.com/bitmovin/bitmovin-player-ui/pull/588 :)

dweinber:
FWIW, I was following http://keepachangelog.com which we state at the top of the CHANGELOG we follow 😉 
And already the very top of the page does it like this:
image
See also https://keepachangelog.com/en/1.1.0/#effort

PR #590, opened 2023-12-27
Automate release on merge in develop

mukulmishra18:
We need to automate the release.

This PR automate the release on every PR merge in `develop` branch.


.github/workflows/merge.yml
```diff
@@ -0,0 +1,69 @@
+name: Trigger release on merge
```
offnertho: pls also rename the workflow file.

.github/workflows/merge.yml
```diff
@@ -0,0 +1,69 @@
...
+run-name: Starting release for ${{ github.actor }} PR merge
```
offnertho: what else could we use? but yeah should also be fine.

.github/workflows/merge.yml
```diff
@@ -0,0 +1,69 @@
...
+
+jobs:
+  on_pr_merge:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout develop
+        uses: actions/checkout@v4
+        with:
+          ref: develop
```
offnertho: we don't need that because the workflow is only triggered on develop.

.github/workflows/merge.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Read package.json version
+        uses: actions/github-script@v6
+        id: define-package-json-version
+        with:
+          script: |
+            const { version } = require('./package.json')
+            core.info(`performing a ${{ inputs.release-type }} release for existing version ${version}`)
```
offnertho: I guess we can remove that input right?

.github/workflows/merge.yml
```diff
@@ -0,0 +1,69 @@
...
+            core.setOutput('packageJsonVersion', version)
+
+      - name: Define release version
+        uses: actions/github-script@v6
+        id: define-release-version
+        with:
+          script: |
+            const { defineReleaseVersion } = require('./.github/scripts/defineVersion.js')
+            return defineReleaseVersion({core}, 'minor', undefined, "${{ steps.define-package-json-version.outputs.packageJsonVersion }}" )
```
offnertho: you can also remove the third parameter from the function, so you don't have to pass undefined.

.github/workflows/merge.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Bump package.json version
+        run: |
+          git config --global user.name 'Automated Release'
+          git config --global user.email 'release-automation@bitmovin.com'
+          npm version "${{ fromJson(steps.define-release-version.outputs.result).full }}"
+
+      - name: Update Changelog
```
offnertho: does this step also add a new `development` section?

offnertho (changes requested):
Looks good in general. 

Some small remarks.

.github/workflows/merge.yml
mukulmishra18: Nope. Do we need to add that? I think it should not be required?

.github/workflows/merge.yml
mukulmishra18: I mean we can add the commit hash, but that might make the message too long 😄

.github/scripts/defineVersion.js
```diff
@@ -0,0 +1,36 @@
+const semver = require('semver');
+
+function getPlayerUiVersion(versionInput) {
+  let targetVersion = versionInput ?? process.env.CI_BRANCH;
+
+  if (!targetVersion) {
+    console.log('no version provided using CI_BRANCH');
+    process.exit(1);
+  }
```
offnertho: I think we can remove the CI_BRANCH fallback.

.github/workflows/release-on-merge.yml
```diff
@@ -0,0 +1,64 @@
+name: Trigger release on merge
```
offnertho: Maybe this should be called `tag-release-version`?

.github/workflows/release-on-merge.yml
```diff
@@ -0,0 +1,64 @@
...
+run-name: Starting release for ${{ github.actor }} PR merge
+on:
+  pull_request:
+    types:
+      - closed
+    branches:
+      - develop
+
+jobs:
+  on_pr_merge:
```
offnertho: I would rather name it what it does, then when it is executed. 

So I guess `trigger-ui-release` or `tag-ui-release` or something would fit better.

.github/workflows/release-on-merge.yml
```diff
@@ -0,0 +1,64 @@
...
+            const stableVersion = '${{ fromJson(steps.define-release-version.outputs.result).full }}'.split('-')[0]
+            const releaseDate = new Date().toISOString().split('T')[0]
+
+            const data = fs.readFileSync('./CHANGELOG.md',{encoding:'utf8', flag:'r'});
+
+            core.info(`Updating ${stableVersion} with date ${releaseDate} in Changelog`);
+
+            const changelogFileContents = updateChangeLog(data, stableVersion, releaseDate);
+
+            fs.writeFileSync('./CHANGELOG.md', changelogFileContents, 'utf-8');
```
offnertho: what happens if there is no changelog entry? should we fail the release?

.github/workflows/merge.yml
offnertho: Yeah makes sense. Lets tick with that.

.github/workflows/merge.yml
offnertho: ok

.github/workflows/release-on-merge.yml
mukulmishra18: Yeah, right now if there is no changelog the release will fail.

I think that should be okay, as every PR should have a changelog IMO.

.github/workflows/release-on-merge.yml
mukulmishra18: `trigger-ui-release` make sense 👍

.github/workflows/release-on-merge.yml
mukulmishra18: This step actually pass but later on when we commit the changes it fails as there is nothing to commit.

offnertho (approved):
LGTM. 

I guess when this is merged we need to cancel the release that is triggered with this?

.github/workflows/release-on-merge.yml
offnertho: 👍

PR #598, opened 2024-01-02
fix fontfamilyselectbox

LudoMichaud:
Addresses PW-13787


stonko1994 (changes requested):
The changes itself look good but the PR itself could get a facelift:
- Please update the PR description with a meaningful description since the internal issue number is not accessible publicly.
- Please add a changelog entry describing the fix.
- The branch name should follow the git-flow naming scheme `feature/...`

PR #599, opened 2024-01-03
Notify slack team during release workflow

mukulmishra18:
With automated release workflow, we need to notify team when workflow succeed or fail.

This PR adds that.


mimcz (approved):
LGTM 👍

PR #600, opened 2024-01-03
Fix failing release workflow

mukulmishra18:
We removed checkout step when we added the release workflow here https://github.com/bitmovin/bitmovin-player-ui/pull/590 although it is required.

This PR adds the checkout step, and also fixes calling `defineReleaseVersion` function.


mimcz (approved):
LGTM 👍

PR #611, opened 2024-03-21
Add compare link in changelog file

mukulmishra18:
## Description

When we update the `changelog` file in our release workflow to add release version and date, we are missing the compare link that should be added at the end of file.

This PR fixes that by adding links automatically.

- [x] `CHANGELOG` entry

### DIFF
--- a/.github/scripts/updateChangelog.js
+++ b/.github/scripts/updateChangelog.js
@@ -17,7 +17,7 @@ function updateChangeLog(changelogString, version, releaseDate) {
 
   return changelogString.replace(
     '## 1.0.0 (2017-02-03)\n- First release\n\n',
-    `## 1.0.0 (2017-02-03)\n- First release\n\n[${version}]: https://github.com/bitmovin/bitmovin-player-ui/compare/v${version}...v${lastReleaseVersion}\n`
+    `## 1.0.0 (2017-02-03)\n- First release\n\n[${version}]: https://github.com/bitmovin/bitmovin-player-ui/compare/v${lastReleaseVersion}...v${version}\n`
   );
 }
 
--- a/.github/scripts/updateChangelog.js
+++ b/.github/scripts/updateChangelog.js
@@ -12,7 +12,7 @@ function updateChangeLog(changelogString, version, releaseDate) {
     'gi',
   );
 
-  const lastReleaseVersion = changelogString.match(/\[(\d+.\d+.\d+)\]/)[1];
+  const lastReleaseVersion = changelogString.match(/## \[(\d+.\d+.\d+)\] - \d{4}-\d{2}-\d{2}/)[1];
   const changelogWithReleaseVersionAndDate = changelogString.replace(changelogVersionRegExp, `[${version}] - ${releaseDate}`);
 
   return changelogWithReleaseVersionAndDate.replace(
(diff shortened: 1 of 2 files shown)


.github/scripts/updateChangelog.js
```diff
@@ -6,12 +6,19 @@
...
-  
-  module.exports.updateChangeLog = updateChangeLog;
+  const optionalBetaOrRc = '(-rc.d+)?(-(b|beta).d+)?';
+  const changelogVersionRegExp = new RegExp(
+    `\\[(development|develop|unreleased|${version})${optionalBetaOrRc}.*`,
+    'gi',
+  );
+
+  const lastReleaseVersion = changelogString.match(/\[(\d+.\d+.\d+)\]/)[1];
+  changelogString = changelogString.replace(changelogVersionRegExp, `[${version}] - ${releaseDate}`);
```
felix-hoc: Mutating input arguments is not recommended - let's use a new variable instead, please 🙏🏼

.github/scripts/updateChangelog.js
```diff
@@ -6,12 +6,19 @@
...
+
+  return changelogString.replace(
+    '## 1.0.0 (2017-02-03)\n- First release\n\n',
+    `## 1.0.0 (2017-02-03)\n- First release\n\n[${version}]: https://github.com/bitmovin/bitmovin-player-ui/compare/v${lastReleaseVersion}...v${version}\n`
```
felix-hoc: interesting approach 😅 

Alternatively, we could have added a comment above the compare link section, and matched that 🤔 
Or matched the previous comparison link.

but I guess this is fine too, as it shouldn't change.

.github/scripts/updateChangelog.js
```diff
@@ -6,12 +6,19 @@
...
-  }
-  
-  module.exports.updateChangeLog = updateChangeLog;
+  const optionalBetaOrRc = '(-rc.d+)?(-(b|beta).d+)?';
+  const changelogVersionRegExp = new RegExp(
+    `\\[(development|develop|unreleased|${version})${optionalBetaOrRc}.*`,
+    'gi',
+  );
+
+  const lastReleaseVersion = changelogString.match(/\[(\d+.\d+.\d+)\]/)[1];
```
felix-hoc: Not sure if this is reliable enough - if we ever use a version in brackets in a changelog entry, this would break.

Maybe we could match the whole link?

```suggestion
  const lastReleaseVersion = changelogString.match(/\[(\d+.\d+.\d+)\]: https:\/\/github\.com\/bitmovin\/bitmovin-player-ui\/compare\//)[1];
```

felix-hoc (changes requested):
Looks good in general 👍🏼

.github/scripts/updateChangelog.js
mukulmishra18: I tried that before, but this is less complicated and has less chances of bug😄

.github/scripts/updateChangelog.js
mukulmishra18: We are matching the last released version. This is present at the starting of the file, after Unreleased section.

felix-hoc: got confused, thought you were taking it from the comparison link, not the header 🙈

felix-hoc (commented):
nice!

felix-hoc (approved):
nice!

PR #612, opened 2024-03-25
Rewrite `hasLocalStorage` check to not produce `localStorage`-entries

Andr3wid:
## Description
Existence of the `localStorage` API was probed so far by creating (and immediately deleting) a `localStorage`-entry. This is undesirable behaviour in case a page wishes to not have the `localStorage`-API utilized in any way. 

This PR introduces the following changes:

- Determine existence of the `localStorage`-API by checking if `getItem` and `setItem` functions are available instead of setting a storage item
- Handle exceptions directly in the `getItem`/`setItem` wrapper functions (e.g. if access is denied or quota got exceeded)

Anyways, a `localStorage` item would still be set if certain feature of the player-UI are used. PR <PR GOES HERE> is a followup to introduce disabling of `localStorage`.

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/storageutils.ts
+++ b/src/ts/storageutils.ts
@@ -17,8 +17,12 @@ export namespace StorageUtils {
    * @param data the item's data
    */
   export function setItem(key: string, data: string): void {
-    if (StorageUtils.isLocalStorageAvailable()) {
-      window.localStorage.setItem(key, data);
+    try {
+      if (StorageUtils.isLocalStorageAvailable()) {
+        window.localStorage.setItem(key, data);
+      }
+    } catch (e) {
+      /* Can get triggered e.g. by QuotaExceededError in Safari private mode */
     }
   }
 
@@ -28,9 +32,11 @@ export namespace StorageUtils {
    * @return {string | null} Returns the string if found, null if there is no data stored for the key
    */
   export function getItem(key: string): string | null {
-    if (StorageUtils.isLocalStorageAvailable()) {
-      return window.localStorage.getItem(key);
-    } else {
+    try {
+      if (StorageUtils.isLocalStorageAvailable()) {
+        return window.localStorage.getItem(key);
+      }
+    } catch (e) {
       return null;
     }
   }
--- a/src/ts/storageutils.ts
+++ b/src/ts/storageutils.ts
@@ -3,8 +3,8 @@ export namespace StorageUtils {
     try {
       return (
         window.localStorage &&
-        typeof localStorage.getItem === 'function' &&
-        typeof localStorage.setItem === 'function'
+        typeof localStorage.getItem === "function" &&
+        typeof localStorage.setItem === "function"
       );
     } catch (e) {
       return false;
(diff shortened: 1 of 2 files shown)


src/ts/storageutils.ts
```diff
@@ -39,8 +17,13 @@ export namespace StorageUtils {
    * @param data the item's data
    */
   export function setItem(key: string, data: string): void {
-    if (StorageUtils.hasLocalStorage()) {
-      window.localStorage.setItem(key, data);
+    try {
+      if (StorageUtils.isLocalStorageAvailable()) {
+        window.localStorage.setItem(key, data);
+      }
```
felix-hoc: the `isLocalStorageAvailable()` check does not require try-catch, that's already done internally. So we could pull it out of the try block and do an early return, or wrap the whole try-catch block with it. (same for `getItem()`)

src/ts/storageutils.ts
```diff
@@ -50,11 +33,14 @@ export namespace StorageUtils {
...
-    if (StorageUtils.hasLocalStorage()) {
-      return window.localStorage.getItem(key);
-    } else {
-      return null;
+    try {
+      if (StorageUtils.isLocalStorageAvailable()) {
+        return window.localStorage.getItem(key);
+      }
+    } catch (e) {
+      console.warn(`Failed to retrieve storage entry for ${key}`);
```
felix-hoc: IMO `warn` is too high of a log level, as we will always check for certain localStorage keys. I'd suggest `debug` or removing the log completely.

src/ts/storageutils.ts
```diff
@@ -1,36 +1,14 @@
...
-
-  export function hasLocalStorage(): boolean {
-    if (hasLocalStorageCache) {
-      return hasLocalStorageCache;
-    }
-
-    // hasLocalStorage is used to safely ensure we can use localStorage
-    // taken from https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Feature-detecting_localStorage
-    let storage: any = { length: 0 };
+  export function isLocalStorageAvailable(): boolean {
```
felix-hoc: Does this need to be exported?

felix-hoc (changes requested):
nice, just some suggestions :)

src/ts/storageutils.ts
Andr3wid: Ah yes, you are right. I thought that, when browser prohibits access, exceptions are triggered when actually trying to call `getItem` / `setItem`, but it already happens when trying to access `window.localStorage`. Hence, it's already covered in `isLocalStorageAvailable`.

I removed the `try-catch`es in this commit: [Removed superfluous try-catch](https://github.com/bitmovin/bitmovin-player-ui/pull/612/commits/3c2df0690034ff24b315e7a3820e13e4970f0571)

Let's have a call if you meant it in a different way :)

src/ts/storageutils.ts
Andr3wid: Removed the log entirely as well in [Removed superfluous try-catch](https://github.com/bitmovin/bitmovin-player-ui/pull/612/commits/3c2df0690034ff24b315e7a3820e13e4970f0571). Just not a fan of empty `catch` blocks :)

src/ts/storageutils.ts
Andr3wid: Doesn't have to I guess, we only use it inside `storageutils`. Just kept it that way because it was exported before as well:

[Remove export of internal function](https://github.com/bitmovin/bitmovin-player-ui/pull/612/commits/f3941616cacbc95a3e4acb94aa1d62581cd6d428)

felix-hoc (approved):

PR #615, opened 2024-03-28
Introduce `disableStorageApi` config option

Andr3wid:
## Description

- Introduces a new config-option named `disableStorageApi` that, if set to `true`, prevents the UI from accessing `localStorage`
- Fixes a bug where subtitle settings would get lost if UI variants switched during playback

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [3.56.0] - 2024-03-26
 
+### Added
+- `disableStorageApi` config option
+
+### Fixed
+- Subtitle settings not being retained when the UI variant switches
+
 ### Changed
 - `localStorage` availability check to not create a test-entry anymore
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## [3.56.0] - 2024-03-26
+## [Unreleased]
 
 ### Added
 - `disableStorageApi` config option
 
 ### Fixed
 - Subtitle settings not being retained when the UI variant switches
+
+## [3.56.0] - 2024-03-26
 
 ### Changed
 - `localStorage` availability check to not create a test-entry anymore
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [3.56.0] - 2024-03-26
 
+### Added
+- `disableStorageApi` config option
+
+### Fixed
+- Subtitle settings not being retained when the UI variant switches
+
```
felix-hoc, line 14: please move to the correct section

src/ts/components/subtitlesettings/subtitlesettingselectbox.ts
```diff
@@ -41,4 +41,8 @@ export class SubtitleSettingSelectBox extends SelectBox {
       this.overlay.getDomElement().addClass(this.currentCssClass);
     }
   }
+
+  configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
+      this.settingsManager = uimanager.subtitleSettingsManager;
```
felix-hoc: is there a chance of `settingsManager` getting accessed before `configure` is called?

src/ts/uimanager.ts
```diff
@@ -111,6 +113,7 @@ export class UIManager {
   private config: InternalUIConfig; // Conjunction of provided uiConfig and sourceConfig from the player
   private managerPlayerWrapper: PlayerWrapper;
   private focusVisibilityTracker: FocusVisibilityTracker;
+  private _subtitleSettingsManager: SubtitleSettingsManager;
```
felix-hoc: nit: would be nice to follow the same pattern we have for `config`, for example

felix-hoc (changes requested):

CHANGELOG.md
Andr3wid: [Move entry to correct section](https://github.com/bitmovin/bitmovin-player-ui/pull/615/commits/687a242328abdcbb5d8631a0a0a9996481fa1c5e)

src/ts/uimanager.ts
Andr3wid: [Replace get accessor with regular method](https://github.com/bitmovin/bitmovin-player-ui/pull/615/commits/c5d5e4f8000621dd6bb1141d45aa557782b83c00)

src/ts/components/subtitlesettings/subtitlesettingselectbox.ts
Andr3wid: Not the way how it's currently implemented. However, when one changes the implementation, we'd need to be careful (that applies for any `localStorage` operations, as its safety depends on the timing).

I at least made the prop optional to indicate it could be undefined:
[Replce get accessor with regular method](https://github.com/bitmovin/bitmovin-player-ui/pull/615/commits/c5d5e4f8000621dd6bb1141d45aa557782b83c00)

felix-hoc (approved):

PR #616, opened 2024-03-29
Add config option to allow seeking inside timeline marker w/ duration

luckygoyal-bitmovin:
## Description
Currently seeking anywhere between start position and duration of a timeline marker forces seek position to be beginning of the marker. The changes in this PR attempt to make this behaviour configurable so that integrators can disable this behaviour and allow seeking to a position inside timeline marker's duration.

- [X] `CHANGELOG` entry

### DIFF
--- a/src/ts/uiconfig.ts
+++ b/src/ts/uiconfig.ts
@@ -105,4 +105,9 @@ export interface UIConfig {
    * If set to true, prevents the UI from using `localStorage`.
    */
   disableStorageApi?: boolean;
+
+  /**
+   * If set to true, allows seeking inside timeline marker duration. Default is false
+   */
+  seekbarAllowSeekInMarkerDuration?: boolean;
 }
--- a/src/ts/uiconfig.ts
+++ b/src/ts/uiconfig.ts
@@ -76,7 +76,23 @@ export interface UIConfig {
    */
   disableAutoHideWhenHovered?: boolean;
   /**
-   * Specifies the seekbar snapping range in percentage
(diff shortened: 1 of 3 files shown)


src/ts/uiconfig.ts
```diff
@@ -105,4 +105,9 @@ export interface UIConfig {
    * If set to true, prevents the UI from using `localStorage`.
    */
   disableStorageApi?: boolean;
+
+  /**
+   * If set to true, allows seeking inside timeline marker duration. Default is false
+   */
+  seekbarAllowSeekInMarkerDuration?: boolean;
```
jmsn: Given that there is already a config option `seekbarSnappingRange` 

https://github.com/bitmovin/bitmovin-player-ui/blob/bffd82b8951734a3b2f4eb905c42481646f1c1cb/src/ts/uiconfig.ts#L79-L82

 I would propose a name like `seekbarSnappingEnabled` whose default value is `true` for the new option (and reverse the logic). This way it's more obvious that the two config options are related. Maybe, also worth clarifying in the doc of the property of `seekbarSnappingRange` that it has no effect if snapping is disabled.

I can take over the PR and do these changes if you agree @luckygoyal-bitmovin 🙂

luckygoyal-bitmovin: Thanks @jmsn . It will be great if you can takeover the PR.

jmsn: on it 👍

jmsn (approved):

PR #619, opened 2024-04-12
Fix release on merge from a fork

felix-hoc:
## Description
Release failed here: https://github.com/bitmovin/bitmovin-player-ui/actions/runs/8614645401?pr=617

Switch the action that triggers creating a tag for the release, so it gets run on the merge commit on the target branch: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

- [x] `CHANGELOG` entry


mukulmishra18 (approved):
🙏

PR #620, opened 2024-04-22
`ControlBar` not auto-hiding on some touch screen devices

zigavehovec:
## Description

Jira ticket: https://bitmovin.atlassian.net/browse/PA-2232

[Based on this comment](https://github.com/bitmovin/bitmovin-player-ui/blob/develop/src/ts/components/controlbar.ts#L43), the `UIConfig.disableAutoHideWhenHovered` flag is meant to be used only on non touch screen devices. 
The code that applied the config checked for `BrowserUtils.isMobile`, which returns false on some Android tablets. This caused the ControlBar to not hide on some Android tablets when `disableAutoHideWhenHovered` is set to `true`.

To fix the issue the `BrowserUtils.isMobile` check has been replaced with `BrowserUtils.isTouchSupported` which also covers tablets and generally better fits the use-case.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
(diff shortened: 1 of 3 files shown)


felix-hoc (approved):
Makes sense

stonko1994 (approved):

PR #621, opened 2024-05-03
Fix no subtitle on switching tracks

mukulmishra18:
## Description

When switching between different subtitle track, sometime it does not show the subtitle cue anymore.

We are clearing subtitle label component on `SubtitleEnabled` and `SubtitleDisabled`. This should be added again on `CueEnter` event.

The event sequence coming from Player is sometime off. When enabling subtitle, the sequence of event looks like, `SubtitleEnable`, `CueEnter` and `SubtitleEnabled`. This makes the Label component added and removed again, causing no subtitle being shown.

We are now not removing the component on `SubtitleEnabled` but just on `SubtitleDisabled`. This should be enough as a previous subtitle is disabled when a new one is enabled.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+- No subtitle is shown when switching between different tracks
+
 ## [3.61.0] - 2024-04-23
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


mukulmishra18:
Did some manual testing and the changes looks okay. The only thing I can see is that we are getting 2 `CueEnter` events when [`PlayerSubtitleAPI.enable`](https://cdn.bitmovin.com/player/web/8/docs/interfaces/Subtitles.PlayerSubtitlesAPI.html#enable.enable-1) is called with exclusive is false. Both of the cues are shown in the UI when second subtitle is enabled. This is also same with old implementation, we get 2 `CueEnter` events and those are shown in the UI.

felix-hoc (approved):
makes sense 👍🏼

PR #622, opened 2024-05-13
Disable auto-hide to not require two touches for click-through and skipping ads

hawk23:
ticket: https://bitmovin.atlassian.net/browse/PI-2655

## Description
Currently, when the small-screen ads UI is used on a touch device, two touches are needed to either hit the "skip ad" button or to open the ad  click-through URL. Especially for the "skip ad" button this is a confusing user experience as it is always visible and one touch should be sufficient to tap it.

To fix this, the auto hide is disabled for the small screen ads UI. After this change, the "skip ad" button can be triggered with a single touch. The same holds for triggering the ad click through URL.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Fixed
-- "Skip ad" button and ad click-through URL cannot be triggered with a single touch when small screen UI is used
+- "Skip ad" button and ad click-through URL cannot be triggered with a single touch when small screen ad UI is used
 
 ## [3.62.0] - 2024-05-06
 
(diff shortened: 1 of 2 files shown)


rolandkakonyi (approved):
Manual testing works as expected!

felix-hoc (approved):

felix-hoc (changes requested):
As discussed offline:
On Web, this causes the black shadow and fullscreen button to be visible during ad playback, which I think we don't want... Would be great if we could find a solution to make the Ad controls independent of the rest of the UI controls (but not sure how to do that 😅 )
Screenshot 2024-05-15 at 12 03 49

hawk23:
> As discussed offline: On Web, this causes the black shadow and fullscreen button to be visible during ad playback, which I think we don't want... Would be great if we could find a solution to make the Ad controls independent of the rest of the UI controls (but not sure how to do that 😅 )

An approach that still allows for auto hiding the control bar is now implemented in [this PR](https://github.com/bitmovin/bitmovin-player-ui/pull/624) (still Draft).

PR #623, opened 2024-05-16
Add Fast-Forward/Rewind button components

dweinber:
## Description
This PR adds a new `QuickSeekButton` component for the control bar. Via the configuration, it can be defined whether it should be a rewind or a fast-forward button. Further, the time that is skipped in the respective direction is configurable.

Please note that the component is exported for custom, externally managed UIs, but it is _not_ added to the default UI variants.

- [x] `CHANGELOG` entry

This is how it looks like when added:

Screenshot 2024-05-16 at 15 06 00

Screenshot 2024-05-16 at 15 05 46


src/ts/components/quickseekbutton.ts
```diff
@@ -0,0 +1,131 @@
...
+   * Specify how many seconds the player should seek forward/backwards in the stream.
+   * Default is 10.
+   */
+  seekSeconds?: number;
+
+  /**
+   * Specify whether the button should fast-forward or rewind.
+   * Default is rewind.
+   */
+  direction?: QuickSeekDirection;
```
felix-hoc: nit: instead of having the direction separate, we could control everything via the `seekSeconds` config: when `> 0`: direction is forward; when `< 0`: direction is back.

src/ts/demofactory.ts
```diff
@@ -118,6 +119,8 @@ export namespace DemoFactory {
         new Container({
           components: [
             new PlaybackToggleButton(),
+            new QuickSeekButton({ direction: QuickSeekDirection.Rewind, seekSeconds: 10 }),
+            new QuickSeekButton({ direction: QuickSeekDirection.Forward, seekSeconds: 10 }),
```
felix-hoc: Would it make sense to have these buttons as part of the default UI? I think they could be a good addition :)

src/scss/skin-modern/components/_quickseekbutton.scss
```diff
@@ -0,0 +1,18 @@
...
+    @include svg-icon-shadow;
+  }
+
+  &[data-#{$prefix}-seek-direction='forward'] {
+    background-image: url('../../assets/skin-modern/images/quickseek-fastforward.svg');
+  }
+
+  &[data-#{$prefix}-seek-direction='rewind'] {
+    background-image: url('../../assets/skin-modern/images/quickseek-rewind.svg');
+  }
```
felix-hoc, line 17: Do we need any specific styling for the mobile UI variant?

src/ts/components/quickseekbutton.ts
```diff
@@ -0,0 +1,131 @@
...
+        isLive = args.live;
+        switchVisibility(isLive, hasTimeShift);
+      },
+    );
+
+    // Initial detection
+    timeShiftDetector.detect();
+    liveStreamDetector.detect();
+
+    this.onClick.subscribe(() => {
```
felix-hoc, line 74: Probably out of scope, but it would be great if these would also trigger the seeks:
- Pressing Left/Right arrow key
- Double-tapping the left/right half of the video on mobile

Maybe for a follow-up :)

felix-hoc (commented):
Nice feature! 🎉  Looks good, just some nits/questions.

Also: Please update the branch.

src/ts/components/quickseekbutton.ts
dweinber: Great idea! Didn't think of that 😅 
Done in 345fcb2a

src/ts/demofactory.ts
dweinber: I think it would. However, changing anything in the default UI is not something taken too lightly, and internal discussions around this are planned. That's why for now we export the Button so it can be easily used for custom UIs and externally managed UIs, and only added it to the additional demo UI in the playground.

src/scss/skin-modern/components/_quickseekbutton.scss
dweinber: IMO there is nothing special needed for now.

src/ts/components/quickseekbutton.ts
dweinber: I agree, I think this is out of scope.


> - Pressing Left/Right arrow key

I agree, but then this might interfere with the functionality we already have in https://github.com/bitmovin/bitmovin-player-ui/blob/6e05cbdfc2d67ea3e11338accc086671b1c53054/src/ts/components/seekbarcontroller.ts#L83-L92


> - Double-tapping the left/right half of the video on mobile

I think this and/or a "HugeQuickSeekButton" for displaying within the video would be separate components (similar to `PlaybackToggleButton` and `HugePlaybackToggleButton`.
This is also the reason I explicitly mention in the PR description that this is a control bar component 🙂

src/ts/demofactory.ts
felix-hoc: Makes sense 👍🏼

src/ts/components/quickseekbutton.ts
felix-hoc: Yup, makes sense. Just wanted to note it down :D

felix-hoc (approved):
Pls update the branch & verify that nothing breaks (the base branch is quite a few commits ahead)

Other than that, LGTM :)

dweinber:
Luckily most changes where just around the automated release process. Seems that all the actual UI changes still work well from what I can tell.

However, I just found a case I didn't handle correctly before: If you hit the QuickSeekButton a few times and we don't have data buffered for that area, the player's currentTime / getTimeShift hasn't changed yet. This means that the additional button clicks where ignored in that case. This is handled now in the last two commits, please take another look at these changes - thanks!

felix-hoc (approved):

PR #624, opened 2024-05-16
Add config to allow default action for first touch

hawk23:
ticket: https://bitmovin.atlassian.net/browse/PI-2655

## Description
Currently, when the small-screen ads UI is used on a touch device, two touches are needed to either hit the "skip ad" button or to open the ad  click-through URL. Especially for the "skip ad" button this is a confusing user experience as it is always visible and one touch should be sufficient to tap it.

To fix this, a new configuration option `allowDefaultActionOnFirstTouch` is introduced on `UIContainerConfig` that can be used to control whether the default action should be executed after showing the hidden UI on a first touch. The `allowDefaultActionOnFirstTouch` flag is now set to `true` for `modernSmallScreenAdsUI`, allowing first touches to trigger the click-through URL and also the "skip ad" button.

## Discussion
A drawback of this approach is that also buttons from the title bar can receive this first touch now during ads. Specifically, the fullscreen button, which is the only one that is visible there during ad playback. 

There would be some ways to improve on this situation:
- (1) Attach a payload to the `onClick` event that determines if it is a first touch. Event receiving components could check for this flag and decide on their own if they want to handle `onClick` in case they are first touches. The "skip ad" button and the ad click overlay would be then implemented in a way that they accept `onClick`s that are first touches and the fullscreen button would ignore such `onClick` events.
- (2) Decide in `UIContainer.touchend` event handler if the default action should be prevented or not based on the event target.

(1) seems a bit more elegant than (2) as the components are allowed to decide for themselves if they want to react to first touches.

## Tests
- Please do manual testing

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/uicontainer.ts
+++ b/src/ts/components/uicontainer.ts
@@ -17,10 +17,12 @@ export interface UIContainerConfig extends ContainerConfig {
    * Default: 5 seconds (5000)
    */
   hideDelay?: number;
+  
   /**
    * An array of player states in which the UI will not be hidden, no matter what the {@link hideDelay} is.
    */
   hidePlayerStateExceptions?: PlayerUtils.PlayerState[];
+  
   /**
    * The HTML element on which user interaction events (e.g. mouse and touch events) will be tracked to detect
    * interaction with the UI. These basically trigger showing and hiding of the UI.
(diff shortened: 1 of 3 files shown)


src/ts/components/uicontainer.ts
```diff
@@ -143,13 +156,13 @@ export class UIContainer extends Container<UIContainerConfig> {
       name: 'touchend',
       handler: (e) => {
         if (!isUiShown) {
-          // Only if the UI is hidden, we prevent other actions (except for the first touch) and reveal the UI
-          // instead. The first touch is not prevented to let other listeners receive the event and trigger an
-          // initial action, e.g. the huge playback button can directly start playback instead of requiring a double
-          // tap which 1. reveals the UI and 2. starts playback.
-          if (isFirstTouch && !player.isPlaying()) {
```
hawk23, line 150: `isFirstTouch` flag was removed as it seems redundant. As we are inside an `if` that checks for `!isUiShown` and because we always call `showUi()` from inside this if, we will always just enter this code path on first touch after UI was hidden.

stonko1994:
Closing in favour of #626 & #627

PR #626, opened 2024-05-23
Assign components to their `HTMLElements`

stonko1994:
### Problem
We are trying to fix an issue on touch devices that certain elements, which are always visible, need to be touched twice in order to trigger the desired click action. This affects the `AdSkipButton` and the `AdClickOverlay`.

See this PR for more details on the problem and the solution: #627

### Changes
As preparation for fixing the problem in a feasible way, this PR attaches the `Component` instances to their corresponding `DOM` / `HTMLElements` for easier access. This removes the need to traverse all components and find the correct instance e.g. in a `touchend` listener.

#### Details
- A new type (`HTMLElementWithComponent`) was introduced to prevent `any` casts. This type adds an optional `component` attribute to the `HTMLElement` for setting and accessing of it.
The `DOM` constructor was extended to allow passing the `Component` on `DOM` element creation.
- Only the root DOM element of Components gets the instance attached. This means that child components from elements (e.g. a `Label`, which is a child element of `Button`) do not have any component attached to it. A parent lookup needs to be performed to get the component from those elements. This was done to prevent confusion about where to get the instance and where not.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Added
+- `Component` instances are now assigned to their `HTMLElements` for easier accessing
+
 ## [3.63.0] - 2024-05-17
 
 ### Added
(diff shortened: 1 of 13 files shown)


Andr3wid (approved):
Nicely separated PRs 🙌

PR #629, opened 2024-06-28
Improve demopage

dweinber:
## Description
Two small but hopefully useful improvements of the playground demo page:
1. A checkbox to enable/disable ads on the fly in the demo page UI as shown below. Previously, this needed to be done somewhere in the demo page's code.
    image
2. Store the basic configuration of the playground (video source, UI, ads enabled) in localStorage of the browser and load on page load

- [ ] `CHANGELOG` entry

### DIFF
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -272,11 +272,17 @@
       updateApiMethods();
     });
 
-    // Add UI
-    uiManager = bitmovin.playerui.UIFactory.buildDefaultUI(player, uiConfig);
+    const playgroundConfig = getConfigFromStorage();
 
-    player.load(sources.fullyFeatured).then(function() {
+    // Add UI
+    uiManager = bitmovin.playerui.UIFactory[playgroundConfig.uiOption || 'buildDefaultUI'](player, uiConfig);
+
+    player.load(sources[playgroundConfig.source || 'fullyFeatured']).then(function() {
       console.log('source successfully loaded');
+
+      if (playgroundConfig.adsEnabled) {
+        scheduleAds();
+      }
     }, function(errorEvent) {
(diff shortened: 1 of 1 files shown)


src/html/index.html
```diff
@@ -256,6 +262,8 @@ <h4 class="card-title">
   var uiManager;
   bitmovin.playerui.UIManager.setLocalizationConfig({ language: 'en', browserLanguageDetection: false });
 
+  bitmovin.player.Player.addModule(bitmovin.player["advertising-bitmovin"].default);
```
felix-hoc, line 265: Why the switch to the Bitmovin Ads module? The default is IMA, shouldn't the UI use the default player config?

felix-hoc (changes requested):
Looks like getting the default config isn't working properly when localstorage is empty: 
Screenshot 2024-06-28 at 14 08 33

src/html/index.html
dweinber: Because with IMA there is no control over the UI as IMA provides the minimal ads UI which cannot officially (and must not) be customized. With BAM this is possible to use our UI, so that made more sense to me.

dweinber:
Thanks @felix-hoc! Classic case of testing it, then refactoring the code and introduce a bug by that 😄 
Should be fixed now in https://github.com/bitmovin/bitmovin-player-ui/pull/629/commits/28cfa59f09cca218a0d91b152f2c2a121820d9a6

felix-hoc (approved):
Ok, nice :) 

I'm just thinking it would be good to a have an option to reset the local storage somehow, so it doesn't need to be done through the browser. Could be a follow-up PR though.

PR #633, opened 2024-07-12
build the structure of the new mobile ui with the old ui components

alesmaye:
## Description

the layout of the new mobile UI version was build using the old/existing components(with old icons etc.)

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/uifactory.ts
+++ b/src/ts/uifactory.ts
@@ -449,7 +449,120 @@ export function superModernAdsUI(){
 }
 
 export function superModernMobileUI(){
-  return new UIContainer({});
+  let subtitleOverlay = new SubtitleOverlay();
+
+  let mainSettingsPanelPage = new SettingsPanelPage({
+    components: [
+      new SettingsPanelItem(i18n.getLocalizer('settings.video.quality'), new VideoQualitySelectBox()),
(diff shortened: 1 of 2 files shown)


src/ts/uifactory.ts
```diff
@@ -449,7 +449,120 @@ export function superModernAdsUI(){
...
+      new RecommendationOverlay(),
+      controlBar,
+      new TitleBar({
+        components: [
+          new MetadataLabel({ content: MetadataLabelContent.Title }),
+          new CastToggleButton(),
+          new AirPlayToggleButton(),
+          new VRToggleButton(),
+          //TODO: make a Share button
+          //new PictureInPictureToggleButton(), probably not needed in mobile version
```
bitmovin-kenny: Since this won't come back i guess you can just remove it.

src/ts/uifactory.ts
```diff
@@ -449,7 +449,120 @@ export function superModernAdsUI(){
...
+        ],
+      }),
+      settingsPanel,
+      new Watermark(),
```
bitmovin-kenny: I don't think this should be in the final version.

bitmovin-kenny (changes requested):
Looks like a good start, just 2 minor requests for change.

bitmovin-kenny (approved):
nice

PR #634, opened 2024-07-15
Subtitle toggle button implementation

alesmaye:
## Description

New component: SubtitleToggleButton
Initially when the button is pressed, the SubtitleSelectBox shows up.
Once a subtitle is selected, it can simply be turned on and off via the SubtitleToggleButton without having to open the settings.

- [x] `CHANGELOG` entry


src/scss/bitmovinplayer-ui.scss
```diff
@@ -1 +1,2 @@
 @import 'skin-modern/skin';
+@import 'skin-super-modern/skin';
```
bitmovin-kenny, line 2: I don't think this should be included in the default ui.
If you look here https://github.com/bitmovin/bitmovin-player-ui/tree/support/v2.x/src/scss there have been two separate css files for different UI variants, but maybe we should double check with @stonko1994 .

src/scss/skin-super-modern/components/_subtitletogglebutton.scss
```diff
@@ -0,0 +1,18 @@
+@import '../../skin-modern/variables';
+@import '../../skin-modern/mixins';
```
bitmovin-kenny: Is this just a placeholder or will this be final?
I think the super-modern ui should have its own variables and I would suggest creating it now to avoid forgetting a usage of the old one.

src/ts/components/subtitletogglebutton.ts
```diff
@@ -0,0 +1,63 @@
+import { ToggleButton, ToggleButtonConfig } from './togglebutton';
+import { i18n } from '../localization/i18n';
+import { PlayerAPI } from 'bitmovin-player';
+import { UIInstanceManager } from '../uimanager';
+import { StorageUtils } from '../storageutils';
+import { SubtitleSelectBox } from './subtitleselectbox';
+import { SettingsPanel } from './settingspanel';
+
+
```
bitmovin-kenny: Please remove one of these empty lines. Weird that the linter does not complain there 🤔

src/ts/components/subtitletogglebutton.ts
```diff
@@ -0,0 +1,63 @@
...
+        const defaultConfig: ToggleButtonConfig = {
+            cssClass: 'ui-subtitletogglebutton',
+            onClass: 'subtitles-on',
+            offClass: 'subtitles-off',
+            text: i18n.getLocalizer('settings.subtitles'),
+            ariaLabel: i18n.getLocalizer('settings.subtitles'),
+        };
+
+        this.config = this.mergeConfig(config, defaultConfig as ToggleButtonConfig, this.config);
```
bitmovin-kenny: default config should already be of type `ToggleButtonConfig`, since it's declared like this.
```suggestion
        this.config = this.mergeConfig(config, defaultConfig, this.config);
```
should do the trick.

src/ts/components/subtitletogglebutton.ts
```diff
@@ -0,0 +1,63 @@
...
+        super.configure(player, uimanager);
+        this.player = player;
+        this.hide();
+
+        player.on(player.exports.PlayerEvent.SubtitleAdded, this.checkSubtitleAvailability);
+        player.on(player.exports.PlayerEvent.SubtitleRemoved, this.checkSubtitleAvailability);
+
+        this.onClick.subscribe(() => {
+            const availableSubtitles = player.subtitles.list();
+            const subtitleID = StorageUtils.getItem('bmpiu-subtitlelanguage');
```
bitmovin-kenny: Two things here:
- The subtitle ID will be different for different videos, I am not sure that it's a good idea to store the ID. Maybe rather store a language + selected state (like `{ language: 'en', active: true }`) and try to find a matching subtitle for the new video.
Something like 
```ts
// const stored = {language: 'en', active: true};
if (stored.active) {
    const toActivate = player.subtitles.list().find(subtitle => subtitle.lang === stored.language);
    if (toActivate) {
        this.on(toActivate)
    } else {
        this.settingsPanel.show();
    }
}
```
- Do not use hard-coded prefix for `bmpiu-subtitlelanguage` rather use `DummyComponent.instance().prefixCss('subtitlelanguage');`

src/ts/subtitleutils.ts
```diff
@@ -36,6 +37,7 @@ export class SubtitleSwitchHandler {
         }
       } else {
         this.player.subtitles.enable(value, true);
+        StorageUtils.setItem('bmpiu-subtitlelanguage', value);
```
bitmovin-kenny: Again, please don't use the hardcoded prefix.

bitmovin-kenny (changes requested):
Looks good, just have some small remarks.

src/scss/bitmovinplayer-ui.scss
stonko1994: Hm good question 🤔 The separate file would have the advantage that customers could explicitly decide they don't need one of the skins (requires forking it). However, when used through one single CSS (as precompiled) both will be in there anyway.

I think it's fine to keep both in here going forward.

src/ts/components/dummycomponent.ts
```diff
@@ -0,0 +1,20 @@
...
+            DummyComponent._instance = new DummyComponent();
+        }
+
+        return DummyComponent._instance;
+    }
+
+    public prefixCss(cssClassOrId: string): string {
+        return super.prefixCss(cssClassOrId);
+    }
+}
```
bitmovin-kenny: I would prefer to just export a function `prefixCSS` instead of a `DummyComponent`, that name is misleading. You can just export a function and hide the DummyComponent like 
```ts
class DummyComponent extends Component<ComponentConfig> {
    private static _instance: DummyComponent;

    public static instance(): DummyComponent {
        if (!DummyComponent._instance) {
            DummyComponent._instance = new DummyComponent();
        }

        return DummyComponent._instance;
    }

    public prefixCss(cssClassOrId: string): string {
        return super.prefixCss(cssClassOrId);
    }
}

/**
 * Prefixes any CSS Class or ID with the configured prefix.
 */
export function prefixCss(cssClassOrId: string): string {
    return DummyComponent.instance().prefixCss(cssClassOrId);
}
```
then using it also becomes easier to read `this.localStorageKey = prefixCss('subtitlesettings');`

src/ts/components/subtitletogglebutton.ts
```diff
@@ -57,7 +64,25 @@ export class SubtitleToggleButton extends ToggleButton<ToggleButtonConfig> {
         });
     }
 
+    private initPlayerEvents = () => {
+        this.player.on(this.player.exports.PlayerEvent.SubtitleAdded, this.checkSubtitleAvailability);
+        this.player.on(this.player.exports.PlayerEvent.SubtitleRemoved, this.checkSubtitleAvailability);
+        this.player.on(this.player.exports.PlayerEvent.PeriodSwitch, this.checkSubtitleAvailability);
```
bitmovin-kenny: I would rather use the `PeriodSwitched` event. This event signals that the content transition has been completed in contrast to `PeriodSwitch` which signals the start of content transition.

src/ts/components/subtitletogglebutton.ts
```diff
@@ -57,7 +64,25 @@ export class SubtitleToggleButton extends ToggleButton<ToggleButtonConfig> {
...
+        this.player.on(this.player.exports.PlayerEvent.SourceUnloaded, this.checkSubtitleAvailability);
+    }
+
     private checkSubtitleAvailability = () => {
```
bitmovin-kenny: I don't think the name is still fitting. Maybe something like `onAvailableSubtitlesChanged`

src/ts/components/subtitletogglebutton.ts
```diff
@@ -57,7 +64,25 @@ export class SubtitleToggleButton extends ToggleButton<ToggleButtonConfig> {
...
-        this.player.subtitles.list().length > 0 ? this.show() : this.hide();
+        const storedSubtitle: StoredSubtitleLanguage = StorageUtils.getObject(DummyComponent.instance().prefixCss('subtitlelanguage'));
+        const subtitleList = this.player.subtitles.list();
+
+        // only shows the button when subtitles are existing
+        subtitleList.length > 0 ? this.show() : this.hide();
+
+        // if the stored subtitle is set active and available, that subtitle is enabled
+        let subtitleToActivate: SubtitleTrack;
+        if (storedSubtitle && storedSubtitle.active && (subtitleToActivate = subtitleList.find(subtitle => subtitle.lang === storedSubtitle.language))) {
```
bitmovin-kenny: This if statement does a bit much and assignments in if statements are always frowned upon.
If you are worried about nesting if statements you can always use early returns:
```ts
        if (!storedSubtitle || !storedSubtitle.active) {
            return;
        }

        // if the stored subtitle is set active and available, that subtitle is enabled
        const subtitleToActivate: SubtitleTrack = subtitleList.find(subtitle => subtitle.lang === storedSubtitle.language);
        if (subtitleToActivate) {
            this.on();
            this.player.subtitles.enable(subtitleToActivate.id);
        }
```

src/ts/subtitleutils.ts
```diff
@@ -54,6 +57,22 @@ export class SubtitleSwitchHandler {
     this.uimanager.getConfig().events.onUpdated.subscribe(this.refreshSubtitles);
   }
 
+  /**
+   * @param subtitleID (optional) If set, the according stored subtitle language will be set to active; If not set, the stored stored subtitle language will be set to inactive
+   */
+  public static setSubtitleLanguageStorage = (player: PlayerAPI, subtitleID?: string) => {
```
bitmovin-kenny, line 63: This only gets called when any UI component is interacted with. In the case that the player would have subtitles already enabled when the UI is built or when something different from the UI enables a subtitle, this would not be affected.

I think the approach here would be to call this on component initialization with `player.subtitles.list().filter(sub => sub.enabled)` and then only reactively call this when the player fires a `SubtitleEnabled` or `SubtitleDisabled` event.

bitmovin-kenny (changes requested):
Nice progress, I still have some minor complaints though :)

src/ts/subtitleutils.ts
alesmaye: Already works as intended.

src/ts/components/subtitletogglebutton.ts
```diff
@@ -65,22 +65,22 @@ export class SubtitleToggleButton extends ToggleButton<ToggleButtonConfig> {
...
+        const storedSubtitle: StoredSubtitleLanguage = StorageUtils.getObject(prefixCss('subtitlelanguage'));
         const subtitleList = this.player.subtitles.list();
 
         // only shows the button when subtitles are existing
         subtitleList.length > 0 ? this.show() : this.hide();
 
         // if the stored subtitle is set active and available, that subtitle is enabled
-        let subtitleToActivate: SubtitleTrack;
-        if (storedSubtitle && storedSubtitle.active && (subtitleToActivate = subtitleList.find(subtitle => subtitle.lang === storedSubtitle.language))) {
+        let subtitleToActivate: SubtitleTrack = subtitleList.find(subtitle => subtitle.lang === storedSubtitle.language);
```
bitmovin-kenny:
```suggestion
        const subtitleToActivate: SubtitleTrack = subtitleList.find(subtitle => subtitle.lang === storedSubtitle?.language);
```
if storedSubtitle can be null, this is a potential `null.language` access

bitmovin-kenny (changes requested):
Almost there

src/ts/components/subtitletogglebutton.ts
```diff
@@ -0,0 +1,88 @@
...
+            if (!subtitleTrack) {
+                this.settingsPanel.show();
+            } else if (this.isOff() && subtitleTrack) {
+                this.on();
+                player.subtitles.enable(subtitleTrack.id);
+                SubtitleSwitchHandler.setSubtitleLanguageStorage(this.player, subtitleTrack.id);
+            } else {
+                this.off();
+                player.subtitles.disable(subtitleTrack.id);
+                SubtitleSwitchHandler.setSubtitleLanguageStorage(this.player);
```
bitmovin-kenny, line 57: Since the bindSelectionEvent already handles all subtitle changes... do we even need the calls to `SubtitleSwitchHandler.setSubtitleLanguageStorage` here or would it also work without them?

src/ts/subtitleutils.ts
bitmovin-kenny: Today I learned :)

src/ts/components/subtitletogglebutton.ts
alesmaye: Tried it out... does not work without

bitmovin-kenny (approved):
Nicely done

PR #643, opened 2024-08-19
Completed new Settings menu

alesmaye:
## Description

Implemented a new Settings Menu. Now the settings are selected in a submenu instead of a drop down menu.
Sub-Settings are getting dynamically loaded and appended to the DOM tree.
Changed a little bit of the design of the settings menu.
Subtitle settings are not integrated yet.

- [x] `CHANGELOG` entry


stonko1994 (commented):
As discussed offline let's try to reduce the code duplication in the ts components by using inheritance for the modern settings components.

stonko1994 (approved):
There are still a few things to do but they will be done on a follow up PR

PR #650, opened 2024-09-05
[PW-11902] Improve auto-hide behavior with active dropdown menus

HalfbyteHeroes:
## Description

This PR improves the auto-hide behavior of the UI when the dropdown menu of a `SelectBox` is shown.

Before the UI would auto-hide after the specified timeout regardless of whether a dropdown menu was currently shown or not. With the changes introduced in this PR, the UI will now stay visible and auto-hiding will be disabled whenever a dropdown is shown.

In order to do so, the following changes were implemented:
- allowed `Timeout` execution to be suspended, meaning all calls to `start` and `reset` will be ignored until timeout execution is resumed
- added a `ViewMode` to `Component` with the following values:
  -  `Persistent`: the component shall be visible at all times and UI auto-hiding shall be disabled
  - `Temporary`: the component may be hidden at any time
- modified the `SelectBox` to set its `ViewMode` to `Persistent` whenever and as long as the dropdown of the associated `select` tag is shown
- modified both `UIContainer` and `SettingsPanel` to suspend hide-timeout execution as long as there is at least one component with a `Persistent` view mode
- fixed a bug in Safari were the dropdown menu of a `SelectBox` would stay visible when the UI is hidden

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Added
+- `Component` now has a `ViewMode` that can either be `Persistent` or `Temporary`
+
+### Fixed
+- `selectbox` dropdown not closing in Safari when the UI is hidden
+
+### Changed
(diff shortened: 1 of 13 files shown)


wasp898 (approved):
Tested on Safari, Firefox and Chrome. New behavior makes more sense. 👍

PR #653, opened 2024-11-19
Reduce forced reflows in Seekbar

felix-hoc:
## Description

There are quite a few [Forced Reflows](https://web.dev/articles/avoid-large-complex-layouts-and-layout-thrashing?utm_source=devtools#avoid-forced-synchronous-layouts) caused by the UI Seekbar:

Screenshot 2024-11-19 at 14 00 35

This can impact performance.

### Changes
- Disable SmoothPlaybackPositionUpdater while UI is hidden
- force-update position when UI becomes visible and on Seek

Note: The Seekbar position is still regularly updated on `TimeChanged` events from the player. This is left as-is to prevent a big jumping animation when the UI is shown.

- [x] `CHANGELOG` entry


Andr3wid (approved):
Nice 👍  Approving already before changelog entry :)

PR #654, opened 2024-12-04
Enable the new UI design on browsers

stonko1994:
## Description

Our new UI design is currently only enabled for small screen variants which is meant for mobile. We also want to have it on all platforms though. This PR enables the new design for browsers.

### Changes:
- Use the new UI design on browsers
- Revert the loading indicator to something we had before
- Revert margins/padding/sizing back to the previous values for better browser look and feel
- Change font weight from bold to medium
- Use the new Fullscreen icon everywhere

### Key parts which are still missing
- TV UI
- Ads UI
- Subtitle styling settings

- [x] `CHANGELOG` entry **A final changelog entry for the new design will be added at the end**


src/ts/uifactory.ts
```diff
@@ -547,8 +543,108 @@ export namespace UIFactory {
...
+      components,
+    });
+
+    let settingsPanel = new ModernSettingsPanel({
+      components: [mainSettingsPanelPage],
+      hidden: true,
+      pageTransitionAnimation: true,
+    });
+
+    // let subtitleSettingsPanelPage = new SubtitleSettingsPanelPage({
```
stonko1994, line 568: Will be added in #655

bitmovin-kenny:
At 100% render size i get a scroll bar when opening the settings, I guess this should not happen.

image

Otherwise LGTM

stonko1994:
> At 100% render size i get a scroll bar when opening the settings, I guess this should not happen.
> 
> image
> 
> Otherwise LGTM

It looks like there is a tiny bit of scrolling. The settings panel will be 'reworked' in #655 where I can't reproduce this anymore.

bitmovin-kenny (approved):
ok

PR #655, opened 2024-12-09
Enable the subtitle settings panel for the new layout

stonko1994:
## Description

The current version of the new UI layout does not support subtitle customization settings.

### Changes
In order to add support for the subtitle customization settings I had to refactor quite a bit inside the `SettingsPanel`. The changes are outlined in detail below.

- Components no longer are separated into modern and non-modern. The name of the components now reflect their purpose.
  - `ModernSettingsPanelItem` was refactored and split into a `DynamicSettingsPanelItem` and a `SettingsPanelSelectOption` instead of handling both in the same component.
  - The `ModernSettingsPanelPage` and `ModernSettingsPanel` were removed and their functionality was moved into the parent components.
- The `SubtitleSettingsPanelPage` now supports the old and the new layout.

- [x] `CHANGELOG` entry **A generic redesign changelog with all changes will be added at the end**


src/ts/components/button.ts
```diff
@@ -71,7 +71,8 @@ export class Button<Config extends ButtonConfig> extends Component<Config> {
     }).html(i18n.performLocalization(this.config.text)));
 
     // Listen for the click event on the button element and trigger the corresponding event on the button component
-    buttonElement.on('click', () => {
+    buttonElement.on('click', (e) => {
+      e.stopPropagation();
```
stonko1994, line 75: This is needed to prevent the event from bubbling up to the parent element. 

Example: In the new design, the whole `SettingsPanelItem` is clickable. The subtitle customization item also contains a button to a different page. Without this change, both would trigger when clicking the button.

src/ts/components/component.ts
```diff
@@ -339,7 +339,7 @@ export class Component<Config extends ComponentConfig> {
    * @param base configuration inherited from a superclass
    * @returns {Config}
    */
-  protected mergeConfig<Config>(config: Config, defaults: Config, base: Config): Config {
+  protected mergeConfig<Config>(config: Config, defaults: Partial<Config>, base: Config): Config {
```
stonko1994, line 342: Not every component can fill all of its default config values. It does not need to fill out all as others are required anyway on initialization. To improve the type handling I decided to make this `Partial` here. The alternative (which was used before) was a cast to the Config type.

src/ts/components/modernsettingspanelitem.ts
stonko1994, line 1: Was refactored to the `DynamicSettingsPanelItem`

src/ts/components/seekbar.ts
```diff
@@ -747,6 +747,8 @@ export class SeekBar extends Component<SeekBarConfig> {
       this.onSeekedEvent(targetPercentage);
     };
 
+    // TODO: the problem is that the SeekbarLabel is also inside the seekbarContainer and therefore reacts to the
+    // TODO: mouse inputs.
```
stonko1994: Will be handled in a follow up PR.

src/ts/components/settingspanel.ts
```diff
@@ -97,9 +97,11 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
         this.hide();
         this.hideHoveredSelectBoxes();
       });
-      this.getDomElement().on('mouseenter', () => {
-        // On mouse enter clear the timeout
-        this.hideTimeout.clear();
+      this.getDomElement().on('mouseenter mousemove', () => {
```
stonko1994, line 100: I had to add the `mousemove` event here as well because the `focusout` event triggers between `SettingsPanelPage` transitions causing the settings panel to hide after selecting anything.

src/ts/components/settingspanel.ts
```diff
@@ -113,6 +115,15 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
...
     }
 
+    if (config.pageTransitionAnimation) {
+      const handleResize = () => {
+        // Reset the dimension of the settingsPanel to let the browser calculate the new dimension after resizing
+        this.getDomElement().css({ width: '', height: '' });
+      };
+
+      player.on(player.exports.PlayerEvent.PlayerResized, handleResize);
+    }
```
stonko1994, line 125: There is a workaround within the animation handling where the width and height of the `SettingsPanel` are set to the current dimension. This breaks the `SettingsPanel` when resizing the browser window. Removing the width and height properties on resize fixes this.

src/ts/components/settingspanel.ts
```diff
@@ -250,6 +261,11 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
     super.addComponent(component);
   }
 
+  addPage(page: SettingsPanelPage) {
+    this.addComponent(page);
+    this.updateComponents();
+  }
```
stonko1994, line 273: Convenience method to add a `SettingsPanelPage` element.

src/ts/components/settingspanel.ts
```diff
@@ -300,8 +316,8 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
     this.animateNavigation(targetPage, sourcePage, skipAnimation);
 
     this.updateActivePageClass();
-    targetPage.onActiveEvent();
     sourcePage.onInactiveEvent();
+    targetPage.onActiveEvent();
```
stonko1994, line 326: The previous order was 'wrong' and caused an issue with the new dynamic `SettingsPanelPage` approach, where the previous page would get the inactive event after the new page got the active event.

src/ts/components/settingspanelitem.ts
```diff
@@ -11,47 +11,74 @@ import {PlaybackSpeedSelectBox} from './playbackspeedselectbox';
...
   private label: Component<ComponentConfig>;
-  protected setting: Component<ComponentConfig>;
+  protected setting: Component<ComponentConfig> | null;
 
   private settingsPanelItemEvents = {
-    onActiveChanged: new EventDispatcher<SettingsPanelItem, NoArgs>(),
+    onActiveChanged: new EventDispatcher<SettingsPanelItem<Config>, NoArgs>(),
   };
 
-  constructor(label: LocalizableText | Component<ComponentConfig>, setting: Component<ComponentConfig>, config: ContainerConfig = {}, addSettingAsComponent: boolean = true) {
```
stonko1994, line 30: I refactored the constructor to follow our config-based approach, which we follow for almost all other Components.

src/ts/components/settingspanelitem.ts
```diff
@@ -11,47 +11,74 @@ import {PlaybackSpeedSelectBox} from './playbackspeedselectbox';
...
+  constructor(config: SettingsPanelItemConfig)
```
stonko1994, line 50: This second constructor overload is a workaround for the [weak-type-detection](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html#weak-type-detection) issue when there is no property of the parent type used.

Not adding this requires to cast the config to a `SettingsPanelItemConfig` on every usage which is cumbersome.

src/ts/components/modernsettingspanelitem.ts
stonko1994, line 1: Was refactored to the DynamicSettingsPanelItem

src/scss/skin-super-modern/components/_settingspanelpage.scss
```diff
@@ -6,92 +6,6 @@
   &.#{$prefix}-active {
     display: block;
   }
-
-  // A "line" in the panel: a container holding a label + control
```
stonko1994, line 10: The styling for a settings-panel-item was extracted into a separate file.

src/ts/components/subtitlesettings/subtitlesettingspanelpage.ts
```diff
@@ -16,16 +14,19 @@ import {WindowOpacitySelectBox} from './windowopacityselectbox';
...
+import { ListSelector, ListSelectorConfig } from '../listselector';
 
 /**
  * @category Configs
  */
-export interface SubtitleSettingsPanelPageConfig extends ContainerConfig {
+export interface SubtitleSettingsPanelPageConfig extends SettingsPanelPageConfig {
   settingsPanel: SettingsPanel;
   overlay: SubtitleOverlay;
+  useDynamicSettingsPanelItem?: boolean;
```
stonko1994, line 29: We still don't know how we want to ship the new UI layout. We are currently leaning towards a new major version. Once this decision is confirmed I'll rework this and just change to the new dynamic approach and throw out the config option again. For now I didn't want to create a subclass for this as the code is very similar.

src/ts/components/settingspanelitem.ts
```diff
@@ -11,47 +11,74 @@ import {PlaybackSpeedSelectBox} from './playbackspeedselectbox';
...
+ */
+export interface SettingsPanelItemConfig extends ContainerConfig {
+  /**
+   * The label component or the text for the label.
+   */
+  label?: LocalizableText | Component<ComponentConfig>;
+  /**
+   * The component that configures a setting.
+   */
+  setting?: Component<ComponentConfig>;
```
wasp898: Nit: Maybe it should just be called component.

wasp898 (approved):

src/ts/components/settingspanelitem.ts
stonko1994: I went for `settingComponent` 👍 [59a7bf5](https://github.com/bitmovin/bitmovin-player-ui/pull/655/commits/59a7bf5ec18e6c9b2615dbb7240dd53fe8dcaac1)

PR #658, opened 2024-12-10
Improve Seekbar mouse handling

stonko1994:
## Description

Revert Seekbar to its current state on development and apply new style by adding a new container element for all bars.

### Problem
Currently, the mouse interactions are bound to the container element of the seekbar. This comes with a problem that the mouse events are also applied to the `SeekbarLabel`, which should not be the case. Also, the hitbox for seeking becomes smaller and the playback position marker no longer acts as a seeking handle.

### Changes
To fix this problem while keeping the new design approach, I added another container around all 'bars' inside the seekbar. Therefore the mouse events can stay where they are and this new container can be used for the new styling (mainly for the rounded corners)

- [x] `CHANGELOG` entry **A general entry will be added at the end**

### DIFF
--- a/src/scss/skin-super-modern/components/_seekbar.scss
+++ b/src/scss/skin-super-modern/components/_seekbar.scss
@@ -46,14 +46,18 @@ $seekbar-height: .3125em;
     @include hidden;
     @include focusable;
 
-    border-radius: 1em;
     cursor: pointer;
     font-size: 1em;
     height: $seekbar-height;
     margin: calc((1em - $seekbar-height) / 2) 0;
-    overflow: hidden;
     position: relative;
     width: 100%;
+
+    .#{$prefix}-seekbar-bars {
+      @extend %bar;
+      border-radius: 1em;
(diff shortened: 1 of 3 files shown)


src/scss/skin-super-modern/components/_seekbar.scss
```diff
@@ -46,15 +46,19 @@ $seekbar-height: .3125em;
     @include hidden;
     @include focusable;
 
-    border-radius: 1em;
     cursor: pointer;
     font-size: 1em;
     height: $seekbar-height;
     margin: calc((1em - $seekbar-height) / 2) 0;
-    overflow: hidden;
```
stonko1994, line 54: Having the overflow hidden here breaks the playback position marker as it is bigger than this container. As a workaround initially we rendered the position marker outside of the seekbar and added the mouse handler to the container as well. This breaks some other UX functionalities. See details in the PR description.

Andr3wid (approved):
Looks good 👌

PR #662, opened 2024-12-20
Switch from tslint to eslint

felix-hoc:
## Description

`tslint` is deprecated. Let's switch to `eslint`.

## Changes
- Add eslint dependencies and config, based on the recommended rules
- Apply quick fixes where possible
- Disable rules that cause many errors
- Remove `tslint` dependencies
- Enable `skipLibCheck` in tsconfig since there is some incompatibility between types (could not drill down exactly where it comes from)

- [x] `CHANGELOG` entry


src/ts/components/listselector.ts
```diff
@@ -77,12 +77,7 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
   }
 
   private getItemIndex(key: string): number {
-    for (let index in this.items) {
-      if (key === this.items[index].key) {
-        return parseInt(index);
-      }
-    }
-    return -1;
+    return this.items.findIndex((item) => item.key === key);
```
bitmovin-kenny: Just to be sure: Is this supported on old TVs?

eslint.config.mjs
```diff
@@ -0,0 +1,49 @@
...
+        parserOptions: {
+          projectService: true,
+          tsconfigRootDir: __dirname,
+        },
+      },
+    },
+  ],
+  rules: {
+    'no-prototype-builtins': 'off',
+    'prefer-const': 'off', // TODO: enable rule and run autofix
```
bitmovin-kenny, line 29: I don't think this TODO should go into main... can you remove it now, did you run autofix?

eslint.config.mjs
```diff
@@ -0,0 +1,49 @@
+// @ts-check
```
bitmovin-kenny, line 1: why is this here?

bitmovin-kenny (changes requested):
Very nice 💪 
Just a few very minor nits.

eslint.config.mjs
felix-hoc: There are a ton of changes that this would produce - so I'd rather do it in an isolated PR. Since https://github.com/bitmovin/bitmovin-player-ui/pull/656 has a lot of changes, I'd also like to avoid too many conflicts for them.

eslint.config.mjs
felix-hoc: It enables type checking in a JavaScript file, which is otherwise disabled in the tsconfig. It was part of the default config, and I thought it made sense to keep it :)

src/ts/components/listselector.ts
felix-hoc: I thought so, since we already use it a couple of times. But according to MDN it's only available since Chrome 45 😳 Let me run some tests.

eslint.config.mjs
bitmovin-kenny: OK that is very considerate of you.

src/ts/components/listselector.ts
felix-hoc: I rewrote the function to not use `findIndex` :)

bitmovin-kenny (approved):
👍

PR #665, opened 2025-01-07
Create separate regions for CEA608 rows

felix-hoc:
## Description
Issue: PW-22216

### Problem
For CEA-608 captions, the “window color” + “window opacity” affect the whole subtitle overlay, blocking the view on the content.

### Changes
- create separate `region` components for CEA-608 rows
- the absolute vertical position is now assigned to each row / region
  - as CEA-608 has exactly 15 rows, this is pre-calculated in CSS
- the actual cues are no longer positioned `absolute`-ly, and only hold the left offset

- [x] `CHANGELOG` entry


bitmovin-kenny (approved):
Looks good even with two CEA Caption tracks active at the same time.

PR #666, opened 2025-01-08
Fix styling for the initially displayed CEA-608 cue

felix-hoc:
## Description

Issue: PW-22216

When enabling CEA captions within the time range of a cue, this cue is not styled as expected.

This is because we reset the CEA caption setup (mainly adding a `cea-608` class to the subtitle overlay that causes caption styling to be applied) on a `subtitleenabled` event. The CEA caption setup gets done on a `cueenter` event, but that one may arrive before.

The event flow is as follows:
1. `subtitleenable`
2. `cueenter`
3. `subtitleenabled`

Thus, we should reset the CEA caption setup on `subtitleenable`, to get a clean slate for whatever subtitle format arrives in the following `cueenter`.

Furthermore, there was a slight issue with vertical alignment of caption cues when the player size was rather small.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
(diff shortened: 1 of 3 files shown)


bitmovin-kenny (approved):
LGTM

PR #670, opened 2025-01-10
Get rid of the terms `skin`, `modern` and `default`

stonko1994:
## Description

For our new UI we want to cleanup the usage of the terms `skin`, `default` and `modern` since none of them was really in use.

- We only have one skin and there is no real 'feature' behind that.
- There is no other UI than the `default` or the `modern` one (which are the same) therefore there is no point in keeping this naming scheme. Also if we ever introduce another skin it's hard to come up with a matching name aside modern.

Don't be afraid of the PR size. Most of it is just renaming and moving Code around. All other changes are explicitly highlighted below.

Those changes are API breaking so pleas review with that in mind ‼️


src/scss/bitmovinplayer-ui.scss
```diff
@@ -1 +1,107 @@
-@import 'skin-modern/skin';
```
stonko1994, line 1: Since we will only have one UI look and feel going forward (like now) there is no need anymore to add this extra level of 'abstraction'. I simply copied the content of `_skin.scss` file directly into this one. The code was not changed.

src/ts/components/dynamicsettingspanelitem.ts
```diff
@@ -91,7 +91,7 @@ export class DynamicSettingsPanelItem extends SettingsPanelItem<DynamicSettingsP
       let selectedOptionLabelText = selectedItem.label;
       if (this.settingComponent instanceof SubtitleSelectBox) {
         let availableSettings = this.settingComponent.getItems().length;
-        selectedOptionLabelText = selectedOptionLabelText + ' (' + (availableSettings - 1) + ')';
+        selectedOptionLabelText = i18n.performLocalization(selectedOptionLabelText) + ' (' + (availableSettings - 1) + ')';
```
stonko1994, line 94: This was a linter error as the + operator is not allowed on the `LocalizableText` type. This is actually a bug fix.

src/ts/demofactory.ts
stonko1994, line 1: The DemoFactory was fully removed as it only adds complexity and we have the very same code already in our player-web samples. This was never meant to be used publicly.

src/ts/uifactory.ts
stonko1994, line 1: This file got slightly restructured but no actual code changes expect method names renaming - removing the default and modern terms.

The structure now is the following:
- UIFactory (Contains all methods which should be used publicly)
  - buildUI
  - buildSmallScreenUI
  - buildCastReceiverUI
  - buildTvUI
- private top level functions which builds only the UIContainer.

This change was done as it was always confusing what is the difference between `buildSmallScreenUI` and `smallScreenUI` functions.

In an effort to futhre reduce complexly I renamed the functions which only build the `UIContainer` and added the `Layout` suffix.

src/ts/uifactory.ts
```diff
@@ -522,122 +150,526 @@ export namespace UIFactory {
...
+  export function buildCastReceiverUI(player: PlayerAPI, config: UIConfig = {}): UIManager {
+    return new UIManager(player, castReceiverUILayout(), config);
   }
 
-  export function buildModernTvUI(player: PlayerAPI, config: UIConfig = {}): UIManager {
+  /**
+   * Builds a UI which is used on TVs.
+   *
+   This UI includes variants for:
+   - TVs
```
jmsn:
```suggestion
   * This UI includes variants for:
   * - TVs
```

src/ts/uifactory.ts
```diff
@@ -522,122 +150,526 @@ export namespace UIFactory {
     );
   }
 
-  export function buildModernCastReceiverUI(player: PlayerAPI, config: UIConfig = {}): UIManager {
-    return new UIManager(player, modernCastReceiverUI(), config);
+  /**
+   * Builds a UI which is used on cast receivers.
+   *
+   This UI includes variants for:
+   - Cast Receivers
```
jmsn:
```suggestion
   * This UI includes variants for:
   * - Cast Receivers
```

src/ts/uifactory.ts
```diff
@@ -55,432 +55,49 @@ import { DynamicSettingsPanelItem } from './components/dynamicsettingspanelitem'
...
-    });
-  }
-
-  export function buildModernUI(player: PlayerAPI, config: UIConfig = {}): UIManager {
+  /**
+   * Builds a fully featured UI with all basic layouts for all platforms.
+   *
+   * This UI includes variants for:
+   * - Default
+   * - Ads
```
jmsn: I'm a bit cautious regarding mixing the terms `layout` and `variant` here 😅 

Should we say `This UI includes layouts for:` instead?

src/ts/uifactory.ts
```diff
@@ -55,432 +55,49 @@ import { DynamicSettingsPanelItem } from './components/dynamicsettingspanelitem'
 import { TouchControlOverlay } from './components/touchcontroloverlay';
 import { AdStatusOverlay } from './components/adstatusoverlay';
 
+/**
+ * Provides factory methods to create different UI layouts.
```
jmsn: Is `layouts` the right term here? I think we are actually not exposing layouts (such as `tvUILayout` etc) but complete UIs, right?

Since we're trying to be more clear in the terms we're using I'm a bit more nitpicky 😉

jmsn (commented):
LGTM, just have some NITs

src/ts/uifactory.ts
stonko1994: [57a46cb](https://github.com/bitmovin/bitmovin-player-ui/pull/670/commits/57a46cbe7d298cebcf1b06d75eda146a3b7c907c)

src/ts/uifactory.ts
stonko1994: [57a46cb](https://github.com/bitmovin/bitmovin-player-ui/pull/670/commits/57a46cbe7d298cebcf1b06d75eda146a3b7c907c)

src/ts/uifactory.ts
stonko1994: Probably not. I tried to improve the documentation here: [9f3d78c](https://github.com/bitmovin/bitmovin-player-ui/pull/670/commits/9f3d78c8e3d96d7396dff1e1e7f7381828d515a9)

src/ts/uifactory.ts
stonko1994: https://github.com/bitmovin/bitmovin-player-ui/pull/670#discussion_r1913552602

jmsn (approved):

PR #671, opened 2025-01-13
Structure files in folders

stonko1994:
### Problem
All files of our UI are basically into a put into a root folder just grouping them by language. This makes it very hard to discover them and get a feel which components belong together.

### Changes
Group all files into folders. Folders either group them by feature (seekbar, settings) or by type of the component (buttons, labels, etc)

_There are no code changes expect import path changes_


wasp898 (approved):
👍

PR #673, opened 2025-01-13
Automatically resolve TV UI when using full UI

stonko1994:
### Problem
Our TV UI does not automatically apply when running the Player/UI unmodified on a TV platform.

### Changes
Automatically resolving to the TV UI variant when running on a TV platform.
- Adding `isTv` to the `UIConditionContext`
- Adding additional platform detections
- Adding keymaps for those platforms


src/ts/uifactory.ts
```diff
@@ -669,3 +681,8 @@ function tvUILayout() {
     spatialNavigation: spatialNavigation,
   };
 }
+
+function tvAdsUILayout() {
+  // TODO: implement once we have a design for TV ads
```
stonko1994: It will be implemented once the Design is finalized.

felix-hoc (commented):
as discussed, we could wait for a player update to simplify the logic :)

src/ts/browserutils.ts
```diff
@@ -47,6 +47,11 @@ export class BrowserUtils {
     return navigator && navigator.userAgent && navigator.platform === 'MacIntel';
   }
 
+  static get isTv(): boolean {
+    return this.isHisense || this.isPlayStation || this.isWebOs
+      || this.isTizen || this.isVizio || this.isXumo || this.isXbox;
```
felix-hoc: we could also check for VIDAA OS:

```ts
/VIDAA/.test(navigator.userAgent);
```

felix-hoc (approved):
makes sense, one suggestion otherwise LGTM

src/ts/browserutils.ts
stonko1994: [8304d91](https://github.com/bitmovin/bitmovin-player-ui/pull/673/commits/8304d91a95ddfebd44aadebb3f18343ac39559f9)

PR #674, opened 2025-01-13
Add empty state UI

stonko1994:
### Problem
When the initial source loading takes longer than loading the UI, the values displayed by the UI are fallback values. Usually it shows a duration of `00:00` and a current time of `00:01`. This is particular problematic when the stream is actually a live stream.

### Changes
To prevent this issue we add an empty state UI with only the `PlaybackToggleButton`, `Buffering-` and `ErrorOverlay`. This improves the initial UI displayed in case no Source is loaded yet.


src/ts/UIManager.ts
```diff
@@ -365,6 +374,7 @@ export class UIManager {
       this.managerPlayerWrapper.getPlayer().on(this.player.exports.PlayerEvent.SourceUnloaded, resolveUiVariant);
       this.managerPlayerWrapper.getPlayer().on(this.player.exports.PlayerEvent.Play, resolveUiVariant);
       this.managerPlayerWrapper.getPlayer().on(this.player.exports.PlayerEvent.Paused, resolveUiVariant);
+      this.managerPlayerWrapper.getPlayer().on(this.player.exports.PlayerEvent.Playing, resolveUiVariant);
```
dweinber, line 377: Didn't you want to resolve the UI variant (also) on `Ready`?

dweinber (approved):
One open question, but code looks good and seems to work nicely in my tests!

src/ts/UIManager.ts
stonko1994: Do you think Ready is needed? 🤔 We have the `SourceLoaded` already which I use as 'state' inside the Context. As I understand it, the duration (and therefore knowing if live or VOD) is available in the `SourceLoaded` event. The UI doesn't need to know if the Player is ready for immediate playback.

I only added the `PlayingEvent` here because the context already contains the `isPlaying` state, however the `resolveUiVariant` is not called in `Playing`.

dweinber: Understood and makes sense 🙂

PR #675, opened 2025-01-14
Remove top level `ui` css class

stonko1994:
## Description

When we started working on the new UI design, we introduced it as a new skin. However, this idea was discarded, and we want to make it the only UI design.

To enable the skinning feature, we added a top-level CSS class that wrapped everything. This is no longer needed. To keep breaking changes in CSS low, we can now remove this top level CSS class again to keep the old CSS selector compatibility.


src/scss/bitmovinplayer-ui.scss
stonko1994, line 1: Review this file with `Hide Whitespaces` enabled.

felix-hoc (approved):

PR #678, opened 2025-01-14
Improve UX on dismissing the `SettingsPanel`

stonko1994:
## Description

Currently the SettingsPanel only dismisses when explicitly clicking the settings button again, or after a timeout (web only). Therefore, it's tedious to close the SettingsPanel after changing a setting.

### Changes
- Adding a key-listener to close the SettingsPanel when hitting ESC (depending on the hardware).
- Adding a new `DismissClickOverlay` which can be used to dismiss any Component when it's visible.


src/ts/components/settings/SettingsPanel.ts
```diff
@@ -124,6 +126,13 @@ export class SettingsPanel extends Container<SettingsPanelConfig> {
       player.on(player.exports.PlayerEvent.PlayerResized, handleResize);
     }
 
+    const maybeCloseSettingsPanel = (event: KeyboardEvent) => {
+      const action = getKeyMapForPlatform()[event.keyCode];
```
stonko1994, line 130: I intentionally use the deprecated `event.keyCode` here to support other Devices that do not have an Escape button. Such as TVs or gaming consoles. The `Action.BACK` mapping is already covered using the `keyCode`.

jmsn (approved):
cool beans

PR #679, opened 2025-01-15
Support for customizing the font style and character edge color of subtitles/captions

dweinber:
## Description
The current subtitle/caption styling did not allow for changing the font style (default, italic, bold) and the character edge color. These two options are now added as well.

- [x] `CHANGELOG` entry


src/scss/skin-modern/_mixins.scss
```diff
@@ -144,3 +144,19 @@
     outline: none;
   }
 }
+
+/// Replace `$search` with `$replace` in `$string`. From https://css-tricks.com/snippets/sass/str-replace-function/
+/// @author Kitty Giraudel
```
stonko1994: Do we want to keep this?

src/ts/components/subtitlesettings/characteredgecolorselectbox.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @category Components
+ */
+export class CharacterEdgeColorSelectBox extends SubtitleSettingSelectBox {
+
+  constructor(config: SubtitleSettingSelectBoxConfig) {
+    super(config);
+
+    this.config = this.mergeConfig(config, {
+      cssClasses: ['ui-subtitlesettingscharacteredgecolorselectbox'],
```
stonko1994: I started to use kebab case for CSS classes for new components. I know that this format aims for consistency but it's very hard to read 🫣 Fine for me to keep it like this but I'd be also fine with switching to kebab case.

stonko1994 (approved):
Code looks good just two minor questions. 

+1 In the current UI, the Subtitle Styling Settings Panel is now very big. This is fixed in the upcoming UI layout but do we want to back-port the scrollable/height handling to the current UI?

src/scss/skin-modern/_mixins.scss
dweinber: I thought it would be nice to keep the attribution. But the link alone could also be sufficient. What's your preference?

dweinber:
Thanks! 

>  In the current UI, the Subtitle Styling Settings Panel is now very big. This is fixed in the upcoming UI layout but do we want to back-port the scrollable/height handling to the current UI?

That would be good if someone could do that 😉

src/ts/components/subtitlesettings/characteredgecolorselectbox.ts
dweinber: Yes, I was aiming for constistency. But I also agree that this is quite hard to read. Changed in 6c81793374643d435561823974469e2ec03e4c9b

src/scss/skin-modern/_mixins.scss
stonko1994: I don't have a strong opinion. Both works for me 👍

dweinber: Updated in 8c5bab4d7413435849d711f5822c42ea1c895383

PR #681, opened 2025-01-15
Remove Watermark per default

stonko1994:
## Description

The Watermark is usually the first thing customers remove when they customize their UI. It was agreed that we remove the watermark per default.

To simplify enabling it e.g. for our own Demos we also introduced a UIConfig property to configure the visibility of it.

- [x] `CHANGELOG` entry **Will be added at the end with all changes**


src/ts/UIConfig.ts
```diff
@@ -131,4 +131,11 @@ export interface UIConfig {
    * Specifies if the `EcoModeToggleButton` should be displayed within the `SettingsPanel`
    */
   ecoMode?: boolean;
+  /**
+   * Specifies if the Watermark element should be included in the UI.
+   * Per default the Watermark shows the Bitmovin Logo.
```
wasp898:
```suggestion
   * Per default, the Watermark shows the Bitmovin Logo.
```

wasp898 (approved):
👍

PR #683, opened 2025-01-16
Improve click handling on touch devices

stonko1994:
## Description

There is an issue on touch input devices where some settings panel elements need to be touched twice in order to change them. This is due to the `:hover` effect which we currently use.
In addition sometimes the click event is not triggered on non button elements ~on touch devices~ in combination with our iOS Player SDK. ~To improve this situation I added the `touchend` to those elements. To keep the functionality for buttons which are inside such elemnts, we also need to add the `touchend` to the button component.~ This will be addressed in the iOS Player SDK itself.


src/scss/components/settings/_settingspanelitem.scss
```diff
@@ -10,7 +10,13 @@
     cursor: pointer;
   }
 
-  &:hover {
+  @media (hover: hover) {
```
stonko1994: This only adds the hover style on devices where hover is supported.

src/ts/components/settings/DynamicSettingsPanelItem.ts
```diff
@@ -102,7 +102,8 @@ export class DynamicSettingsPanelItem extends SettingsPanelItem<DynamicSettingsP
     const handleItemClick = () => {
       this.displayItemsSubPage();
     };
-    this.getDomElement().on('click', (e) => {
+    this.getDomElement().on('click touchend', (e) => {
```
stonko1994: @dweinber found an issue with this approach. The touchend triggers and opens the sub menu when releasing the touch after scrolling within the settings panel.

bitmovin-kenny (changes requested):
I guess you are missing a changelog, otherwise LGTM

bitmovin-kenny (approved):
OK since it's "only" going into the V4 branch, not to main

src/ts/components/settings/DynamicSettingsPanelItem.ts
stonko1994: This is fixed again by removing the `touchend` listeners again in: [d596dbf](https://github.com/bitmovin/bitmovin-player-ui/pull/683/commits/d596dbfa16efed555da83b45962ea694210d97fe)

Note: The `touchend` listeners were a solution for an issue that the click event did not trigger every time on non button components. However, this was only observed in combination with the iOS Player SDK. This issue is already known and the rootcause is within the Player SDK where we call the native code from the Javascript context. This is not performant enough and blocks the Javascript execution. We will tackle this in the iOS Player SDK.

PR #684, opened 2025-01-17
Add keyboard support to the new `SettingsPanel` design

stonko1994:
## Description

The new `SettingsPanel` design doesn't support keyboard interactions yet. In the old layout this was handled by the browser as we used `SelectBox`es. Since we switched to a custom element in the new layout we need to add the keyboard support manually.

This PR introduces a new `InteractiveSubtitleSettingsPanelItem`, adding key listeners that trigger the existing click event. The new type is used as super type for the `DynamicSettingsPanelItem` and the `SettingsPanelSelectOption`.


src/scss/_mixins.scss
```diff
@@ -129,9 +129,13 @@
...
 }
 
-@mixin focusable {
+@mixin focusable($inset: false) {
   &:focus {
-    box-shadow: $focus-element-box-shadow;
+    @if $inset == false {
+      box-shadow: $focus-element-box-shadow;
+    } @else {
+      box-shadow: inset $focus-element-box-shadow;
```
stonko1994, line 137: The box shadow has to be inset for settings panel items as otherwise it is not visible (they are at the edge of the settings panel and it has the overflow hidden)

src/ts/components/Container.ts
```diff
@@ -129,6 +129,10 @@ export class Container<Config extends ContainerConfig> extends Component<Config>
       'aria-label': i18n.performLocalization(this.config.ariaLabel),
     }, this);
 
+    if (typeof this.config.tabIndex === 'number') {
```
stonko1994, line 132: The `ComponentConfig` already has the tabIndex property but it's not consistently applied when creating the dom elements.

src/ts/components/buttons/Button.ts
```diff
@@ -77,6 +77,10 @@ export class Button<Config extends ButtonConfig> extends Component<Config> {
       this.onClickEvent();
     });
 
+    buttonElement.on('focusin focusout', (e) => {
+      e.stopPropagation();
+    });
```
stonko1994, line 82: This is needed to stop bubbling of the `focusin` event to parents. This is a problem in the new settings panel where the subtitle settings button is inside a focusable settings panel item. Without that the `focusin` fires on both elements which is undesired.

bitmovin-kenny (approved):
As discussed in slack, it would be a nice to have to be able to navigate the settings panel with the arrow keys (up/down selects the row and left/right goes into/out of sub menus).

But that might be a feature for another day.

PR #686, opened 2025-02-20
Upgrade github actions to v4

wasp898:
## Description

Fixes build error:

```
This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`.
```

## Checklist (for PR submitter and reviewers)

- [/] `CHANGELOG` entry not needed

PR #680, opened 2025-01-15
Migrate Akamai CDN URLs to Bitmovin CDN

rolandkakonyi:
## Description

We are moving away from Akamai CDN in favor of Bitmovin CDN, therefore this PR updates all URLs to use the same assets from the Bitmovin CDN.

## Checklist (for PR submitter and reviewers)

- [x] `CHANGELOG` entry - not needed

PR #687, opened 2025-02-20
Upgrade release workflow to v4

wasp898:
## Description

The previous [PR](https://github.com/bitmovin/bitmovin-player-ui/pull/686) updated the actions for the CI build workflow, but the release workflow is still failing.
Since the release failed, we have an inconsistent release version, so we need to downgrade the package version again and remove the tag.
Furthermore, it looks like a release without changelog causes an inconsistent changelog, so we need to fix that as well as another PR was also recently merged without any entries: https://github.com/bitmovin/bitmovin-player-ui/pull/680.

- [x] `CHANGELOG` entry

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,7 +31,7 @@ jobs:
             const { version } = require('./package.json')
             core.setOutput('packageJsonVersion', version)
 
-      - uses: actions/download-artifact@v3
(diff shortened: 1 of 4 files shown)


mukulmishra18 (approved):
🤞

felix-hoc (approved):

PR #688, opened 2025-02-20
Stop release early when no changelog entry is added

felix-hoc:
## Description

Merging a PR without a changelog entry could get us into an inconsistent state, as we will still push a tag and try to release, but eventually fail trying to update the version in the Changelog file.

This change allows us to merge "chore" type PRs without a changelog, skipping a release.

- [x] `CHANGELOG` entry

### DIFF
--- a/.github/workflows/tag-release-version.yml
+++ b/.github/workflows/tag-release-version.yml
@@ -12,6 +12,9 @@ jobs:
     if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
     steps:
+      - name: Ensure Changelog
+        run: gh pr view ${{github.event.pull_request.number}} --json files -q '.files[].path' | grep CHANGELOG.md
+
       - name: Checkout
         uses: actions/checkout@v4
         with:
--- a/.github/workflows/tag-release-version.yml
+++ b/.github/workflows/tag-release-version.yml
@@ -17,7 +17,8 @@ jobs:
     - name: Ensure Changelog
       shell: bash
       run: |
-        if gh pr view ${{github.event.pull_request.number}} --json files -q '.files[].path' | grep CHANGELOG.md; then
+        changes=$(gh pr view ${{github.event.pull_request.number}} --json files -q '.files[].path')
+        if echo $changes | grep -q CHANGELOG.md; then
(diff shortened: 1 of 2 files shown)


.github/workflows/tag-release-version.yml
```diff
@@ -12,6 +12,9 @@ jobs:
     if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
     steps:
+      - name: Ensure Changelog
+        run: gh pr view ${{github.event.pull_request.number}} --json files -q '.files[].path' | grep CHANGELOG.md
```
mukulmishra18: Is this command available on ubuntu by default or we need to install? Maybe some authentication is required to access the PR or public repos does not require that? 🤔 

I was actually thinking to add a step in between that checks if we have `Unreleased` section in the changelog, but if this command works then it will be better.

felix-hoc: Afaik it's preinstalled on GitHub runners. PR access should also be possible.

The caveat is, that this will throw an error (exit code 1) when the changelog has not been modified, which is treated as a workflow failure.

mukulmishra18: > The caveat is, that this will throw an error (exit code 1) when the changelog has not been modified, which is treated as a workflow failure.

That means the `failure()` check will not work to notify slack team?

We could try moving this step as a separate workflow, maybe that works out?

Probably the easier would be to just have a check for `Unreleased` section. We have a regex in `updateChangelog.js` that can be reused:
```ts
const changelogVersionRegExp = new RegExp(
  `\\[(development|develop|unreleased|${version})${optionalBetaOrRc}.*`,
  'gi',
);
```

felix-hoc: > That means the failure() check will not work to notify slack team?

It will work - as it is now, a missing changelog will result in a slack notification. I'm not sure if we want that (open for discussion) but IMO it's better to have half of a release.

I have not yet found a clean way to stop the execution without causing a failure, apart from splitting it into a separate job and adding a condition to the existing one.

mukulmishra18 (approved):
Make sense 👍 

As discuss maybe as next step we can also skip the release if changelog only has `Chores` entry.

PR #689, opened 2025-02-24
Fix release

felix-hoc:
## Description

The PR https://github.com/bitmovin/bitmovin-player-ui/pull/688 broke the release workflow.

### Changes
Instead of `gh`, use git directly with existing key.

### Tests

Temporarily added the same step to the CI workflow:
- w/o changelog entry: https://github.com/bitmovin/bitmovin-player-ui/actions/runs/13502493628/job/37724151331#step:3:13
- with changelog entry: https://github.com/bitmovin/bitmovin-player-ui/actions/runs/13503723044/job/37728300549#step:3:13

- [x] `CHANGELOG` entry

### DIFF
--- a/.github/workflows/tag-release-version.yml
+++ b/.github/workflows/tag-release-version.yml
@@ -16,6 +16,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
 
     - name: Ensure Changelog
       shell: bash
--- a/.github/workflows/tag-release-version.yml
+++ b/.github/workflows/tag-release-version.yml
@@ -17,13 +17,15 @@ jobs:
     - name: Ensure Changelog
       shell: bash
       run: |
-        changes=$(gh pr view ${{github.event.pull_request.number}} --json files -q '.files[].path')
+        changes=$(git diff --name-only $GITHUB_BASE_REF)
         if echo $changes | grep -q CHANGELOG.md; then
           echo "CHANGELOG.md has been updated"
           echo "HAS_CHANGELOG=true" >> $GITHUB_OUTPUT
         else
           echo "CHANGELOG.md has not been updated, skipping release"
         fi
+      with:
+        ssh-key: ${{ secrets.RELEASE_DEPLOY_KEY }}
 
   trigger-ui-release:
     needs: verify-changelog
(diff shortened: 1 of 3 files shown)


.github/workflows/tag-release-version.yml
```diff
@@ -14,10 +14,17 @@ jobs:
       HAS_CHANGELOG: ${{ steps.ensure-changelog.outputs.HAS_CHANGELOG }}
     runs-on: ubuntu-latest
     steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
mukulmishra18, line 20: Does it checkout develop by default? 🤔 

Also regarding the `fetch-depth`, the documentation says:
```
Number of commits to fetch. 0 indicates all history for all branches and tags.
```

Do we want to fetch everything? For all branches? Can we limit this just to development?

.github/workflows/tag-release-version.yml
```diff
@@ -14,10 +14,17 @@ jobs:
...
+
     - name: Ensure Changelog
       shell: bash
       run: |
-        changes=$(gh pr view ${{github.event.pull_request.number}} --json files -q '.files[].path')
+        git config --global user.name 'Automated Release'
+        git config --global user.email 'release-automation@bitmovin.com'
+        changes=$(git diff --name-only ${{github.event.pull_request.base.ref}})
```
mukulmishra18, line 27: I guess this will not include files of some other PR that is merged in between current PR base ref?

mukulmishra18 (changes requested):
Couple of remarks from my side.

Would be nice if we can test this somehow. Maybe change the workflow a bit so it trigger if PR is merged in a temp branch and not in develop.

.github/workflows/tag-release-version.yml
felix-hoc: yes we need to checkout the whole history, otherwise it's possible that no merge point can be found between the current branch and `develop`. Technically we could also just fetch `develop` and the current branch, but there shouldn't be many active branches anyway so it's not really expensive and doesn't take long.

.github/workflows/tag-release-version.yml
felix-hoc: that's an interesting question - I think git only shows files as diffs when there is a change on the current branch on top of the base branch, and not the other way

felix-hoc:
> Would be nice if we can test this somehow. Maybe change the workflow a bit so it trigger if PR is merged in a temp branch and not in develop.

I tested it by adding the step to the `CI` workflow (see executions linked in the PR description).

.github/workflows/tag-release-version.yml
mukulmishra18: Got it.

mukulmishra18 (approved):
> I tested it by adding the step to the CI workflow (see executions linked in the PR description).

Missed that. Thank you for testing 🙌

PR #690, opened 2025-02-25
Fix release - again

felix-hoc:
### Problem
https://github.com/bitmovin/bitmovin-player-ui/pull/688 broke the release workflow and https://github.com/bitmovin/bitmovin-player-ui/pull/689 didn't fix it.

The issue is that the workflow runs _after_ the PR is merged, meaning `git diff` cannot be used to determine which files have changed.

### Changes
Switch back to using `gh`, which accesses GitHub via API to get the data.

The token is a default token that is available in workflows: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication

### Tests
This was tested on a separate branch with the same step, in the `CI` workflow:
- https://github.com/bitmovin/bitmovin-player-ui/actions/runs/13518733904
- https://github.com/bitmovin/bitmovin-player-ui/actions/runs/13518746502

- [x] `CHANGELOG` entry

### DIFF
--- a/.github/workflows/tag-release-version.yml
+++ b/.github/workflows/tag-release-version.yml
@@ -16,21 +16,19 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v4
-      with:
-        fetch-depth: 0
 
     - name: Ensure Changelog
       shell: bash
       run: |
-        git config --global user.name 'Automated Release'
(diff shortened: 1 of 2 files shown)


mukulmishra18 (approved):
Third time's a charm 🤞

PR #692, opened 2025-03-19
Change font size also for CEA-608 captions

mimcz:
## Description

As CEA-608 captions are using grid view which scales relatively to the player UI it did not take the font size settings into consideration. However the spec suggest to allow adjustment between 50% to 200%. As the grid view display implementation is quite complex this PR adds adjustment to the whole grid setup so the font size scales without making the letters or lines to overlap. As the grid positioning may break on larger sizes we also disable the positions and letter spacing if the font size factor is larger than 1.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-### Fixed
+### Added
 - Changing font size now also takes an effect in CEA-608 captions with cap from 50% to 200%. Larger than default font size also disables the grid view.
 
 ## [3.88.0] - 2025-02-25
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+- Changing font size now also takes an effect in CEA-608 captions with cap from 50% to 200%. Larger than default font size also disables the grid view.
+
 ## [3.88.0] - 2025-02-25
 
 ### Changed
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -103,7 +103,8 @@ gulp.task('lint-sass', function() {
   return gulp.src(paths.source.sass)
   .pipe(sassLint({
     rules: {
-      'no-css-comments': 0
+      'no-css-comments': 0,
+      'property-sort-order': 0
     }
   }))
   .pipe(sassLint.format())
--- a/spec/components/subtitleoverlay.spec.ts
+++ b/spec/components/subtitleoverlay.spec.ts
@@ -87,34 +87,30 @@ describe('SubtitleOverlay', () => {
       subtitleOverlay.configure(playerMock, uiInstanceManagerMock);
     });
 
-    it('correctly applies font size factor within allowed range (0.5 to 2.0)', () => {
-      // Set to minimum allowed factor
-      subtitleOverlay.setFontSizeFactor(0.2);
-      expect(subtitleOverlay['FONT_SIZE_FACTOR']).toBe(0.5); // Should be clamped to 0.5
-
-      // Set to maximum allowed factor
-      subtitleOverlay.setFontSizeFactor(3.0);
-      expect(subtitleOverlay['FONT_SIZE_FACTOR']).toBe(2.0); // Should be clamped to 2.0
-
-      // Set a valid value within range
-      subtitleOverlay.setFontSizeFactor(1.5);
-      expect(subtitleOverlay['FONT_SIZE_FACTOR']).toBe(1.5);
+    // Font size factor clamping
+    test.each([
+      [0.2, 0.5],  // Clamped to minimum
+      [3.0, 2.0],  // Clamped to maximum
+      [1.5, 1.5],  // Within range, no clamping
+      [0.5, 0.5],  // Exact minimum
+      [2.0, 2.0],  // Exact maximum
+      [1.0, 1.0],  // Default value
+    ])('setFontSizeFactor(%f) results in FONT_SIZE_FACTOR = %f', (inputFactor, expectedFactor) => {
+      subtitleOverlay.setFontSizeFactor(inputFactor);
+      expect(subtitleOverlay['FONT_SIZE_FACTOR']).toBe(expectedFactor);
     });
 
-    it('recalculates CEA 608 grid values correctly based on font size factor', () => {
-      subtitleOverlay.setFontSizeFactor(1.0);
+    // Grid recalculations
+    test.each([
+      [1.0, 15 / 1.0, 32 / 1.0],  // Standard grid
+      [2.0, 15 / 2.0, 32 / 2.0],  // Larger factor, smaller grid
+      [0.5, 15 / 1.0, 32 / 0.5],  // Factor <1 → rows stay 15, columns grow
+    ])('setFontSizeFactor(%f) recalculates grid: ROWS = %f, COLUMNS = %f', (factor, expectedRows, expectedColumns) => {
+      subtitleOverlay.setFontSizeFactor(factor);
       subtitleOverlay.recalculateCEAGrid();
 
-      expect(subtitleOverlay['CEA608_NUM_ROWS']).toBe(15 / 1.0);
-      expect(subtitleOverlay['CEA608_NUM_COLUMNS']).toBe(32 / 1.0);
-
-      // Test with a different font size factor
-      subtitleOverlay.setFontSizeFactor(2.0);
-      subtitleOverlay.recalculateCEAGrid();
-
-      expect(subtitleOverlay['CEA608_NUM_ROWS']).toBe(15 / 2.0);
-      expect(subtitleOverlay['CEA608_NUM_COLUMNS']).toBe(32 / 2.0);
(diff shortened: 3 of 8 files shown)


gulpfile.js
```diff
@@ -103,7 +103,8 @@ gulp.task('lint-sass', function() {
   return gulp.src(paths.source.sass)
   .pipe(sassLint({
     rules: {
-      'no-css-comments': 0
+      'no-css-comments': 0,
+      'property-sort-order': 0
```
Andr3wid, line 107: Was this enabled by default?

src/scss/skin-modern/components/_subtitleoverlay-cea608.scss
```diff
@@ -1,6 +1,8 @@
 @use 'sass:math';
 
 .#{$prefix}-ui-subtitle-overlay {
+  --cea608-row-height: math.div(100%, 15);
```
Andr3wid, line 4: What's the resulting unit when we divide `%` with a scalar number?

spec/components/subtitleoverlay.spec.ts
```diff
@@ -78,4 +78,43 @@ describe('SubtitleOverlay', () => {
...
+
+      expect(subtitleOverlay['CEA608_NUM_ROWS']).toBe(15 / 1.0);
+      expect(subtitleOverlay['CEA608_NUM_COLUMNS']).toBe(32 / 1.0);
+
+      // Test with a different font size factor
+      subtitleOverlay.setFontSizeFactor(2.0);
+      subtitleOverlay.recalculateCEAGrid();
+
+      expect(subtitleOverlay['CEA608_NUM_ROWS']).toBe(15 / 2.0);
+      expect(subtitleOverlay['CEA608_NUM_COLUMNS']).toBe(32 / 2.0);
```
Andr3wid: Nit: Good candidate for `test.each`

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+- Changing font size now also takes an effect in CEA-608 captions with cap from 50% to 200%. Larger than default font size also disables the grid view.
```
Andr3wid, line 10: Nit, but was that actually a bug? If not, I'd add it to the `Changed` or `Added` section. If we keep it in `Fixed`, we might want to rephrase, e.g.:

```suggestion
- font size adjustments for CEA-608 captions not taking effect.
```

src/ts/components/subtitleoverlay.ts
```diff
@@ -34,12 +34,13 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+  private FONT_SIZE_FACTOR: number = 1;
   // The number of rows in a cea608 grid
-  private static readonly CEA608_NUM_ROWS = 15;
+  private CEA608_NUM_ROWS = 15 / Math.max(this.FONT_SIZE_FACTOR, 1);
   // The number of columns in a cea608 grid
-  private static readonly CEA608_NUM_COLUMNS = 32;
+  private CEA608_NUM_COLUMNS = 32 / this.FONT_SIZE_FACTOR;
   // The offset in percent for one column (which is also the width of a column)
-  private static readonly CEA608_COLUMN_OFFSET = 100 / SubtitleOverlay.CEA608_NUM_COLUMNS;
+  private CEA608_COLUMN_OFFSET = 100 / this.CEA608_NUM_COLUMNS;
```
Andr3wid: Nit: We might want to move these into the constructor by calling `recalculateCEAGrid` directly.

Andr3wid (commented):
Looking good generally - just some nits :)

src/scss/skin-modern/components/_subtitleoverlay-cea608.scss
Andr3wid: Discussed in call: `%` per caption row

Andr3wid (commented):
As discussed in the call, let's try to update the active subtitle's font-size as well 👍

CHANGELOG.md
mimcz: I mean its considered as bug but we can also put it as Added as its new behaviour.

Andr3wid (approved):

src/ts/components/subtitleoverlay.ts
```diff
@@ -151,6 +154,21 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+    this.FONT_SIZE_FACTOR = Math.max(0.5, Math.min(2.0, factor));;
+
+    this.recalculateCEAGrid();
+  }
+
+  recalculateCEAGrid() {
+    // Needs to get recalculated in case the font size will change
+    this.CEA608_NUM_ROWS = 15 / Math.max(this.FONT_SIZE_FACTOR, 1);
+    this.CEA608_NUM_COLUMNS = 32 / this.FONT_SIZE_FACTOR;
+    this.CEA608_COLUMN_OFFSET = 100 / this.CEA608_NUM_COLUMNS;
```
dweinber: Shouldn't we keep the `15`/`32` (and maybe `100`) as readonly (static) variables instead of using magic numbers here?

src/ts/components/subtitleoverlay.ts
```diff
@@ -151,6 +154,21 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+    // https://www.ecfr.gov/current/title-47/part-79/section-79.103#p-79.103(c)(4)
+    this.FONT_SIZE_FACTOR = Math.max(0.5, Math.min(2.0, factor));;
+
+    this.recalculateCEAGrid();
+  }
+
+  recalculateCEAGrid() {
+    // Needs to get recalculated in case the font size will change
+    this.CEA608_NUM_ROWS = 15 / Math.max(this.FONT_SIZE_FACTOR, 1);
+    this.CEA608_NUM_COLUMNS = 32 / this.FONT_SIZE_FACTOR;
```
dweinber: Not sure if we should use `Math.max(this.FONT_SIZE_FACTOR, 1)` here as well. Any thoughts?

src/ts/components/subtitleoverlay.ts
```diff
@@ -228,6 +246,19 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
     // Flag telling if the CEA-608 mode is enabled
     let enabled = false;
 
+    const settingsManager = uimanager.getSubtitleSettingsManager();
+
+    settingsManager.fontSize.onChanged.subscribe((_sender, property) => {
+      if (property.isSet()) {
+        // We need to convert from percent
+        const factorValue = parseInt(property.value) / 100;
```
dweinber: Does `property.value` need to be an integer? One could change the available options for `fontSize` to floating point values and we would ignore it here. I think the difference might be minimal (probably even on big TVs, not sure) but wanted to point it out nevertheless.

dweinber (changes requested):
This part looks good so far, just a few nit picks on the code.

What I found during testing: 

1. If you have background/window colors set, and you change the font size while captions are being displayed, the font will change but the background/window is not immediately resized. Starting with the next captions, it will be (usually/mostly) correct again. Can this be updated right away as well please?

2. The font size options in the subtitle customization settings still shows up to 400%, but anything above 200% has no effect. While we agreed to cap the size to 200%, it seems odd that 300% and 400% are available but don't change anything. This looks buggy from a user perspective. I'd suggest removing the two options that don't have any effect.

src/ts/components/subtitleoverlay.ts
mimcz: @dweinber What we have now is that it always parses percentage values so it would end up always as an integer. Also the normal subtitles styles and font size options are done the way they convert percent integers into css `em` units so I don't think there is a case for floating points here.

src/ts/components/subtitleoverlay.ts
mimcz: @dweinber I dont think we need to, if the font size factor is larger than 1 we would disable the grid anyway and for smaller font size more columns is ok.

src/ts/components/subtitleoverlay.ts
mimcz: @dweinber These properties will get eventually recalculated so they can't be static, but if you are pointing to magic numbers I can extract the values to constants and take it from there 👍

mimcz:
@dweinber Both mentioned issues with bg color and filtering the redundant font size options should now be addressed.

dweinber:
When a cue is visible, and a different font size was already saved in the localStorage (or pre-set in the code of the page/app I guess), the font size starts in the default size and changes to the correct one only once the Subtitle Settings Panel Page in the menu is opened. See attached screen recording:

https://github.com/user-attachments/assets/16973fe4-bc94-4466-b5f1-00078acd7e0c

Additionally, as also shown in the recording, the subtitle window size (height) is not adjusted either when the font size changes.

src/ts/components/subtitleoverlay.ts
dweinber: I'm not sure why percentage values (or values with `em` unit) can't be or won't be floating points, but anyway, fine for me. As said, I think the impact might be low or even non-noticable.

src/ts/components/subtitleoverlay.ts
dweinber: Yes, I get that these are re-calculated and can't be static, was just pointing to the magic numbers

dweinber (approved):
Looks good now, thanks!

PR #693, opened 2025-04-02
[PW-22833] Fix missing CEA-608 captions rows

mimcz:
## Description

- If cues are positioned on the bottom and the larger font size gets applied, the row position is not recalculated and can get hidden based on the precompiled row styles.
- We now recalculate the rows and apply the offset in case the cues are positioned above the available grid.
- We also need to store the original row position from the cue. 

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -162,14 +162,15 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
   setFontSizeFactor(factor: number): void {
     // We only allow range from 50% to 200% as suggested by spec
     // https://www.ecfr.gov/current/title-47/part-79/section-79.103#p-79.103(c)(4)
-    this.FONT_SIZE_FACTOR = Math.max(0.5, Math.min(2.0, factor));;
+    this.FONT_SIZE_FACTOR = Math.max(0.5, Math.min(2.0, factor));
 
     this.recalculateCEAGrid();
   }
 
   recalculateCEAGrid() {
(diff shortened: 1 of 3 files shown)


src/ts/components/subtitleoverlay.ts
```diff
@@ -162,14 +162,15 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
     this.recalculateCEAGrid();
   }
 
   recalculateCEAGrid() {
-    // Needs to get recalculated in case the font size will change
-    this.CEA608_NUM_ROWS = SubtitleOverlay.DEFAULT_CEA608_NUM_ROWS / Math.max(this.FONT_SIZE_FACTOR, 1);
+    // Needs to get recalculated in case the font size will change also we need to floor this
+    // to always align to the whole number represented in styles.
+    this.CEA608_NUM_ROWS = Math.floor(SubtitleOverlay.DEFAULT_CEA608_NUM_ROWS / Math.max(this.FONT_SIZE_FACTOR, 1));
     this.CEA608_NUM_COLUMNS = SubtitleOverlay.DEFAULT_CEA608_NUM_COLUMNS / this.FONT_SIZE_FACTOR;
```
dweinber: Shouldn't we use `Math.floor` here, too? Maybe not as big an issue as with rows, but I guess the number of columns should still be always an integer 😄

dweinber (approved):
Seems to work now, thanks! 👍

Andr3wid (approved):
Agree with Daniel's nit, looks good otherwise 👍

PR #694, opened 2025-04-07
Text to speech improvements

dweinber:
## Description
The previous implementation of the ARIA spec in the player UI works well on desktop, but has problems on e.g. Samsung Tizen TVs and their integrated Talk Back / Text to Speech functionality.

This PR improves TV support by providing the option to use on/off ARIA labels instead of using the [aria-pressed](https://www.w3.org/TR/wai-aria/#aria-pressed) attribute, as [Tizen doesn't support `pressed`](https://developer.samsung.com/smarttv/develop/guides/fundamentals/text-to-speech.html#ARIA-State-and-Property-Attribute-Support). Additionally, `aria-pressed` is not used automatically for all buttons anymore but is (optionally, if no on/off labels are provided) used for ToggleButtons.

[Tizen does also not support](https://developer.samsung.com/smarttv/develop/guides/fundamentals/text-to-speech.html#ARIA-State-and-Property-Attribute-Support) `aria-valuemax`, `aria-valuemin`, `aria-valuenow` and `aria-valuetext`, which are primarily used for the seekbar and similar sliders. 
This PR adds an optional configuration option to set the current playback time also on the `aria-label` (in addition to the element name).

The `ErrorMessageOverlay` didn't contain any ARIA attribute with the Error, which is added in this PR as well.

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/seekbar.ts
+++ b/src/ts/components/seekbar.ts
@@ -60,6 +60,13 @@ export interface SeekBarConfig extends ComponentConfig {
    * Used to enable/disable seek preview
    */
   enableSeekPreview?: boolean;
+
+  /**
+   * Enables a fallback mode for handling ARIA details of the seekbar for platforms that don't support the 
+   * `aria-valuetext` attribute, such as Samsung Tizen. In this mode, the `aria-label` attribute is used in addition
+   * to provide the current time.
+   */
+  ariaFallbackMode?: boolean;
 }
 
 /**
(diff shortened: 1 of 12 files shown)


src/ts/components/seekbar.ts
```diff
@@ -60,6 +60,13 @@ export interface SeekBarConfig extends ComponentConfig {
    * Used to enable/disable seek preview
    */
   enableSeekPreview?: boolean;
+
+  /**
+   * Enables a fallback mode for handling ARIA details of the seekbar for platforms that don't support the 
+   * `aria-valuetext` attribute, such as Samsung Tizen. In this mode, the `aria-label` attribute is used in addition
+   * to provide the current time.
```
felix-hoc: nit: it's not 100% clear IMO, maybe we could reprase it like this?
```suggestion
   * `aria-valuetext` attribute, such as Samsung Tizen. In this mode, the `aria-label` attribute is used to also
   * provide the current time.
```

src/ts/components/seekbar.ts
```diff
@@ -60,6 +60,13 @@ export interface SeekBarConfig extends ComponentConfig {
...
+   */
+  ariaFallbackMode?: boolean;
```
felix-hoc: The variable name kind of sounds like you could configure a `mode`, but it's only a flag - I would suggest to name it `enableAriaFallbackMode` or so

src/ts/stringutils.ts
```diff
@@ -36,7 +36,7 @@ export namespace StringUtils {
         .replace('ss', leftPadWithZeros(seconds, 2));
   }
 
-  export function secondsToText(totalSeconds: number): string {
+  export function secondsToText(totalSeconds: number, shouldLeftPadNumbers = true): string {
```
felix-hoc: Having a default padding of 2 feels a bit weird.

Since we only use this value in one place (seekbar.ts) I'd recommend to replace the `shouldLeftPadNumbers` with an attribute that specifies the left pad length, and make it either mandatory or default to `0`.

e.g.
```ts
  export function secondsToText(totalSeconds: number, leftPadLength = 0): string {
```
or
```ts
  export function secondsToText(totalSeconds: number, leftPadLength: number): string {
```

felix-hoc (changes requested):
Just some smaller comments, looks good in general 🙌🏼

src/ts/components/seekbar.ts
dweinber: I've made an attempt to clarify this in 919bce2423100e248459fc7694f9201930ce6a59. Let me know what you think.

src/ts/components/seekbar.ts
dweinber: I've renamed it to 919bce2423100e248459fc7694f9201930ce6a59, hopefully this makes it clearer.

src/ts/stringutils.ts
dweinber: I didn't want to change the default. The default is always to have two-digit time strings, such as "06" seconds or "25" minutes etc.
But agree, it might make more sense to provide an actual number as parameter.
I'll update this.

dweinber: Updated in a9594dd75777c648101580700787b80ac9a2c2dc

dweinber: As discussed offline, I've removed the left pad option in 382b2270d97ac753c2124e7f7054b179b5b52441

felix-hoc (approved):
cool stuff 🙌🏼

PR #698, opened 2025-04-22
Restyle seekbar chapter markers

Andr3wid:
## Description
Apply styles to seekbar chapter markers to match the new v4 UI design.

**Before:**
image

**After**:
image

- [x] `CHANGELOG` entry


src/scss/components/seekbar/_seek-bar.scss
```diff
@@ -105,11 +108,30 @@ $seekbar-height: .3125em;
...
         > .#{$prefix}-seekbar-marker-image {
           height: $seekbar-height * 4;
           position: absolute;
           transform: translate(-50%, calc(-100% - .2em));
         }
+
+        > .#{$prefix}-seekbar-marker-indicator {
+          height: $marker-indicator-height;
+          width: $marker-width;
+          background-color: $color-primary;
```
bitmovin-kenny, line 131: Are we sure about this color?
If you are playing before a range, you don't really see the buffer indicator anymore

image

bitmovin-kenny (commented):
Code LGTM, not sure about the color of the range though since it masks the buffer level.

src/scss/components/seekbar/_seek-bar.scss
Andr3wid: Just X-checked, and it seems that's already the way we currently handle it as well:

image

Andr3wid: Just X-checked, that seems to be the case already

image

bitmovin-kenny: Well ok then

bitmovin-kenny (approved):
I would still like some opacity on the region marker in order for the buffer indicator to still be visible, but since this is the way it currently is ... 👍

123mpozzi:
Seekbar colors changes will be handled in #719

PR #700, opened 2025-05-07
Make `SettingsPanel` generic

stonko1994:
## Description

We want to leverage the usage of `ListBox`es in v4. Especially for the TV UI or wherever we put subtitle / audio track selection to the root UI.

The `ListBox` was originally implemented as an alternative to the native select boxes for exactly that use-case. For v4 we already migrated away from using the select boxes to use a more modern settings panel approach. This approach (and especially its styling) can now be re-used for `ListBox`es. To enable the inheritance from `SettingsPanel`, we need to make it generic first.

This PR prepares that by introducing a `SettingsPanelConfig` and making the `SettingsPanel` generic.

- [x] `CHANGELOG` entry **A final v4 changleog entry will be added in a separate PR**

### DIFF
--- a/src/ts/components/settings/SettingsPanel.ts
+++ b/src/ts/components/settings/SettingsPanel.ts
@@ -60,7 +60,7 @@ export enum NavigationDirection {
  *
  * @category Components
  */
-export class SettingsPanel extends Container<SettingsPanelConfig> {
+export class SettingsPanel<Config extends SettingsPanelConfig> extends Container<Config> {
 
   private static readonly CLASS_ACTIVE_PAGE = 'active';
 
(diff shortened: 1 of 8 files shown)


src/ts/components/settings/SettingsPanel.ts
```diff
@@ -60,7 +60,7 @@ export enum NavigationDirection {
  *
  * @category Components
  */
-export class SettingsPanel extends Container<SettingsPanelConfig> {
+export class SettingsPanel<Config extends SettingsPanelConfig> extends Container<Config> {
```
felix-hoc, line 63: nit: I feel like `Config` does not indicate that the type is generic. I think convention would be something like `TConfig`, or a single character, like `C`. But it should also fit with the generic naming of the `Container` class, so feel free to ignore

felix-hoc (approved):
just a nit, otherwise LGTM

src/ts/components/settings/SettingsPanel.ts
stonko1994: Agree but it's arealdy called `Config` everywhere in the code base.

PR #701, opened 2025-05-07
Refactor `ListBox` to extend from `SettingsPanel`

stonko1994:
## Description

We want to leverage the usage of `ListBoxe`s in v4. Especially for the TV UI or wherever we put subtitle / audio track selection to the root UI.

The `ListBox` was originally implemented as an alternative to the native select boxes for exactly that use-case. For v4 we already migrated away from using the select boxes to use a more modern settings panel approach. This approach (and especially its styling) can now be re-used for ListBoxes. To enable the inheritance from SettingsPanel, we need to make it generic first.

This PR changes the inheritance for the `ListBox` to leverage the `SettingsPanel` behaviour and styling. There were some additional changes needed which are hightlighed in detail by comments.

You can use the following patch for testing such a `ListBox`:
[Uploading subtitle-list-box-testing.patch…]()

- [x] `CHANGELOG` entry **There will be a separate PR for a v4 changelog entry**


src/scss/components/lists/_list-box.scss
```diff
@@ -1,38 +0,0 @@
-@import '../../variables';
-@import '../../mixins';
-
-%ui-listbox {
```
stonko1994, line 4: This is no longer needed as a `ListBox` now fully uses the `SettingsPanel` styling

src/ts/components/lists/ListBox.ts
```diff
@@ -1,149 +1,89 @@
...
- *     <button class='ui-listbox-button'>label</button>
- *     ...
- *   </div
- * </code>
- *
- * @category Components
- */
-// TODO: change ListSelector to extends container in v4 to improve usage of ListBox.
-//       Currently we are creating the dom element of the list box with child elements manually here.
-//       But this functionality is already covered within the Container component.
```
stonko1994, line 23: 😏

src/ts/components/settings/SettingsPanelItem.ts
```diff
@@ -29,6 +29,11 @@ export interface SettingsPanelItemConfig extends ContainerConfig {
    * If the setting should be added as a component to this item.
    */
   addSettingAsComponent?: boolean;
+  /**
+   * Indicates if the SettingsPanelItem is representing an actual setting or if it just a label. E.g. when used
+   * as a title.
+   */
+  isSetting?: boolean;
```
stonko1994, line 36: This indicator is needed as we have the feature to automatically hide `SettingsToggleButton`s (the buttons which open a settings panel) if there is no active setting in there. This is now also working for `ListBox`es.

However, since we sometimes add a title `SettingsPanelItem`, we need an indicator to not count them as active settings. To distinguish them, I added this flag.

src/ts/components/settings/SettingsPanelPage.ts
```diff
@@ -114,4 +114,20 @@ export class SettingsPanelPage extends Container<SettingsPanelPageConfig> {
...
+    this.updateComponents();
+  }
+
+  /**
+   * Dynamically remove a settings panel item from the page.
+   */
+  removeSettingsPanelItem(settingsPanelItem: SettingsPanelItem<SettingsPanelItemConfig>) {
+    this.removeComponent(settingsPanelItem);
+    this.updateComponents();
+  }
```
stonko1994, line 132: This helper functions were needed as the `updateComponents` method is not exposed publicly. It also adds a proper API to do dynamic changes to a `SettingsPanelPage`.

src/ts/utils/SubtitleUtils.ts
```diff
@@ -87,6 +87,8 @@ export class SubtitleSwitchHandler {
     if (this.listElement.hasItem(subtitle.id)) {
       this.listElement.removeItem(subtitle.id);
     }
+
+    this.selectCurrentSubtitle();
```
stonko1994, line 91: This is a minor fix which is needed since no select boxes are used anymore (the problem exists on v2 as well).

In case the active SubtitleTrack gets removed we need to update to the now selected subtitle track (usually `off`)

src/ts/spatialnavigation/TypeGuards.ts
```diff
@@ -16,10 +15,6 @@ export function isContainer(obj: unknown): obj is Container<unknown> {
   return obj !== null && obj !== undefined && obj instanceof Container;
 }
 
-export function isListBox(obj: unknown): obj is ListBox {
-  return obj instanceof ListBox;
-}
-
```
stonko1994, line 22: This special handling for a `ListBox` is no longer needed as it is already covered by the `SettingsPanel` support.

src/ts/components/settings/SettingsPanelItem.ts
wasp898: I suppose this can't be inferred from the existing properties?

wasp898 (approved):

src/ts/components/settings/SettingsPanelItem.ts
stonko1994: No I couldn't think of any other indicator as it can be anything.

PR #705, opened 2025-05-13
WebVTT doesn't render when region styles are present 

SInCE:
## Description
When WebVTT subtitles with a STYLES definition are used, the subtitle CSS `position:static` property, which affects rendering

- [x] `CHANGELOG` entry


stonko1994 (changes requested):
The current solution breaks CEA caption rendering. I don't know yet why and what could be an alternative but I'm looking into it.

hawk23:
@stonko1994 Could you have another look at the solution and do maybe some quick manual testing? I tested with VTT and CEA streams and it's looking good.

stonko1994 (approved):
Looks good now 👍

hawk23 (approved):

PR #706, opened 2025-05-13
Make subtitle window fully enclose CEA captions

jmsn:
### Problem

The subtitle window of CEA captions does not enclose the caption on all sides.

To demonstrate this, in below example there are two captions with green background and blue subtitle window.

Screenshot 2025-05-13 at 15 56 12

### Fix

Context: With CEA captions there is a display of 15 equal-height rows, i.e. there's a height allocated for each row.

- Introduce a ratio that defines what amount of the row height should be used for the caption window, effectively a padding at top and bottom. E.g. a ratio of `0.2` means 20% of row height will be padding, 10% at top and bottom, respectively. This also means that the actual cue text has less space, i.e. `0.8` of the row height in our example.
- Ensure that the cues do not stick to the left side of the window but rather leave a little space to let some of the window through. As this must scale with Player resizes and Font changes, this will now be set to 0.5% of the width. Calculating the offset in JS did not work out well, since the width of the component can be 0 if it's not shown.

This results in the captions looking like this now:
image

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Changed
+- Background of CEA captions now does not cover the subtitle window fully
+
 ## [3.93.0] - 2025-05-09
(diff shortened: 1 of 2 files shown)


Andr3wid (approved):
LGTM! Seems to work as expected now :)

PR #712, opened 2025-07-17
Modern UI - Add TitleBar in Ads UI layout

123mpozzi:
## Description

image

Add the TitleBar to the **generic**\* ads screen, following the planned UI design.

It will fetch metadata for linear ads from their `LinearAdUiConfig.message` field.  

> [!NOTE]
> \* Small-screen and TV-screen will, in case, be planned as separate PRs

> [!TIP]
> Metadata description is not currently set since there is no field in `LinearAdUiConfig` to use

## Testing

Check the new Ads TitleBar by adding a message to the `LinearAdUiConfig` of the `AdBreak`

```diff
  var adBreaks = [
    {
      tag: {
        url: '...',
        type: 'vast',
      },
      position: 'pre',
      linearAdUiConfig: {
+        message: "my lovely title",
      }
    },
    ...
```

- [x] `CHANGELOG` entry - `no need on this base branch it seems`

### DIFF
--- a/src/ts/UIFactory.ts
+++ b/src/ts/UIFactory.ts
@@ -376,6 +376,7 @@ function adsUILayout() {
             cssClasses: ['ui-titlebar-top'],
           }),
         ],
+        keepHiddenWithoutMetadata: true,
       }),
       new ErrorMessageOverlay(),
     ],
--- a/src/ts/UIFactory.ts
+++ b/src/ts/UIFactory.ts
@@ -367,7 +367,6 @@ function adsUILayout() {
       new AdStatusOverlay(),
       controlBar,
       new TitleBar({
-        // keepHiddenWithoutMetadata: true,
         components: [
           new Container({
             components: [
--- a/src/ts/UIFactory.ts
+++ b/src/ts/UIFactory.ts
@@ -50,6 +50,7 @@ import { DynamicSettingsPanelItem } from './components/settings/DynamicSettingsP
 import { TouchControlOverlay } from './components/overlays/TouchControlOverlay';
 import { AdStatusOverlay } from './components/ads/AdStatusOverlay';
 import { DismissClickOverlay } from './components/overlays/DismissClickOverlay';
+import { AdMessageLabel } from './main';
 
 /**
  * Provides factory methods to create Bitmovin provided UIs.
@@ -370,7 +371,7 @@ function adsUILayout() {
         components: [
           new Container({
             components: [
-              new MetadataLabel({ content: MetadataLabelContent.AdMessage }),
+              new AdMessageLabel(),
             ],
             cssClasses: ['ui-titlebar-top'],
           }),
--- a/src/ts/UIFactory.ts
+++ b/src/ts/UIFactory.ts
@@ -366,6 +366,23 @@ function adsUILayout() {
       new PlaybackToggleOverlay(),
       new AdStatusOverlay(),
       controlBar,
+      new TitleBar({
+        // keepHiddenWithoutMetadata: true,
+        components: [
+          new Container({
+            components: [
+              new MetadataLabel({ content: MetadataLabelContent.Title }),
+            ],
+            cssClasses: ['ui-titlebar-top'],
+          }),
+          new Container({
+            components: [
+              new MetadataLabel({ content: MetadataLabelContent.Description }),
+            ],
+            cssClasses: ['ui-titlebar-bottom'],
+          }),
+        ],
+      }),
       new ErrorMessageOverlay(),
     ],
     hideDelay: 2000,
--- a/src/ts/UIFactory.ts
+++ b/src/ts/UIFactory.ts
@@ -370,15 +370,9 @@ function adsUILayout() {
         components: [
           new Container({
             components: [
-              new MetadataLabel({ content: MetadataLabelContent.Title }),
+              new MetadataLabel({ content: MetadataLabelContent.AdMessage }),
             ],
             cssClasses: ['ui-titlebar-top'],
-          }),
-          new Container({
-            components: [
-              new MetadataLabel({ content: MetadataLabelContent.Description }),
-            ],
-            cssClasses: ['ui-titlebar-bottom'],
           }),
         ],
       }),
--- a/src/ts/UIFactory.ts
+++ b/src/ts/UIFactory.ts
@@ -50,7 +50,7 @@ import { DynamicSettingsPanelItem } from './components/settings/DynamicSettingsP
 import { TouchControlOverlay } from './components/overlays/TouchControlOverlay';
 import { AdStatusOverlay } from './components/ads/AdStatusOverlay';
 import { DismissClickOverlay } from './components/overlays/DismissClickOverlay';
-import { AdMessageLabel } from './main';
+import { AdMessageLabel } from './components/ads/AdMessageLabel';
 
 /**
  * Provides factory methods to create Bitmovin provided UIs.
--- a/src/ts/components/ads/AdMessageLabel.ts
+++ b/src/ts/components/ads/AdMessageLabel.ts
@@ -12,7 +12,7 @@ export class AdMessageLabel extends Label<LabelConfig> {
     super(config);
 
     this.config = this.mergeConfig(config, {
-      cssClasses: ['label-metadata', 'label-metadata-title'],
+      cssClass: 'ui-ad-message-label',
     }, this.config);
   }
 
--- a/src/ts/components/ads/AdMessageLabel.ts
+++ b/src/ts/components/ads/AdMessageLabel.ts
@@ -26,6 +26,8 @@ export class AdMessageLabel extends Label<LabelConfig> {
     clearText();
 
     player.on(player.exports.PlayerEvent.SourceUnloaded, clearText);
+    player.on(player.exports.PlayerEvent.AdError, clearText);
+    player.on(player.exports.PlayerEvent.AdSkipped, clearText);
     player.on(player.exports.PlayerEvent.AdFinished, clearText);
     player.on(player.exports.PlayerEvent.AdStarted, (event) => {
       const ad = (event as AdEvent).ad;
--- a/src/ts/components/ads/AdMessageLabel.ts
+++ b/src/ts/components/ads/AdMessageLabel.ts
@@ -1,51 +1,50 @@
-import {Label, LabelConfig} from '../labels/Label';
 import {UIInstanceManager} from '../../UIManager';
-import {StringUtils} from '../../utils/StringUtils';
 import { AdEvent, LinearAd, PlayerAPI } from 'bitmovin-player';
-import { i18n } from '../../localization/i18n';
+import { Label, LabelConfig } from '../labels/Label';
 
 /**
- * A label that displays a message about a running ad, optionally with a countdown.
+ * A label that displays a message regarding the ad that's currently being played.
  *
- * @category Components
+ * @category Labels
  */
 export class AdMessageLabel extends Label<LabelConfig> {
+  private adMessage: string = '';
 
   constructor(config: LabelConfig = {}) {
     super(config);
 
     this.config = this.mergeConfig(config, {
-      cssClass: 'ui-label-ad-message',
-      text: i18n.getLocalizer('ads.remainingTime') ,
+      cssClasses: ['label-metadata', 'label-metadata-title'],
     }, this.config);
   }
 
   configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
     super.configure(player, uimanager);
 
-    let config = this.getConfig();
-    let text = config.text;
-
-    let updateMessageHandler = () => {
-      this.setText(StringUtils.replaceAdMessagePlaceholders(i18n.performLocalization(text), null, player));
+    const init = () => {
+      this.setText(this.adMessage);
     };
 
-    let adStartHandler = (event: AdEvent) => {
-      let uiConfig = (event.ad as LinearAd).uiConfig;
-      text = uiConfig && uiConfig.message || config.text;
-
-      updateMessageHandler();
-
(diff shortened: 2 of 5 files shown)


src/ts/components/labels/MetadataLabel.ts
```diff
@@ -64,11 +67,34 @@ export class MetadataLabel extends Label<MetadataLabelConfig> {
...
+      uiconfig.metadata.description = mainContentDescription;
+      init();
+    };
+
     // Init label
     init();
     // Clear labels when source is unloaded
     player.on(player.exports.PlayerEvent.SourceUnloaded, unload);
 
+    player.on(player.exports.PlayerEvent.AdBreakStarted, () => {
```
stonko1994: IMO, adding the advertising title feature to the `MetadataLabel` doesn't feel right. Different visibilities between main content and ad content are usually covered by different `UIVariants`. There will never be the same `MetadataLabel` instance visible for main-content and ad-content playback. Adding this here means the `MetadataLabel` will change its content even though not even visible.
In addition, this way there is no way to use a `MetadataLabel` with the `uiConfig` information during ad playback as well (if a customer desires it).

That said, what I do see as a possibility is to introduce a new `MetadataLabelContent` enum case indicating that it will display the ad-message (e.g. `AdMessage`) and handling it here accordingly.

src/ts/components/labels/MetadataLabel.ts
```diff
@@ -64,11 +67,34 @@ export class MetadataLabel extends Label<MetadataLabelConfig> {
...
+    });
+    player.on(player.exports.PlayerEvent.AdStarted, (event) => {
+      const ad = (event as AdEvent).ad;
+      if (!ad.isLinear) {
+        return;
+      }
+
+      const linearAd = ad as LinearAd;
+      uiconfig.metadata.title = linearAd.uiConfig?.message ?? '';
+      uiconfig.metadata.description = '';
```
stonko1994: Changing the user-configured `uiconfig` is a no-go and we should never do that. The `uiconfig` is an input from the customer and should stay consistent throughout the lifecycle of the UI.

In this implementation, the worst case could be that the `AdBreakFinished` is never received, and the `uiconfig` will stay in this Ad state.

stonko1994 (changes requested):

src/ts/components/labels/MetadataLabel.ts
123mpozzi: I see, it makes sense to separate the ad title feature from `MetadataLabel`'s main logic.
I went with the new `AdMessage` enum, thanks for the suggestion! 🙌

Address in 658a3252edc7e4288d4dbfcc2bd7d841007c2992

src/ts/components/labels/MetadataLabel.ts
123mpozzi: Thanks for the comprehensive comment, I've addressed this in 658a3252edc7e4288d4dbfcc2bd7d841007c2992 by not going through `uiconfig` anymore

src/ts/components/labels/MetadataLabel.ts
```diff
@@ -34,6 +38,7 @@ export interface MetadataLabelConfig extends LabelConfig {
  * @category Labels
  */
 export class MetadataLabel extends Label<MetadataLabelConfig> {
+  private adMessage: string = '';
```
stonko1994: This component should be generic and therefore shouldn't have any ad or non-ad related instance properties. It it possible to get rid of this?

src/ts/components/labels/MetadataLabel.ts
```diff
@@ -57,18 +62,33 @@ export class MetadataLabel extends Label<MetadataLabelConfig> {
...
+    player.on(player.exports.PlayerEvent.AdStarted, (event) => {
+      const ad = (event as AdEvent).ad;
+      if (!ad.isLinear) {
+        return;
+      }
+
+      const linearAd = ad as LinearAd;
+      this.adMessage = linearAd.uiConfig?.message ?? '';
+      init();
+    });
```
stonko1994: Following the previous comment can we do the following here?

```suggestion
    player.on(player.exports.PlayerEvent.AdStarted, (event) => {
      if (config.context != MetadataLabelContent.AdMessage) { return }
      
      const ad = (event as AdEvent).ad;
      if (!ad.isLinear) {
        return;
      }

      const linearAd = ad as LinearAd;
      this.setText(linearAd.uiConfig?.message ?? '';);
    });
```

src/ts/components/labels/MetadataLabel.ts
```diff
@@ -57,18 +62,33 @@ export class MetadataLabel extends Label<MetadataLabelConfig> {
...
       this.setText(null);
+      this.adMessage = '';
     };
 
     // Init label
     init();
     // Clear labels when source is unloaded
     player.on(player.exports.PlayerEvent.SourceUnloaded, unload);
 
+    player.on(player.exports.PlayerEvent.AdStarted, (event) => {
```
stonko1994: I think we should clean up the text in `AdFinished` event

stonko1994 (changes requested):
Based on all the ad related handling we add to the `MetadataLabel` I'm wondering if it would be better to introduce an `AdMessageLabel` which only handles the Ad related text and keep the existing component untouched 🤔

src/ts/components/labels/MetadataLabel.ts
stonko1994: Or: we could only subscribe the AdStarted even when the type is AdMessage. This was we don't need the early return at all.

123mpozzi:
> Based on all the ad related handling we add to the `MetadataLabel` I'm wondering if it would be better to introduce an `AdMessageLabel` which only handles the Ad related text and keep the existing component untouched 🤔

Agree. In e5b32fc7e069a1dd4923f3ffa84203c550a1a7f2 I have extracted the component into its own class - by reusing `AdMessageLabel` as we discussed on Slack

src/ts/components/labels/MetadataLabel.ts
123mpozzi: Address in e5b32fc7e069a1dd4923f3ffa84203c550a1a7f2

src/ts/components/labels/MetadataLabel.ts
123mpozzi: The new code has now been extracted into its own class, so this should not apply anymore from e5b32fc7e069a1dd4923f3ffa84203c550a1a7f2

src/ts/components/labels/MetadataLabel.ts
123mpozzi: The new code has now been extracted into its own class, so this should not apply anymore from e5b32fc7e069a1dd4923f3ffa84203c550a1a7f2

src/ts/components/ads/AdMessageLabel.ts
```diff
@@ -1,51 +1,50 @@
...
 
 /**
- * A label that displays a message about a running ad, optionally with a countdown.
+ * A label that displays a message regarding the ad that's currently being played.
  *
- * @category Components
+ * @category Labels
  */
 export class AdMessageLabel extends Label<LabelConfig> {
+  private adMessage: string = '';
```
stonko1994: Similar to before: Can we still get rid of this here? Or is there a reason to keep this here? A general rule of thumb is: less state handling is better state handling. I'd suggest to only use the text from the event and set it on the label intsead of here.

src/ts/components/ads/AdMessageLabel.ts
```diff
@@ -1,51 +1,50 @@
...
 
   constructor(config: LabelConfig = {}) {
     super(config);
 
     this.config = this.mergeConfig(config, {
-      cssClass: 'ui-label-ad-message',
-      text: i18n.getLocalizer('ads.remainingTime') ,
+      cssClasses: ['label-metadata', 'label-metadata-title'],
```
stonko1994: This is a mixup of converns now. In our UI there should be one SCSS file for one TS component (there are exceptions). So re-using the style of another component isn't something we usually do in the UI.

Please create a dedicated `ad-message-label.scss` files and put the styling there.

src/ts/components/ads/AdMessageLabel.ts
```diff
@@ -1,51 +1,50 @@
...
     super.configure(player, uimanager);
 
-    let config = this.getConfig();
-    let text = config.text;
-
-    let updateMessageHandler = () => {
-      this.setText(StringUtils.replaceAdMessagePlaceholders(i18n.performLocalization(text), null, player));
+    const init = () => {
+      this.setText(this.adMessage);
     };
```
stonko1994: No need to have this in a function.

src/ts/components/ads/AdMessageLabel.ts
```diff
@@ -1,51 +1,50 @@
...
+      if (!ad.isLinear) {
+        return;
+      }
 
-    let adEndHandler = () => {
-      player.off(player.exports.PlayerEvent.TimeChanged, updateMessageHandler);
-    };
+      const linearAd = ad as LinearAd;
+      this.adMessage = linearAd.uiConfig?.message ?? '';
+      init();
```
stonko1994: You can us `this.setText` directly here and get rid of the state variable.

src/ts/components/ads/AdMessageLabel.ts
```diff
@@ -1,51 +1,50 @@
...
+    });
 
-    player.on(player.exports.PlayerEvent.AdStarted, adStartHandler);
-    player.on(player.exports.PlayerEvent.AdSkipped, adEndHandler);
-    player.on(player.exports.PlayerEvent.AdError, adEndHandler);
-    player.on(player.exports.PlayerEvent.AdFinished, adEndHandler);
+    uimanager.getConfig().events.onUpdated.subscribe(init);
```
stonko1994: Why do we need toe `onUpdated` here? 🤔

src/ts/components/ads/AdMessageLabel.ts
```diff
@@ -1,51 +1,50 @@
...
-    let updateMessageHandler = () => {
-      this.setText(StringUtils.replaceAdMessagePlaceholders(i18n.performLocalization(text), null, player));
+    const init = () => {
+      this.setText(this.adMessage);
     };
 
-    let adStartHandler = (event: AdEvent) => {
-      let uiConfig = (event.ad as LinearAd).uiConfig;
-      text = uiConfig && uiConfig.message || config.text;
+    const unload = () => {
```
stonko1994: Since this is not only called from unload but also from other events please name it in a way describing what it does. 

```suggestion
    const clearAdMessageText = () => {
```

src/ts/components/ads/AdMessageLabel.ts
```diff
@@ -1,51 +1,50 @@
...
+      this.setText('');
+      this.adMessage = '';
+    };
 
-      updateMessageHandler();
+    init();
 
-      player.on(player.exports.PlayerEvent.TimeChanged, updateMessageHandler);
-    };
+    player.on(player.exports.PlayerEvent.SourceUnloaded, unload);
```
stonko1994: Please also call the clear method from an `AdError`.

stonko1994 (changes requested):

src/ts/components/ads/AdMessageLabel.ts
123mpozzi: I see, I was wondering whether it would have been better to re-use the properties of `label-metadata` since the style should always be connected.

Extracted into own styles in e0b1507fa16b338f3092818948bb4eed382c1822

src/ts/components/ads/AdMessageLabel.ts
123mpozzi: Ah I agree, the less state the better. Removed the var in dccaf9354c3e125aeecc955ae629f1b3e39abd22

src/ts/components/ads/AdMessageLabel.ts
123mpozzi: Address in dccaf9354c3e125aeecc955ae629f1b3e39abd22

src/ts/components/ads/AdMessageLabel.ts
123mpozzi: Address in dccaf9354c3e125aeecc955ae629f1b3e39abd22

src/ts/components/ads/AdMessageLabel.ts
123mpozzi: True, that's not necessary since that refers to the generic `UiConfig`, but here we use `LinearAdUiConfig` — removed in a09375455bdfa1eb1ccba02bbdf9f3adf5d8f480

src/ts/components/ads/AdMessageLabel.ts
123mpozzi: Address in dccaf9354c3e125aeecc955ae629f1b3e39abd22

src/ts/components/ads/AdMessageLabel.ts
123mpozzi: Good idea! Added text clearing on ad error and skipping in 65a4e0807ff41971bb83c18af604ef56412bc646

src/ts/UIFactory.ts
```diff
@@ -366,6 +367,16 @@ function adsUILayout() {
...
       controlBar,
+      new TitleBar({
+        components: [
+          new Container({
+            components: [
+              new AdMessageLabel(),
+            ],
+            cssClasses: ['ui-titlebar-top'],
+          }),
+        ],
```
stonko1994, line 379: Let's add `keepHiddenWithoutMetadata` to not show the shadow if there is no message. This way we overlay less of the ad which is better for ad-revenue.

```suggestion
        ],
        keepHiddenWithoutMetadata: true,
```

123mpozzi: Ah I see, thanks for the insight!

stonko1994 (approved):

PR #713, opened 2025-07-21
Modern UI - Ad Interface SeekBar Labels

123mpozzi:
## Description

image

Rework the labels around the SeekBar, following the planned ad screen UI re-design.

#### Ad Playback Info
- **Total ad duration** is removed - it was previously at the right side of the SeekBar
- **Current ad playback time** is removed
- **Remaining ad playback time** is added and placed to the right of the SeekBar

#### Ad Index Label
- Introduce a new label component to indicate the current ad index out of the total amount of ads
- It is placed on the left side of the SeekBar

- [ ] `CHANGELOG` entry - `no need since base is modern UI`

### DIFF
--- a/src/ts/components/ads/AdCounterLabel.ts
+++ b/src/ts/components/ads/AdCounterLabel.ts
@@ -19,21 +19,16 @@ export class AdCounterLabel extends Label<LabelConfig> {
   configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
     super.configure(player, uimanager);
 
-    let currentAdIndex: number = 0;
-    let totalAdsCount: number = 0;
-
-    const reset = () => {
-      currentAdIndex = 0;
-      totalAdsCount = 0;
+    const clearText = () => {
       this.setText('');
     };
 
     player.on(player.exports.PlayerEvent.AdStarted, () => {
-      currentAdIndex++;
-      totalAdsCount = player.ads.getActiveAdBreak().ads?.length ?? 0;
-      this.setText(`Ad ${currentAdIndex} of ${totalAdsCount}`);
+      const activeAdIndex = player.ads.getActiveAdBreak().ads.findIndex((ad) => ad === player.ads.getActiveAd()) + 1;
+      const totalAdsCount = player.ads.getActiveAdBreak().ads?.length ?? 0;
+      this.setText(`Ad ${activeAdIndex} of ${totalAdsCount}`);
     });
-    player.on(player.exports.PlayerEvent.AdBreakStarted, reset);
-    player.on(player.exports.PlayerEvent.AdBreakFinished, reset);
+    player.on(player.exports.PlayerEvent.AdBreakStarted, clearText);
+    player.on(player.exports.PlayerEvent.AdBreakFinished, clearText);
   }
 }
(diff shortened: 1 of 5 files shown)


src/ts/components/ads/AdCounterLabel.ts
```diff
@@ -0,0 +1,39 @@
+import { UIInstanceManager } from '../../UIManager';
+import {LabelConfig, Label} from '../labels/Label';
```
stonko1994:
```suggestion
import { LabelConfig, Label } from '../labels/Label';
```

src/ts/components/ads/AdCounterLabel.ts
```diff
@@ -0,0 +1,39 @@
...
+    const reset = () => {
+      currentAdIndex = 0;
+      totalAdsCount = 0;
+      this.setText('');
+    };
+
+    player.on(player.exports.PlayerEvent.AdStarted, () => {
+      currentAdIndex++;
+      totalAdsCount = player.ads.getActiveAdBreak().ads?.length ?? 0;
+      this.setText(`Ad ${currentAdIndex} of ${totalAdsCount}`);
```
stonko1994: Same as on the other PR: less state handling is better. In this case here, when we miss a reset in some events the `currentAdIndex` will just continue to grow. 

What about something like this: 🤔 
```suggestion
      const activeAdIndex = player.ads.getActiveAdBreak().ads.findIndex((ad) => ad === player.ads.getActiveAd()) + 1;
      const totalAdsCount = player.ads.getActiveAdBreak().ads?.length ?? 0;
      this.setText(`Ad ${activeAdIndex} of ${totalAdsCount}`);
```

src/ts/components/ads/AdCounterLabel.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    const reset = () => {
+      currentAdIndex = 0;
+      totalAdsCount = 0;
+      this.setText('');
+    };
+
+    player.on(player.exports.PlayerEvent.AdStarted, () => {
+      currentAdIndex++;
+      totalAdsCount = player.ads.getActiveAdBreak().ads?.length ?? 0;
```
stonko1994: Should we fallback to 1 instead of 0? 🤔 As the assumption would be that as soon as we get an AdStartedEvent we have at least 1 ad 🤔 Potentially even fallback to the `currentAdIndex` 🤔

stonko1994 (changes requested):

src/ts/components/ads/AdCounterLabel.ts
123mpozzi: Good point!
I like the idea of using the active ad index since that is overall more precise than an hardcoded `1`.

Address in 650cd4054903662df2c3f068bac032cb83adc23b

src/ts/components/ads/AdCounterLabel.ts
123mpozzi: Good idea, and thanks for the snippet 💪 

Address in 1abb1f672e065823193b34b43d6e755d74b1ce0b together with the state variables removal

stonko1994 (approved):

PR #714, opened 2025-07-21
Modern UI - Always show the seeking ControlBar during ad playback

123mpozzi:
## Description

image

> ControlBar bottom is hidden

With the ad screen UI re-design, we will always show the bottom part of the **control bar** (seek bar, ad duration info), together with the ad skip button, during the ad playback.

The ad skip button is already always visible, so this PR only handles the control bar.

### Changes
- Always show the ControlBar if an ad is playing
- Resume seek bar updates when ad starts
- Add a new hide animation mixin: it will make the element fade by sliding down
  - Use it for the hideable part of the ControlBar, its bottom part

- [x] `CHANGELOG` entry - `not needed since base is modern ui`


stonko1994 (commented):
The actual design idea was to only show the elements on the leve of the seekbar during ad playback. So to hide the `control-bar-bottom` elements but keep the `control-bar-top` elements.

123mpozzi:
> The actual design idea was to only show the elements on the leve of the seekbar during ad playback. So to hide the `control-bar-bottom` elements but keep the `control-bar-top` elements.

I see, I addressed this in the recent commits. Now only the bottom part gets hidden

stonko1994 (commented):
There seems to be some issues with the hiding 🤔 When testing it, it sometimes doesn't hide at all and when it hides it is not animated so it's pretty distracting.

123mpozzi:
> There seems to be some issues with the hiding 🤔 When testing it, it sometimes doesn't hide at all and when it hides it is not animated so it's pretty distracting.

I've addressed the hiding animation in recent commits:
- 18e68f23055809acf213009c2c42cc400e4a7012
- c47645ad98e4f13a769cfac3fd6a6503e39eefd7

It is a generic style and re-usable via mixins 🙂

123mpozzi:
> There seems to be some issues with the hiding 🤔 When testing it, it sometimes doesn't hide at all and when it hides it is not animated so it's pretty distracting.

Since this is a bug unrelated to the changes, but instead already present on `feature/modern-ui-base`, I'll handle it in a different PR.

Bug is: `controlsHide` event not being emitted in ad screen without a mouse enter/leave event being triggered at least once

stonko1994 (commented):
Screenshot 2025-07-31 at 15 16 39

There seems to be an issue that the seekbar isn't moving down once the control bar is gone. 🤔 

I think we need to implement a Javasript based animation approach here and eventually set `display: none` in css to achieve this.

123mpozzi:
> Screenshot 2025-07-31 at 15 16 39
> There seems to be an issue that the seekbar isn't moving down once the control bar is gone. 🤔
> 
> I think we need to implement a Javasript based animation approach here and eventually set `display: none` in css to achieve this.

Ah true.
Address it in the recent commits:
- dad2b151406a8469d193e08f4f12db7ce314dd14
- 0b13ec3b7cb2880f08d4e5860db592ceec516ce0

src/scss/_mixins.scss
stonko1994, line 1: Please check the linter issues in this file

src/scss/components/ads/_ad-control-bar-bottom.scss
stonko1994, line 1: Please check the linter issues in this file

src/ts/components/ads/AdControlBarBottom.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    uimanager.onControlsShow.subscribe(() => {
+      this.show();
+    });
+
+    uimanager.onControlsHide.subscribe(() => {
+      this.hide();
+    });
+  }
+}
```
stonko1994: Please add an empty line at EOF.

src/ts/components/seekbar/SeekBar.ts
stonko1994, line 1: I think those changes here in this file are not needed anymore or? 🤔

src/ts/components/ads/AdControlBarBottom.ts
123mpozzi: Address in 7617fc20a11f21f881e5f429599097cd352a385c

src/ts/components/seekbar/SeekBar.ts
123mpozzi: They are necessary, otherwise the seekbar will not show progress

src/scss/components/ads/_ad-control-bar-bottom.scss
123mpozzi: This warning isn’t easily solvable, I actually looked into it in earlier commits (e.g. 3fd9c637eb73142acbec64c38d7b3744547855a2). Issue is it makes other warnings pop up.

Given the effort involved and that it’s just a linter *warning*, I personally don’t think it’s worth addressing right now

src/scss/_mixins.scss
123mpozzi: Address in b09b923df549201017de3506411dbb0bb1ee1b7b

src/scss/components/ads/_ad-control-bar-bottom.scss
```diff
@@ -0,0 +1,20 @@
...
+  .#{$prefix}-ad-controlbar-bottom {
+    @include hidden-slide-down;
+
+    white-space: nowrap; // Required for iOS 8.2 to avoid wrapped controlbar due to wrong size calculation
+
+    > .#{$prefix}-container-wrapper {
+      display: flex;
+      margin: .5em 0;
+
+      .#{$prefix}-ui-volumeslider {
```
stonko1994, line 17: Please fix or silence the linter warning with a meaningfull message.

123mpozzi: 7570139eccfca95540318359c837dffea781d5a6

stonko1994 (approved):

PR #715, opened 2025-07-24
Modern UI -  Chapter Markers, redesign metadata labels

123mpozzi:
## Description

Redesign the chapter markers as planned

#### Before
image

#### After
image

**Biggest changes:**
- Move the text labels (title and time) out of the Thumbnail container
  - Place the title label on top of the Thumbnail
- Use an uniform background for the text labels

- [x] `CHANGELOG` entry - `no need since base is modern UI branch`

### DIFF
--- a/src/ts/components/seekbar/SeekBarLabel.ts
+++ b/src/ts/components/seekbar/SeekBarLabel.ts
@@ -41,18 +41,16 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
   constructor(config: SeekBarLabelConfig = {}) {
     super(config);
 
-    this.timeLabel = new Label({ cssClasses: ['seekbar-label-time'] });
-    this.titleLabel = new Label({ cssClasses: ['seekbar-label-title'] });
+    this.timeLabel = new Label({ cssClasses: ['seekbar-label-metadata', 'seekbar-label-time'] });
+    this.titleLabel = new Label({ cssClasses: ['seekbar-label-metadata', 'seekbar-label-title'] });
     this.thumbnail = new Component({ cssClasses: ['seekbar-thumbnail'], role: 'img' });
     this.thumbnailImageLoader = new ImageLoader();
 
     this.container = new Container({
       components: [
+        this.titleLabel,
         this.thumbnail,
-        new Container({
-          components: [this.titleLabel, this.timeLabel],
-          cssClass: 'seekbar-label-metadata',
-        }),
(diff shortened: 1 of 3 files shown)


src/ts/components/seekbar/SeekBarLabel.ts
```diff
@@ -76,7 +74,7 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
       this.timeFormat = Math.abs(player.isLive() ? player.getMaxTimeShift() : player.getDuration()) >= 3600 ?
         StringUtils.FORMAT_HHMMSS : StringUtils.FORMAT_MMSS;
       // Set initial state of title and thumbnail to handle sourceLoaded when switching to a live-stream
-      this.setTitleText(null);
+      this.setTitleText('');
```
stonko1994: can we stick to null here?

src/ts/components/seekbar/SeekBarLabel.ts
```diff
@@ -173,6 +171,12 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
    */
   setTitleText(text = '') {
```
stonko1994: I think we should change this to take an optional value and if it's `null` to hide the label. using the empty string kinda feels wrong to me 🤔 (It's not clear from the input)

stonko1994 (commented):
Oh it looks so sexy 😍 

The only thing I think we should still discuss is the range-marker design for mobile devices (and while scrubbing). Maybe we can somehow also thicken them when scrubbing over them (as this is the only interaction you have on touch input devices)

src/ts/components/seekbar/SeekBarLabel.ts
123mpozzi: Ok, address in 9e4862acbc5e902b37eec4688a25052d243b0324

src/ts/components/seekbar/SeekBarLabel.ts
123mpozzi: Address in 9e4862acbc5e902b37eec4688a25052d243b0324

stonko1994 (approved):

123mpozzi:
> Oh it looks so sexy 😍
> 
> The only thing I think we should still discuss is the range-marker design for mobile devices (and while scrubbing). Maybe we can somehow also thicken them when scrubbing over them (as this is the only interaction you have on touch input devices)

This has been handled in #717

PR #716, opened 2025-07-24
Modern UI - Make SeekBar endings round

123mpozzi:
## Description

**Before**
image

**After**
image

## Changes

- Make SeekBar endings round
- Fix ghost marker at position `0`: with markers metadata, an additional marker was always placed at position `0`


src/ts/components/seekbar/SeekBar.ts
```diff
@@ -737,9 +737,9 @@ export class SeekBar extends Component<SeekBarConfig> {
     this.seekBarMarkersContainer = seekBarChapterMarkersContainer;
 
     seekBarBarsContainer.append(this.seekBarBackdrop, this.seekBarBufferPosition, this.seekBarSeekPosition,
```
stonko1994: This line break seems weird. Can we move every parameter to a new line?

stonko1994 (commented):
Nice one 💪

src/ts/components/seekbar/SeekBar.ts
123mpozzi: I agree that consistent formatting helps readability, hopefully we can invest time in an automatic linter/formatter soon to handle these cases uniformly - would save us time in reviews.
For now, I’d consider this more of a nit, but happy to update it.

Address in 6b1a1cb64ed2c21a182bbc7e359bc5862db38197

stonko1994 (approved):

PR #717, opened 2025-07-27
Modern UI - Chapter Markers, thicken markers on scrub

123mpozzi:
## Description

https://github.com/bitmovin/bitmovin-player-ui/pull/715#pullrequestreview-3052374472

Thicken the chapter range-marker on scrubbing.

This especially improves UX on mobile devices since there is no hover event there, therefore it would not be possible to see the highlighter chapter marker without this change

### DIFF
--- a/src/scss/components/seekbar/_seek-bar.scss
+++ b/src/scss/components/seekbar/_seek-bar.scss
@@ -114,7 +114,7 @@ $seekbar-height: .3125em;
         justify-content: space-between;
         align-items: center;
 
-        &:hover {
+        &:hover, &.#{$prefix}-thicken {
           height: $marker-indicator-height;
           border-radius: $marker-indicator-border-radius;
         }
@@ -132,11 +132,6 @@ $seekbar-height: .3125em;
           border-radius: $marker-indicator-border-radius;
         }
 
-      }
-
-      .#{$prefix}-seekbar-marker.#{$prefix}-thicken {
-        height: $marker-indicator-height;
-        border-radius: $marker-indicator-border-radius;
       }
     }
   }
--- a/src/scss/components/seekbar/_seek-bar.scss
+++ b/src/scss/components/seekbar/_seek-bar.scss
@@ -114,7 +114,7 @@ $seekbar-height: .3125em;
         justify-content: space-between;
         align-items: center;
 
-        &:hover, &.#{$prefix}-thicken {
+        &:hover, &.#{$prefix}-hovered {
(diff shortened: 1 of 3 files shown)


src/scss/components/seekbar/_seek-bar.scss
```diff
@@ -133,6 +133,11 @@ $seekbar-height: .3125em;
         }
 
       }
+
+      .#{$prefix}-seekbar-marker.#{$prefix}-thicken {
+        height: $marker-indicator-height;
+        border-radius: $marker-indicator-border-radius;
+      }
```
stonko1994: Please merge this together with the `&:hover` selector above to reduce code duplication. You can write it like this:

```
&:hover, &.#{$prefix}-thicken {
```

src/ts/components/seekbar/SeekBar.ts
```diff
@@ -1067,6 +1067,18 @@ export class SeekBar extends Component<SeekBarConfig> {
     this.label.setPositionInBounds(pixelPosition, this.uiBoundingRect);
   };
 
+  private clearAllThickenedMarkers(): void {
+    this.seekBarMarkersContainer
+      .find(`.${this.prefixCss('seekbar-marker')}`)
+      ?.removeClass(this.prefixCss('thicken'));
```
stonko1994: Please extract the class name into a variable (with the prefix) or if not easily possible the string into a constant.

stonko1994: I think we shouldn't call the class `thicken`. The reason to do that is that customers could e.g. change the UI representation of a hovered marker. So instead of naming it after which effect it will have on the UI, we should rather name it which behaviour triggers this. In thise case this would be `hovered`.

123mpozzi: I see, address in ca826a976ae6b7f4b639f24eb04a6ee92adda599

123mpozzi: I looked around at the existing code, and it seems we often extract the vars in a constant starting with `CLASS_`, I've applied the same in 24133b06af4b0b4e74dcbe1bb0ed69424f03f2cd

src/scss/components/seekbar/_seek-bar.scss
123mpozzi: Address in d1d1782fc93e94aac9d3d61a3fb98f4b1c72da70

stonko1994 (approved):

PR #720, opened 2025-07-31
Modern UI -  Fix AdMessageLabel not working with empty metadata

123mpozzi:
## Description

`TitleBar.keepHiddenWithoutMetadata: true` makes `AdMessageLabel` hidden.
This is due to it checking only for `MetadataLabel`s.

Changes:
- `keepHiddenWithoutMetadata` supports any `Label` type now
- `TitleBar` does not provide a title and description by default anymore if there are custom components (like `AdMessageLabel`)
  - This prevents unexpected default component leakage in UI layouts where they are not needed

### DIFF
--- a/src/ts/components/TitleBar.ts
+++ b/src/ts/components/TitleBar.ts
@@ -85,7 +85,7 @@ export class TitleBar extends Container<TitleBarConfig> {
 
   private hasNonEmptyComponents(components: Component<ComponentConfig>[]): boolean {
     for (const component of components) {
-      if (component instanceof Label && !component.isEmpty()) {
+      if (hasIsEmpty(component) && !component.isEmpty()) {
         return true;
       }
 
@@ -109,4 +109,8 @@ export class TitleBar extends Container<TitleBarConfig> {
       }
     }
   }
+}
+
+function hasIsEmpty(obj: unknown): obj is { isEmpty: () => boolean } {
+  return typeof obj === 'object' && obj !== null && typeof (obj as any).isEmpty === 'function';
 }
--- a/src/ts/components/TitleBar.ts
+++ b/src/ts/components/TitleBar.ts
@@ -52,7 +52,7 @@ export class TitleBar extends Container<TitleBarConfig> {
     let hasVisibleText = true;
 
     const checkTextAndUpdateVisibility = () => {
-      hasVisibleText = this.scanLabelsForText(this.getComponents());
+      hasVisibleText = this.hasNonEmptyComponents(this.getComponents());
 
       if (this.isShown()) {
         if (config.keepHiddenWithoutMetadata && !hasVisibleText) {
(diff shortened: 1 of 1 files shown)


src/ts/components/TitleBar.ts
```diff
@@ -10,84 +12,101 @@ import { PlayerAPI } from 'bitmovin-player';
...
       shouldBeShown = false;
       this.hide();
     });
 
-    // init
-    checkMetadataTextAndUpdateVisibility();
+    checkTextAndUpdateVisibility();
+  }
+
+  private scanLabelsForText(components: Component<ComponentConfig>[]): boolean {
```
stonko1994: Nit: We very likely will have also non Label component in the title bar in v4. Especially on mobile.

Can we extend this now already to handle non Label components as well and don't hide the title bar in this case?


_Not blocking for this PR_

stonko1994: I think we should improve the naming of this. The name `scanlabelsForText` with a return value of type Boolean doesn't really indicate what the `Boolean` represents. Is it now `true` if there are labels with text or is it true if there are no labels with text.

+ in combination with the previous comment:
```suggestion
  private hasNonEmptyComponents(components: Component<ComponentConfig>[]): boolean {
```

stonko1994 (commented):
That's a very nice change. Two minor things.

src/ts/components/TitleBar.ts
123mpozzi: Address in a36a15f1b4bd67856658a4a086a25cee728198f7

123mpozzi: I see, then I'll use a generic type guard instead of checking the `Label` type. Address in dd652d98914ed50b2902b7719b18d9948f066adb

src/ts/components/TitleBar.ts
```diff
@@ -10,84 +12,105 @@ import { PlayerAPI } from 'bitmovin-player';
...
+      }
+
+      if (component instanceof Container) {
+        this.subscribeToTextChanges(component.getComponents(), onChange);
+      }
+    }
+  }
+}
+
+function hasIsEmpty(obj: unknown): obj is { isEmpty: () => boolean } {
```
stonko1994, line 114: 👀 can you explain what this type is? (I'm not super into Typescript but this looks like black magic 😅)

stonko1994 (approved):

src/ts/components/TitleBar.ts
stonko1994: This isn't what I was thinking about. You could have kept the check as it is. I was more thinking about the use-case when there is e.g. a Button inside the TitleBar. In the current implementation we would hide the title bar when the labels are empty even though there are buttons inside of it.

We shouldn't do that. But as mentioned no blocker for this PR.

src/ts/components/TitleBar.ts
123mpozzi: It matches any object that has an `isEmpty(): boolean` function

PR #724, opened 2025-08-01
Modern UI - Fix controls auto-hide logic does not start on UI Layout switch until there is a mouse movement

123mpozzi:
### Issue:
ControlsHide event is not automatically triggered on any Ads Layout interface UNTIL the user moves the mouse.

### Why it is happening:
Our UI logic (in `UIContainer`) only starts the hide-timer when `mousemove` or similar fires - since that's what we use to detect the user is there.

But when you switch to the ads layout, the mouse is already inside the player - so no `mousemove` event fires.

> [!NOTE]
> The bug is happening on **every** UI Layout switch. So from `generic -> ad`, but also `ad -> generic`.
> The proposed fix solves the issue for every scenario.

### The fix:
On UI switching (e.g. from generic Layout to Ad Layout), trigger a synthetic `mousemove` event.
This ensures that our UI logic sees user activity, and schedules the auto-hide timeout.


src/ts/UIManager.ts
```diff
@@ -461,7 +461,20 @@ export class UIManager {
...
+
+    // Trigger synthetic 'mousemove' to ensure UI show/hide logic is initialized.
+    // In UIContainer, we detect user activity with mouse events (e.g. `mousemove`)
+    // and start the auto-hide logic accordingly. However, when switching UI,
+    // the mouse can already be inside the player, so no natural `mousemove` event fires. 
+    const adUiContainer = this.currentUi.getUI();
+    const htmlElement = adUiContainer.getDomElement().get(0);
+    if (htmlElement) {
+      htmlElement.dispatchEvent(new MouseEvent('mousemove', { bubbles: true }));
+    }
```
stonko1994: Why did you decide to fake a mouse event insead of just directly calling the same method which is used within the event listener?

```suggestion
    adUiContainer.showUi();
```

stonko1994: Update: I just figured out that this is already happening in line 466 above.

src/ts/components/UIContainer.ts
```diff
@@ -99,13 +99,21 @@ export class UIContainer extends Container<UIContainerConfig> {
...
     let config = this.getConfig();
+    let isUiShown = false;
+
+    uimanager.onConfigured.subscribe(() => {
+      if (isUiShown) {
+        uimanager.onControlsShow.dispatch(this);
+      } else {
+        uimanager.onControlsHide.dispatch(this);
+      }
+    });
```
stonko1994, line 110: We never dispatched the initial ui controls state in case a `hideDelay` is configured. This led to the issue that hiding/showing the UI didn't work properly.

src/ts/components/ads/AdControlBar.ts
```diff
@@ -84,10 +84,6 @@ export class AdControlBar extends ControlBar {
   configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
     super.configure(player, uimanager);
 
-    player.on(player.exports.PlayerEvent.AdStarted, () => {
-      this.show();
-    });
-
```
stonko1994, line 90: We don't want to reveal the UI in case it is currently not visible so this is not needed.

PR #727, opened 2025-08-08
Immediately update the playback position with click based seeking

stonko1994:
### Problem
When seeking by clicking on the seekbar, the playback position doesn't update immediately, but only after the Seek is completed. This is particularly bad when the seek results in stalling.

### Changes
Immediately update the playback position to the seek target position. This now matches the behaviour of scrubbing on the SeekBar.

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/seekbar.ts
+++ b/src/ts/components/seekbar.ts
@@ -767,7 +767,6 @@ export class SeekBar extends Component<SeekBarConfig> {
       this.setSeekPosition(targetPercentage);
       this.setPlaybackPosition(targetPercentage);
 
-
       // Fire seeked event
       this.onSeekedEvent(targetPercentage);
     };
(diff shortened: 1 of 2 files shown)


src/ts/components/seekbar.ts
```diff
@@ -763,6 +763,11 @@ export class SeekBar extends Component<SeekBarConfig> {
       this.setSeeking(false);
       seeking = false;
 
+      // Update the UI in case we only have a click or touch
+      this.setSeekPosition(targetPercentage);
+      this.setPlaybackPosition(targetPercentage);
+
+
```
jmsn:
```suggestion
      // Update the UI in case we only have a click or touch
      this.setSeekPosition(targetPercentage);
      this.setPlaybackPosition(targetPercentage);

```

too generous, dude 😄

felix-hoc (approved):
nice UX improvement, particularly on slow network speed! 🙌🏼

PR #729, opened 2025-08-08
Prevent overlap between CEA captions and control bar

jmsn:
### Problem

CEA captions are rendered behind the player's control bar.

### Fix

This PR implements scaling the CEA grid down and moving it up when the control bar becomes visible, so that the two do not overlap.

Side fix: Fixed undesired word wraps with caption cues that use the full length of 32 characters

Before

Screenshot 2025-08-08 at 14 45 36

After

Screenshot 2025-08-08 at 14 44 28

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -1,14 +1,14 @@
-import { Container, ContainerConfig } from './container';
-import { UIInstanceManager } from '../uimanager';
-import { Label, LabelConfig } from './label';
-import { ComponentConfig, Component } from './component';
-import { ControlBar } from './controlbar';
-import { EventDispatcher } from '../eventdispatcher';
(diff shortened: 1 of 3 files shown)


src/ts/components/subtitleoverlay.ts
```diff
@@ -864,3 +880,25 @@ export class SubtitleRegionContainer extends Container<ContainerConfig> {
     return this.labelCount === 0;
   }
 }
+
+function isCea608SubtitleCue(cue: SubtitleCueEvent): boolean {
+  return cue.position != null;
+}
+
+function awaitTransitionEnd(domElement: DOM) {
```
stonko1994, line 888: I might take that for another feature I'm currently working on 🙌

stonko1994 (approved):

PR #730, opened 2025-08-12
Fix some subtitle and caption issues related to font size

jmsn:
## Description

Fixing the following issues:
- Caption font size change in subtitle rendering options menu has no effect for CEA captions (broken since v3.100.0)
- Subtitle font size changes for VTT/TTML subtitles when the subtitle rendering options menu is opened (broken since v3.89.0)
- Subtitle/caption font size preference from local storage is not restored upon UI initialization (broken since v3.89.0)

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/subtitleoverlay.ts
+++ b/src/ts/components/subtitleoverlay.ts
@@ -344,7 +344,9 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
 
     this.onShow?.subscribe(() => {
       // ensure CEA grid is updated whenever the overlay becomes visible
-      this.ensureCea608GridSizeUpdated();
+      if (this.cea608Enabled) {
+        this.ensureCea608GridSizeUpdated();
+      }
     });
 
     this.ensureCea608GridSizeUpdated = () => {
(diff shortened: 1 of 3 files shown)


src/ts/components/subtitleoverlay.ts
```diff
@@ -336,12 +336,17 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
       } else {
         this.setFontSizeFactor(1);
       }
-      this.ensureCea608GridSizeUpdated();
+
+      if (this.cea608Enabled) {
+        this.ensureCea608GridSizeUpdated();
```
123mpozzi, line 341: Nit: since all usages of `ensureCea608GridSizeUpdated()` check `cea608Enabled` first, we could put an early return in `ensureCea608GridSizeUpdated` itself. So that callers don’t need to remember to guard it.

123mpozzi (approved):
Looks good! And manual testing works fine
Thanks for fixing the local storage issue as well 🙌

src/ts/components/subtitleoverlay.ts
jmsn: I'll consider doing that in my next PR 😄

PR #731, opened 2025-08-12
Add icon component to buttons

stonko1994:
## Description

We current use background-image on buttons to display our icons within the UI. With this approach it's not easily possible to style the button and the icon differently. E.g. our new TV UI design will feature a new background color for focused buttons which is currently not possible when already using the background for the icon.

### Changes
To prepare for this already, this PR introduces a dedicated icon element for all buttons to detach the icon from the background.

- [x] `CHANGELOG` entry **Will be added at the end of v4**

### DIFF
--- a/src/scss/components/buttons/_subtitle-toggle-button.scss
+++ b/src/scss/components/buttons/_subtitle-toggle-button.scss
@@ -9,12 +9,16 @@
   }
 
   &.#{$prefix}-subtitles-on {
-    background-image: url('../../assets/images/subtitles-on.svg');
+    .#{$prefix}-icon {
+      background-image: url('../../assets/images/subtitles-on.svg');
+    }
     background-size: 1.5em;
   }
 
   &.#{$prefix}-subtitles-off {
-    background-image: url('../../assets/images/subtitles-off.svg');
+    .#{$prefix}-icon {
+      background-image: url('../../assets/images/subtitles-off.svg');
+    }
     background-size: 1.5em;
   }
 }
--- a/src/scss/components/buttons/_subtitle-toggle-button.scss
+++ b/src/scss/components/buttons/_subtitle-toggle-button.scss
@@ -12,13 +12,11 @@
     .#{$prefix}-icon {
       background-image: url('../../assets/images/subtitles-on.svg');
     }
-    background-size: 1.5em;
   }
 
   &.#{$prefix}-subtitles-off {
     .#{$prefix}-icon {
       background-image: url('../../assets/images/subtitles-off.svg');
     }
-    background-size: 1.5em;
   }
 }
--- a/src/scss/components/settings/_settings-panel-page-open-button.scss
+++ b/src/scss/components/settings/_settings-panel-page-open-button.scss
@@ -6,15 +6,17 @@
 
   .#{$prefix}-icon {
     background-image: url('../../assets/images/setting.svg');
+    max-height: .8em;
   }
-  max-height: .8em;
 
   &:hover {
     @include svg-icon-shadow;
   }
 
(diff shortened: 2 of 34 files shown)


src/scss/components/settings/_settings-panel-page-open-button.scss
```diff
@@ -4,15 +4,19 @@
...
 
   &:hover {
     @include svg-icon-shadow;
   }
 
   &.#{$prefix}-on {
-    transform: rotate(30deg);
+    .#{$prefix}-icon {
+      transform: rotate(30deg);
+    }
```
stonko1994, line 19: Only rotate the image and not the whole button as this would break additional styling.

src/ts/DOM.ts
```diff
@@ -347,7 +347,7 @@ export class DOM {
 
   /**
    * Appends one or more DOM elements as children to all elements.
-   * @param childElements the chrild elements to append
+   * @param childElements the child elements to append
```
123mpozzi, line 350: Nice catch!

src/scss/components/settings/_settings-panel-page-open-button.scss
```diff
@@ -4,15 +4,19 @@
 %ui-settingspanelpageopenbutton {
   @extend %ui-button;
```
123mpozzi, line 5: By the way I am not sure if this button even exists anymore in modern UI. At least it doesn't seem to have an icon anymore  -- unrelated to the changes of this PR 

image

> In v3 UI it is on the side of "Subtitles"

src/ts/components/buttons/Button.ts
```diff
@@ -24,6 +48,12 @@ export interface ButtonConfig extends ComponentConfig {
    * Default: false
    */
   acceptsTouchWithUiHidden?: boolean;
+
+  /**
+   * The style of the button.
+   * Default: `ButtonStyle.Icon`
```
123mpozzi: We could directly link the reference

```suggestion
   * Default: {@link ButtonStyle.Icon}
```

src/ts/components/buttons/Button.ts
```diff
@@ -62,9 +93,45 @@ export class Button<Config extends ButtonConfig> extends Component<Config> {
...
+        break;
+      case ButtonStyle.TextWithLeadingIcon:
+        addIconElement();
+        addLabelElement();
+        break;
+      case ButtonStyle.TextWithTrailingIcon:
+        addLabelElement();
+        addIconElement();
+        break;
+    }
```
123mpozzi, line 134: Looks very clean, good job! 💪

src/scss/components/buttons/_subtitle-toggle-button.scss
```diff
@@ -9,12 +9,14 @@
   }
```
123mpozzi, line 9: Note: this element seems to not be used anywhere at the moment (only its `StoredSubtitleLanguage` interface is). Could be worth considering to delete it in the future to simplify the codebase

123mpozzi (commented):
Looks good, just left a few comments

src/scss/components/settings/_settings-panel-page-open-button.scss
stonko1994: Yea I know but I'd like to keep it in case someone used it in a custom UI

src/scss/components/buttons/_subtitle-toggle-button.scss
stonko1994: Yes I know but the Component still exists. We might want to keep it as it could be that some customer have use-cases like that.

123mpozzi (approved):

PR #733, opened 2025-08-13
Switch control- and title-bar to display flex

stonko1994:
## Description

The control-bar and title-bar are currently layouted using custom margin and padding values. The currently used values are quite complicated as e.g. the control-bar uses `padding: 1em 1em 0.5em` and it's hard to understand why this is the case. The `0.5em` bottom is needed as its `-container-wrapper` also uses `margin: 0 0.5em` to create spacing between the top and bottom row.

Using `dispaly: flex` with `gap` simplifies this as the outer padding/margin no longer needs to be respected manually.

| Before | After |
| --- | --- |
| Screenshot 2025-08-13 at 13 07 18 | Screenshot 2025-08-13 at 13 07 10 |

_Note the difference in the blue container_

- [x] `CHANGELOG` entry **Will be added on the base branch toward the end**


src/scss/_cast-receiver.scss
```diff
@@ -11,7 +11,6 @@
 
     %box {
       background: $background-color;
-      padding: .3em .5em;
```
stonko1994, line 14: There is no longer a need to customize the padding for the cast receiver.

src/scss/components/_control-bar.scss
```diff
@@ -33,10 +40,7 @@
   }
 
   .#{$prefix}-controlbar-bottom {
-    white-space: nowrap; // Required for iOS 8.2 to avoid wrapped controlbar due to wrong size calculation
```
stonko1994, line 36: I guess no player supports iOS 8 anymore 🫠

felix-hoc (approved):

PR #734, opened 2025-08-13
Update TV UI design for v4

stonko1994:
## Description

Updates the UX for TV to match the overall v4 design language. This include changes to how an focused element is looking. The new UX is also applied for non TV variants, e.g. when using keyboard navigation

### Seekbar Web
| Before | After |
| --- | --- |
| Screenshot 2025-08-13 at 13 51 23 | Screenshot 2025-08-13 at 13 51 37 |

### Seekbar TV
| Before | After |
| --- | --- |
| Screenshot 2025-08-13 at 13 52 00 | Screenshot 2025-08-13 at 13 52 07 |

### Buttons Web
| Before | After |
| --- | --- |
| Screenshot 2025-08-13 at 13 51 26 | Screenshot 2025-08-13 at 13 51 40 |

### Buttons TV
| Before | After |
| --- | --- |
| Screenshot 2025-08-13 at 13 56 11 | Screenshot 2025-08-13 at 13 51 50 |


src/scss/_tv.scss
```diff
@@ -2,100 +2,19 @@
 @import 'mixins';
```
123mpozzi, line 3: The Reset button is not visible (but still working) in Subtitle Customisation setting page

| expected | actual |
| --- | --- |
| image | image |

src/ts/UIFactory.ts
```diff
@@ -548,9 +548,9 @@ function tvUILayout() {
     components: [
       subtitleOverlay,
```
123mpozzi, line 549: For some reason, re-loading the index page displays a different layout. This does not happen to other UI layouts, or on modern base branch.
Workaround is to select the tv layout from the dropdown.

Non-blocking, but it is annoying. Would be nice to know the reason why this is happening

image

src/scss/_mixins.scss
```diff
@@ -150,16 +150,23 @@
   width: $marker-dimension;
 }
 
-@mixin focusable($inset: false) {
+@mixin focusable($inset: false, $border-radius: .3em, $focus-element-box-shadow: 0 0 0 .1em rgba($color-focus, .4)) {
```
123mpozzi, line 153: Really cool, thanks for making this customizable! 💪

src/scss/_tv.scss
```diff
@@ -2,100 +2,19 @@
...
```
123mpozzi, line 2: When using keyboard navigation, and toggling some dropdown(up?) like **subtitles** or **audiotracks** by opening and closing it via _spacebar_, the focus moves to a different element.

It has a different behavior than the rest of the UI layouts, where the focus stays on the dropdown toggle element (e.g. PiP, or FullScreen in normal UI layout).
Is this expected?

-- could this be due to `playbackToggleOverlay` reordering in `UIFactory`?

src/scss/components/buttons/_huge-playback-toggle-button.scss
```diff
@@ -81,4 +81,14 @@
...
+
+  &:focus {
+    border-radius: unset;
+    background-color: unset;
+    backdrop-filter: none;
+  }
+
+  &:active {
+    transform: unset;
+  }
```
123mpozzi, line 93: Would be nice for UX to have some indicator that the play button is focused apart from the border indicator. Especially since we had it before:


image

src/scss/_mixins.scss
```diff
@@ -150,16 +150,23 @@
...
   &:focus {
     @if $inset == false {
       box-shadow: $focus-element-box-shadow;
     } @else {
       box-shadow: inset $focus-element-box-shadow;
     }
+
+    background-color: $color-focus;
+    backdrop-filter: blur(5px);
```
123mpozzi, line 162: It seems that `blur()` might be a performance killer on low-end devices, there are multiple reports on the web.

Non blocking, but we should test on some low-end device and monitor how it goes.

123mpozzi, line 162: It seems adding the prefix property as well as a fallback increases the [support by ~4%](https://caniuse.com/css-backdrop-filter).

```suggestion
    -webkit-backdrop-filter: blur(5px);
    backdrop-filter: blur(5px);
```

src/scss/_tv.scss
stonko1994: I'm still working on the whole focusing topic for the TV UI. This is not considered part of this PR.

src/scss/components/buttons/_huge-playback-toggle-button.scss
stonko1994: Thx. I'll collect this feedback and we'll discuss it.

src/scss/_mixins.scss
stonko1994: Thx. I'll check on TVs

stonko1994: Thx. Our build tooling does this for us automatically already.

src/scss/_tv.scss
stonko1994: [b5eaf30](https://github.com/bitmovin/bitmovin-player-ui/pull/734/commits/b5eaf3063a20a86b56547623ed7c367e6d690ca6)

src/ts/UIFactory.ts
stonko1994: Thanks. I had that already on another PR but it makes sense to already include it here: [9c4b92b](https://github.com/bitmovin/bitmovin-player-ui/pull/734/commits/9c4b92b925887f4b3bc0e66cd79e9f333600f4b2)

123mpozzi (approved):

PR #736, opened 2025-08-14
Modern UI - Ad Screen ControlBar adjustments

123mpozzi:
## Description

This PR provides some UX adjustments to the control bar shown in ad layouts.

## Changes
- Slide down **background shadow** on controls hide
  > Gives more space to the ad itself
- Keep `AdSkipButton` in a fixed position
  > Since sliding it down on controls hide has worse UX: what if the
user is pointing with the mouse or finger at the button, and the
button goes down at that same moment?
Keeping it fixed position will ensure the user does not have to
adjust their pointer position again
- Refactor `AdControlBarBottom` and part of `ControlBar` logic into its own `AdControlBar` to better separate the concerns

- [x] `CHANGELOG` entry - `no need since base is modern/ui, a changelog will be added at the end release`


src/ts/components/ads/AdControlBar.ts
```diff
@@ -0,0 +1,96 @@
+import { ControlBar, ControlBarConfig } from '../ControlBar';
+import { Container, ContainerConfig } from '../Container';
+import { PlayerAPI } from 'bitmovin-player';
+import { UIInstanceManager } from '../../UIManager';
+import { i18n } from '../../localization/i18n';
+
+export interface AdControlBarConfig extends ControlBarConfig {
+  topComponents?: Container<ContainerConfig>;
+  bottomComponents?: Container<ContainerConfig>;
+}
```
stonko1994: I'm not sure if this is a good API design if we add this top and bottom components to the Config. The reason is that the `ControlBarConfig` still accepts the `components` attribute as well. This would be quite confusing having all 3. `components`, `topComponents` and `bottomComponents`. Is there a ways to get the same functionality working without the need for this two new properites?

src/scss/components/ads/_ad-control-bar.scss
```diff
@@ -2,8 +2,33 @@
...
 .#{$prefix}-ui-controlbar {
+  @extend %ui-container;
+
+  @include layout-align-bottom;
+  @include slide-down;
+
+  background: linear-gradient(to bottom, $color-transparent, $color-background-bars);
+  box-sizing: border-box;
+  line-height: 1em;
+  padding: 1em 1em .5em;
```
stonko1994: I changd the bars to display flex in [this](https://github.com/bitmovin/bitmovin-player-ui/pull/733/files) PR. Can you do the same for this new bar?

src/scss/components/ads/_ad-status-overlay.scss
stonko1994, line 1: Is the `AdStatusOverlay` still used?

src/ts/components/ads/AdControlBar.ts
```diff
@@ -0,0 +1,96 @@
...
+ * @example
+ * const adBar = new AdControlBar({
+ *   topComponents: new Container(...),
+ *   bottomComponents: new Container(...)
+ * });
+ * 
+ * @category Components
+ */
+export class AdControlBar extends ControlBar {
+  private static readonly CLASS_SLID_DOWN = 'slid-down';
```
stonko1994: Intended to use past tens for this class? `slid` vs `slide`?

src/ts/components/ads/AdControlBar.ts
```diff
@@ -0,0 +1,96 @@
...
+      role: 'region',
+      ariaLabel: i18n.getLocalizer('controlBar'),
+    }, <ControlBarConfig>this.config);
+
+    this.topContainer = config.topComponents || null;
+    this.bottomContainer = config.bottomComponents || null;
+  }
+
+  hide(): void {
+    if (this.bottomContainer) {
```
stonko1994: What if we recursively check all components in `config.components` and look for the seekbar and hide all other containers? 🤔 Too much magic?

src/ts/components/ads/AdControlBar.ts
stonko1994: Idea: https://github.com/bitmovin/bitmovin-player-ui/pull/736#discussion_r2282574875

src/scss/components/ads/_ad-status-overlay.scss
123mpozzi: Yes, the control bar does not contain the skip button itself. That is still inside `AdStatusOverlay`

src/ts/components/ads/AdControlBar.ts
123mpozzi: Yes, the past tense is intentional. The idea is to indicate that the transition has already been applied.
So `slid` means the `slide-down` effect has been applied and is currently in effect.

src/scss/components/ads/_ad-control-bar.scss
123mpozzi: Sure, the design is based on the normal ControlBar so no issues :)
Address in 3fe4c07a054802a644895ea60fe241e2f850607a

src/ts/components/ads/AdControlBar.ts
123mpozzi: Switched to dynamically computer top/bottom components in b432ff0c0f5bfe0562f05cae976c1642c1407d62 .
Now `AdControlBar` is used via `components` like a normal element

src/ts/UIFactory.ts
```diff
@@ -356,11 +356,16 @@ function adsUILayout() {
...
       }),
-      new AdControlBarBottom(),
+      new Container(<ContainerConfig>{
+        components: [
+          new PlaybackToggleButton(),
+          new VolumeToggleButton(),
+          new Spacer(),
+          new FullscreenToggleButton(),
+        ],
+      }),
```
stonko1994: Please keep the previous approach, where it's possible to set CSS classes from here. This allows more flexibility in case someone wants to build his own UI layout. **Applies for all occasions**

```suggestion
      new Container({
        components: [
          new PlaybackToggleButton(),
          new VolumeToggleButton(),
          new Spacer(),
          new FullscreenToggleButton(),
        ],
        cssClasses: ['controlbar-top ad-controlbar-top'],
      }),
```

src/ts/components/ads/AdControlBar.ts
```diff
@@ -0,0 +1,115 @@
...
+import { PlayerAPI } from 'bitmovin-player';
+import { UIInstanceManager } from '../../UIManager';
+import { i18n } from '../../localization/i18n';
+import { SeekBar } from '../seekbar/SeekBar';
+
+export interface AdControlBarConfig extends ControlBarConfig {
+  // Uses the inherited 'components' property from ControlBarConfig
+  // Components containing SeekBar will be treated as "top" components
+  // Components without SeekBar will be treated as "bottom" components
+}
```
stonko1994: I think we don't need this at all if there is no additional config option.

stonko1994 (commented):
I think we need to rethink the whole slide down solution. Customization of our UI is a big aspect and it should be easo to support the same feature also e.g. when the seekbar is in the bottom container of the `AdControlBar` or if there is only one `Container`.

Let's talk offline.

src/ts/components/ads/AdControlBar.ts
123mpozzi: Address in d2e834f0831c578144c93616a53a8527a5bdb30e

src/ts/UIFactory.ts
123mpozzi: Address in d2e834f0831c578144c93616a53a8527a5bdb30e

123mpozzi:
> I think we need to rethink the whole slide down solution. Customization of our UI is a big aspect and it should be easo to support the same feature also e.g. when the seekbar is in the bottom container of the `AdControlBar` or if there is only one `Container`.
> 
> Let's talk offline.

I address this in d2e834f0831c578144c93616a53a8527a5bdb30e

src/scss/components/ads/_ad-control-bar.scss
```diff
@@ -0,0 +1,67 @@
+@import '../../variables';
+@import '../../mixins';
+
+// Keep this mixin here for now since it is specific to the ad control bar
+@mixin hidden-animated-collapse-and-fade-out($duration: $animation-duration, $original-height: 2em) {
+  // sass-lint:disable no-transition-all
+  transition: all $duration ease-out;
```
stonko1994: Is the `transition: all` necessary? can we specify opacity and height explicitly?

src/ts/components/ads/AdControlBar.ts
```diff
@@ -0,0 +1,120 @@
...
+ */
+export class AdControlBar extends ControlBar {
+  private containersToHide: Container<ContainerConfig>[] = [];
+  private containersToKeepVisible: Container<ContainerConfig>[] = [];
+
+  constructor(config: AdControlBarConfig) {
+    super({
+      ...config,
+      cssClasses: ['ad-controlbar'],
+    });
```
stonko1994: Can we keep our current style of calling super in components: (and see comment below)


```suggestion
    super(config);
```

src/ts/components/ads/AdControlBar.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    this.config = this.mergeConfig(config, {
+      cssClass: 'ui-controlbar',
```
stonko1994: Let's use the cssClasses approach to add additional classes to this component:


```suggestion
      cssClasses: ['ad-controlbar'],
```

src/ts/components/ads/AdControlBar.ts
```diff
@@ -0,0 +1,120 @@
...
+    uimanager.onControlsShow.subscribe(() => {
+      this.show();
+    });
+
+    uimanager.onControlsHide.subscribe(() => {
+      this.hide();
+    });
+  }
+
+  private findContainerWithSeekBar(container: Container<ContainerConfig>): boolean {
```
stonko1994: The naming and the return value doesn't match IMO. The naming of `findContainer` would indicate that the return value is a `Container`. Which we could do as well which could also simplify its implementaiton.

A return type of boolean rather indicates that the name should be `containsSeekBar` or similar.

src/ts/UIFactory.ts
```diff
@@ -345,9 +345,9 @@ function uiLayout(config: UIConfig) {
 }
 
 function adsUILayout() {
-  let controlBar = new ControlBar({
+  let controlBar = new AdControlBar({
     components: [
-      new Container({
+      new Container(<ContainerConfig>{
```
stonko1994: I think we can remove the type cast here


```suggestion
      new Container({
```

src/ts/UIFactory.ts
```diff
@@ -356,11 +356,19 @@ function adsUILayout() {
             cssClasses: ['text-right'],
           }),
         ],
-        cssClasses: ['controlbar-top ad-controlbar-top'],
+        cssClasses: ['ad-controlbar-top'],
+
+      }),
+      new Container(<ContainerConfig>{
```
stonko1994: I think we can remove the type cast here


```suggestion
      new Container({
```

src/ts/UIFactory.ts
```diff
@@ -513,9 +521,9 @@ function smallScreenUILayout() {
 }
 
 function smallScreenAdsUILayout() {
-  let controlBar = new ControlBar({
+  let controlBar = new AdControlBar({
     components: [
-      new Container({
+      new Container(<ContainerConfig>{
```
stonko1994:
```suggestion
      new Container({
```

src/ts/UIFactory.ts
```diff
@@ -524,11 +532,18 @@ function smallScreenAdsUILayout() {
             cssClasses: ['text-right'],
           }),
         ],
-        cssClasses: ['controlbar-top ad-controlbar-top'],
+        cssClasses: ['ad-controlbar-top'],
+      }),
+      new Container(<ContainerConfig>{
```
stonko1994:
```suggestion
      new Container({
```

src/ts/UIFactory.ts
```diff
@@ -697,9 +712,9 @@ function tvAdsUILayout() {
       new AdClickOverlay(),
       playbackToggleOverlay,
       adStatusOverlay,
-      new ControlBar({
+      new AdControlBar({
         components: [
-          new Container({
+          new Container(<ContainerConfig>{
```
stonko1994:
```suggestion
          new Container({
```

stonko1994 (commented):
Getting there 💪 Do you know where this padding (purple) is coming from? Makes the seekbar look out of place slightly. Image

src/scss/components/ads/_ad-control-bar.scss
123mpozzi: Yes, we can. Address in ae178d099f3fdb1cf217f504d6f1b68f326ced4c

src/ts/components/ads/AdControlBar.ts
123mpozzi: Address in 75d263672ddb3d9313b927396e09d3156d252636

src/ts/components/ads/AdControlBar.ts
123mpozzi: Address in 75d263672ddb3d9313b927396e09d3156d252636

src/ts/components/ads/AdControlBar.ts
123mpozzi: Make sense, address in 45951a81c21ae67071fd4da170376364428d30bd

123mpozzi:
> Getting there 💪 Do you know where this padding (purple) is coming from? Makes the seekbar look out of place slightly. Image

This is coming from `row-gap: 0.5em` and was introduced in 3fe4c07a054802a644895ea60fe241e2f850607a to address [this comment](https://github.com/bitmovin/bitmovin-player-ui/pull/736#discussion_r2282569437).

The goal of that commit was to style the ad controlbar the same way controlbar is styled in `feature/modern-ui-base` since it was changed in #733.

The style is the exact same one used for the normal controlbar.

src/ts/UIFactory.ts
123mpozzi: Address in 895ed81d36d15421f116395c4ba261f0cec44621

src/ts/components/ads/AdControlBar.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+export class AdControlBar extends ControlBar {
+  private containersToHide: Container<ContainerConfig>[] = [];
+  private containersToKeepVisible: Container<ContainerConfig>[] = [];
+
+  constructor(config: AdControlBarConfig) {
+    super(config);
+
+    this.config = this.mergeConfig(config, {
+      cssClasses: ['ui-controlbar', 'ad-controlbar'],
```
stonko1994: The `ui-controlbar` class is already added by the super class so you don't have to add it again. This way it results in adding the same class twice:

Image


```suggestion
      cssClasses: ['ad-controlbar'],
```

stonko1994:
> This is coming from `row-gap: 0.5em` and was introduced in [3fe4c07](https://github.com/bitmovin/bitmovin-player-ui/commit/3fe4c07a054802a644895ea60fe241e2f850607a) to address [this comment](https://github.com/bitmovin/bitmovin-player-ui/pull/736#discussion_r2282569437).
> 
> The goal of that commit was to style the ad controlbar the same way controlbar is styled in `feature/modern-ui-base` since it was changed in #733.
> 
> The style is the exact same one used for the normal controlbar.

Ah I see, 💩. Do you have an idea how we could fix this problem?

src/scss/components/ads/_ad-control-bar.scss
```diff
@@ -0,0 +1,66 @@
...
+  height: $original-height;
+
+  &.#{$prefix}-hidden {
+    opacity: 0;
+    height: 0;
+    pointer-events: none;
+  }
+}
+
+.#{$prefix}-ui-controlbar {
```
stonko1994: This should be `ad-controlbar` no? The styles in here are specific for the ad control bar and shouldn't affect the regular `ui-controlbar` as far as I know.


```suggestion
.#{$prefix}-ad-controlbar {
```

123mpozzi: Yes, let's do that

123mpozzi:
> > This is coming from `row-gap: 0.5em` and was introduced in [3fe4c07](https://github.com/bitmovin/bitmovin-player-ui/commit/3fe4c07a054802a644895ea60fe241e2f850607a) to address [this comment](https://github.com/bitmovin/bitmovin-player-ui/pull/736#discussion_r2282569437).
> > The goal of that commit was to style the ad controlbar the same way controlbar is styled in `feature/modern-ui-base` since it was changed in #733.
> > The style is the exact same one used for the normal controlbar.
> 
> Ah I see, 💩. Do you have an idea how we could fix this problem?

I have a few, let's continue with this on another PR tho

stonko1994 (approved):

PR #732, opened 2025-08-13
Update TV UI layout for v4

stonko1994:
## Description

This PR updates the layout of the TV UI to improve it for v4. It does not update the design yet, just the arrangement of elements.

| Old | New |
| --- | --- |
| Screenshot 2025-08-13 at 09 57 38 | Screenshot 2025-08-13 at 09 57 26 |

What doesn't work yet:
- Spatial navigation to the control bar

## Checklist (for PR submitter and reviewers)

- [x] `CHANGELOG` entry **Will be added on the base branch at the end**

PR #738, opened 2025-08-20
Extend Spatial Navigation capabilities

stonko1994:
## Description

Our SpatialNavigation implementation is tailored to our current UI layout and doesn't allow much flexibility in organizing elements. Our new TV UI layout #732 needs more flexibility in that regard.

### Changes
There were a few major changes to the Spatial Navigation implementation:
- All focus tracking is now Component based instead of HTML element based to keep more contexts
- The cutoffAngle changed from 45 to 89° to allow more elements into the target direction.
- Introducing `FocusableContainers`:
  - `FocusableContainers` group together certain elements into its container and allow navigating to the container instead of navigating to the individual components. This is especially helpful for our ControlBar bottom element. In the new layout, this container has a Play button on the left and a few other buttons to the right with spacing inbetween. To allow navigating to the bar (row) itself, a `FocusableContainer` can be focused when navigating downwards. 
  - It also allows to specify a priority element which should get focus when navigating to the container
  - Elements within a `FocusableContainer` are always prioritized over all other components on screen
- Prioritizing elements with a 0 angle
  -  When there is an element within the perfect direction of the navigation (aka angle of 0), it will always be prioritized over elements even though others might be closer. This is helpful when _not_ using `FocusableContainers` and allowing navigating to every element on screen from everywhere.
- Add support for our new dynamic SettingsPanel approach by changing the `ListNavigationGroup` to a `SettingsPanelNavigationGroup` (`ListBoxes` are also `SettingsPanels` now)

- [x] `CHANGELOG` entry **Will be collected at the end**


src/ts/components/Component.ts
```diff
@@ -233,6 +233,7 @@ export class Component<Config extends ComponentConfig> {
       cssClasses: [],
       hidden: false,
       disabled: false,
+      tabIndex: -1,
```
stonko1994, line 236: Per default always adding the `tabIndex` of `-1` to disable focusing every element. Buttons e.g. have a default tab index of 0 allowing focusing.

TeslaAdis (approved):
Great approach with focusable containers, all changes make sense - did not manually test it but assume everything still behaves the same

PR #739, opened 2025-08-21
Modern UI - `AdControlBar`: fix padding from `row-gap` not being collapsed with hide animation

123mpozzi:
## Description

- use `margin` instead of adding a padding via `row-gap`
- animate `margin` in the hide transitions as well

- [x] `CHANGELOG` entry - ` no need since base is modern/ui`


src/scss/components/ads/_ad-control-bar.scss
```diff
@@ -3,13 +3,14 @@
 
 // Keep this mixin here for now since it is specific to the ad control bar
 @mixin hidden-animated-collapse-and-fade-out($duration: $animation-duration, $original-height: 2em) {
-  transition: opacity $duration ease-out, height $duration ease-out;
+  transition: opacity $duration ease-out, height $duration ease-out, margin-bottom $duration ease-out;
```
stonko1994: You're using `margin-top` but animating `margin-bottom`.

src/scss/components/ads/_ad-control-bar.scss
```diff
@@ -3,13 +3,14 @@
...
   opacity: 1;
   height: $original-height;
 
   &.#{$prefix}-hidden {
     opacity: 0;
     height: 0;
+    margin: 0;
```
stonko1994: You're using `margin-top` but here you're resetting all of the `margins`

```suggestion
    margin-top: 0;
```

123mpozzi: Address in 82f484a546487ee6781c1e49d3e34ab715c8b8b9

src/scss/components/ads/_ad-control-bar.scss
123mpozzi: Nice eye! Address in 82f484a546487ee6781c1e49d3e34ab715c8b8b9

stonko1994 (approved):

PR #741, opened 2025-08-25
Add icon component to labels

stonko1994:
## Description

In an attempt to streamline all icon and text sizes in our UI, I figured out that using pseudo-elements (`::after`, `::before`) for adding icons to buttons/labels isn't a good approach and it's super hard to have consistent styling.

E.g. we have css code like this to add icons to labels:
```css
&::before {
  background-image: url('../../assets/images/check.svg');
  background-repeat: no-repeat;
  background-size: 1.5em auto;
  content: ' ';
  display: inline-block;
  height: 1.5em;
  vertical-align: -.4em;
  width: 1.5em;
}
```

The main problems with this approach are:
- repetitive (every icon uses the very same code resulting in code duplication)
- `vertical-align: -.4em;`, what is this magic?
- It breaks the layout
  - We will incease the icon and font size in v4 (separe PR). Using pseudo elements breaks a consistent height of e.g. different SettingsPanelItems depending if they contain an icon or now. Using a dedicated component fixes/simpliefies this.

## Changes
- Introducing a dedicated `Icon` Component to extract similarities between button-icons and label-icons into a separate component
  - Since our top level component css naming scheme is to prefix with `ui` I had to change the css selectors for applying the icons again.
- Introducing a `LabelStyle` (similar to #731) to add a dedicated icon element to labels.


src/scss/components/buttons/_button.scss
```diff
@@ -14,27 +14,8 @@
   transition: transform .15s ease;
   // sass-lint:disable no-vendor-prefixes
   -webkit-tap-highlight-color: transparent;
-  display: flex;
```
stonko1994, line 17: This stlye was moved to `_icon.scss`

felix-hoc (approved):
nice, looking good!

PR #742, opened 2025-08-26
Streamline font and icon size

stonko1994:
## Description

We got the feedback multiple times already that our icons do feel blurry/fuzzy. To address this topic we streamling all our icon and text sizes for v4. The new icon size is 24x24 with a default font size of 18px.

## Changes
To simplify the whole icon/font size topic I removed all custom font sizes and introduces variables for this which can be easily simplified. All other sizes are based on the base font size.


src/scss/components/buttons/_huge-playback-toggle-button.scss
```diff
@@ -46,9 +46,7 @@
 
   .#{$prefix}-ui-icon {
     background-image: url('../../assets/images/play.svg');
-    background-position: center;
-    background-repeat: no-repeat;
-    background-size: 4em;
+    background-size: $icon-size-large;
```
123mpozzi, line 49: The button is not centered anymore. Can reproduce better on `buildUI` and `buildTvUI`: the button is pushed a little to the right.

Image

Image

> It seems due to the new default padding

src/scss/components/settings/_settings-panel-item.scss
```diff
@@ -3,8 +3,7 @@
 %ui-settings-panel-item {
   @include focusable(true, 0, 0);
 
-  line-height: 1.5em;
-  padding: .5em .7em;
+  padding: 1rem;
```
123mpozzi, line 6: This padding works well with _portrait_ mode, but badly for _landscape_ mode: only a few elements can be shown at once.

I'd propose to reduce it considerably for landscape mode, to something like `.3rem` - I guess we can use media-queries if we wanna apply this smaller padding to only landscape

| Preview | Padding |
|---------|---------|
| Image | 1rem |
| Image | 0.3rem |

src/scss/components/seekbar/_seek-bar.scss
```diff
@@ -93,11 +93,11 @@ $seekbar-height: .3125em;
     .#{$prefix}-seekbar-markers {
       @extend %bar;
 
-      $marker-width: .25em;
-      $marker-indicator-height: .625em;
-      $marker-indicator-border-radius: .125em;
+      $marker-width: .25rem;
+      $marker-indicator-height: .625rem;
```
123mpozzi: With the new font and icon sizes, the marker looks a bit short to me, especially on small screens. But not sure how to improve the situation. What do you think?

src/scss/components/overlays/_subtitle-overlay.scss
```diff
@@ -8,7 +8,6 @@
     @include hidden;
 
     bottom: 0;
-    font-size: 1.2em;
```
123mpozzi, line 11: I guess it's not related to the changes, but is it expected that subtitles are on the complete bottom of the playerview?
Not necessary to change it here, just wanted to bring this up.


Image

E.g., for CAE it looks like we apply some margins via styling:

```scss
// _subtitle-overlay-cea608.scss

    bottom: 2em;
    left: 3em;
    right: 3em;
    top: 2em;
```

src/scss/components/overlays/_recommendation-overlay.scss
```diff
@@ -64,7 +63,6 @@
         right: $margin;
 
         .#{$prefix}-innertitle {
-          font-size: 1.2em;
```
123mpozzi, line 67: ###  Issue 1

With the new padding and font-sizes it's not possible to see all the recommendations on some screens. E.g. in `fullyFeatured` we have 3 recommendation items, but only 1 is shown.

Note _- empty thumbnail is unrelated issue: it looks like `placehold.it` is no more_

| Before | After |
|---------|---------|
| Image |  Image |


--- 

### Issue 2

Re-play button gets hidden as well!

src/scss/components/labels/_label.scss
```diff
@@ -5,6 +5,10 @@
   @include hidden;
   @include ui-icon-container;
 
+  margin: 0;
+  // sass-lint:disable no-vendor-prefixes
+  -webkit-tap-highlight-color: transparent;
```
123mpozzi, line 10: nit: could add a comment here explaining this field

src/scss/components/_control-bar.scss
```diff
@@ -22,26 +21,27 @@
...
+    > .#{$prefix}-container-wrapper {
+      column-gap: $font-size;
       margin: 0 .5em;
     }
   }
 
   .#{$prefix}-controlbar-bottom {
+    // We can't use the row-gap property here because the AdControlBar collapses certain elements which
+    // would result in additional margins.
     margin-top: .5em;
```
123mpozzi, line 39: Nice addition! 💪

src/scss/bitmovinplayer-ui.scss
```diff
@@ -60,6 +60,8 @@
   color: $color-primary;
   font-family: $font-family;
   font-size: $font-size;
+  font-weight: $font-weight-medium;
+  line-height: $font-size;
```
123mpozzi: Nit: It’s usually better to use a unitless `line-height` instead of tying it to a fixed length. A unitless value means children inherit the ratio, not the computed pixels, so their `line-height` adjusts automatically to their own font-size. That keeps headings, captions, and body text looking balanced without extra overrides.

Using a length (e.g. `line-height: $font-size;`) forces all children to inherit the same fixed value, which often leads to cramped or misaligned text.

We could create a unitless variable in this case: `$line-height-default: 1.2 !default;`


```suggestion
  $line-height-default: 1.2 !default; // into variables.scss
  
  line-height: $line-height-default;
```

123mpozzi (commented):
Left some comments, but overall the changes looks good! 💪 

Quickly tested on Chrome, Firefox, and Safari

src/scss/components/buttons/_huge-playback-toggle-button.scss
stonko1994: this is already the case on the base branch and was introduced because of the new TV UI layout. I'll check if we can do something. (not in this PR)

src/scss/components/settings/_settings-panel-item.scss
stonko1994: I'll not this as feedback but would like to keep it as it is for now. Maybe we need to tweak font sizes in land-scape as well. Making the padding very small doesn't look good IMO.

src/scss/components/overlays/_subtitle-overlay.scss
stonko1994: This was always the case already. out of scope for this PR

src/scss/components/overlays/_recommendation-overlay.scss
stonko1994: the recommendation overlay wasn't udpated for the v4 UI yet. Please ignore for now.

src/scss/components/labels/_label.scss
stonko1994: IMO the property name is pretty self explanatory already 🤔

src/scss/bitmovinplayer-ui.scss
stonko1994: > It’s usually better to use a unitless line-height instead of tying it to a fixed length.

Do you have any references to this?

123mpozzi: There are a bunch:

https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#number
https://css-tricks.com/almanac/properties/l/line-height/
https://allthingssmitty.com/2017/01/30/nope-nope-nope-line-height-is-unitless/

stonko1994: [7e6bf4c](https://github.com/bitmovin/bitmovin-player-ui/pull/742/commits/7e6bf4c14d9730bafc1be3e02a1088276f93a82e)

src/scss/components/seekbar/_seek-bar.scss
stonko1994: [ec4909b](https://github.com/bitmovin/bitmovin-player-ui/pull/742/commits/ec4909b41ea5ae8812e4ef96f130b4236b0f2378)

src/scss/components/buttons/_huge-playback-toggle-button.scss
123mpozzi: Ok, thanks 🙌

src/scss/components/settings/_settings-panel-item.scss
123mpozzi: Fine to not modify the design right now.
I agree that the small padding does not look optimal, however I think it's better than being able to only look at one option at a time - I'd give usability an edge compared to design

Font size tweak sounds promising tho 💪

src/scss/components/seekbar/_seek-bar.scss
```diff
@@ -93,9 +93,13 @@ $seekbar-bar-height: calc($font-size / 3);
...
-      $marker-width: .25rem;
-      $marker-indicator-height: .625rem;
-      $marker-indicator-border-radius: .125rem;
+      // Marker width is 2/3 of the seekbar height
+      $marker-width: calc($seekbar-bar-height * (2 / 3));
+      // Marker indicator is 5/3 of the seekbar height to extend the indicator above and below the seekbar by 1/3
+      // of the seekbar height
+      $marker-indicator-height: calc($seekbar-bar-height * (5 / 3));
+      // Marker indicator border radius is half the height to make it fully rounded
+      $marker-indicator-border-radius: calc($marker-indicator-height / 2);
```
123mpozzi: Using `indicator height / 2` seems to be too aggressive and does not provide visual benefits to the roundness of the marker; instead, it messes up the range marker.

```suggestion
      $marker-indicator-border-radius: calc($marker-indicator-height / 4);
```

### Marker Range

| current | proposed |
|-|-|
| image | image |
 
### Single Marker
 
 | current | proposed |
|---|---|
| image | image |

_Marker height in the pics is increased to highlight the border radius difference_

123mpozzi: Nit: I'd suggest making the chapter markers a bit slimmer so they carry less visual weight. That way, the playback position marker remains the dominant element on the seekbar, which helps reinforce its primary role.

```suggestion
      // Marker width is 2/5 of the seekbar height
      $marker-width: calc($seekbar-bar-height * (2 / 5));
      // Marker indicator is 5/3 of the seekbar height to extend the indicator above and below the seekbar by 1/3
      // of the seekbar height
      $marker-indicator-height: calc($seekbar-bar-height * (5 / 3));
      // Marker indicator border radius is 1/6 the height to make it fully rounded
      $marker-indicator-border-radius: calc($marker-indicator-height / 6);
```

**Current**  
image  

**Proposed**  
image

stonko1994: they are according to the design. IMO the Markers in the proposed design are too slim in comparison to the Seekbar height. Feels off IMO. Since design is subjective I'd prefer to stick to the design template instead of re-defining it in PRs.

stonko1994: The problem wasn't the factor but that we dynamically calculate it at all. Since a range marker has a different width, a dynamic corner radius will always look different, no matter the factor.

I reverted back to a consistent border radius: [9a9bf94](https://github.com/bitmovin/bitmovin-player-ui/pull/742/commits/9a9bf94254cc0c25364bc8708e51217817ab8c15)

123mpozzi: Ah I see, good job fixing it! 💪

123mpozzi: Sure, makes sense 🙂

123mpozzi (approved):

PR #743, opened 2025-08-26
Fix title bar orientation with additional buttons

stonko1994:
## Description

The `TitleBar` doesn't support having buttons next to labels. However, in v4 we want to move some buttons up for the small screen UI.

## Changes
To support this we need to introduce `TitleBar` 'rows' to align elements horizontally. This PR introduces the css class `titlebar-row` to allow exactly that.

### DIFF
--- a/src/scss/components/_title-bar.scss
+++ b/src/scss/components/_title-bar.scss
@@ -19,6 +19,13 @@
     row-gap: .5em;
   }
 
+  .#{$prefix}-titlebar-row {
+    > .#{$prefix}-container-wrapper {
+      display: flex;
+      flex-direction: row;
+    }
+  }
+
   > .#{$prefix}-container-wrapper {
     pointer-events: none;
 
--- a/src/ts/UIFactory.ts
+++ b/src/ts/UIFactory.ts
@@ -356,7 +356,6 @@ function smallScreenUILayout() {
           new VolumeToggleButton(),
           new Spacer(),
           new SettingsToggleButton({ settingsPanel: settingsPanel }),
-          // new SubtitleToggleButton(subtitleSelectItem, subtitleSelectBox),
           new FullscreenToggleButton(),
         ],
         cssClasses: ['controlbar-bottom'],
@@ -374,10 +373,16 @@ function smallScreenUILayout() {
       controlBar,
       new TitleBar({
         components: [
-          new MetadataLabel({ content: MetadataLabelContent.Title }),
-          new CastToggleButton(),
-          new AirPlayToggleButton(),
-          new VRToggleButton(),
+          new Container({
+            components: [
+              new MetadataLabel({ content: MetadataLabelContent.Title }),
+              new Spacer(),
+              new CastToggleButton(),
+              new AirPlayToggleButton(),
+              new VRToggleButton(),
+            ],
+            cssClasses: ['titlebar-row'],
+          }),
         ],
       }),
       new DismissClickOverlay({ target: settingsPanel }),
--- a/src/ts/components/TitleBar.ts
+++ b/src/ts/components/TitleBar.ts
@@ -28,19 +28,24 @@ export class TitleBar extends Container<TitleBarConfig> {
   constructor(config: TitleBarConfig = {}) {
     super(config);
 
-    // Only use default MetadataLabels if no custom components are provided.
-    const hasCustomComponents = Array.isArray(config.components) && config.components.length > 0;
-
     this.config = this.mergeConfig(config, {
       cssClass: 'ui-titlebar',
       hidden: true,
       keepHiddenWithoutMetadata: false,
-      components: hasCustomComponents
-        ? config.components
-        : [
+      components: [
+        new Container({
+          components: [
             new MetadataLabel({ content: MetadataLabelContent.Title }),
+          ],
+          cssClasses: ['titlebar-row']
+        }),
+        new Container({
+          components: [
             new MetadataLabel({ content: MetadataLabelContent.Description }),
           ],
+          cssClasses: ['titlebar-row']
+        }),
+      ],
     }, <TitleBarConfig>this.config);
   }
 


src/ts/components/TitleBar.ts
```diff
@@ -28,19 +28,24 @@ export class TitleBar extends Container<TitleBarConfig> {
   constructor(config: TitleBarConfig = {}) {
     super(config);
 
-    // Only use default MetadataLabels if no custom components are provided.
-    const hasCustomComponents = Array.isArray(config.components) && config.components.length > 0;
```
stonko1994, line 32: There is no need to manually detect this as the `mergeConfig` method only overrides in case no components are present.

src/ts/UIFactory.ts
```diff
@@ -374,10 +373,16 @@ function smallScreenUILayout() {
       controlBar,
       new TitleBar({
         components: [
```
stonko1994, line 375: This button placement isn't final yet and just an example.

src/ts/UIFactory.ts
```diff
@@ -356,7 +356,6 @@ function smallScreenUILayout() {
           new VolumeToggleButton(),
           new Spacer(),
           new SettingsToggleButton({ settingsPanel: settingsPanel }),
-          // new SubtitleToggleButton(subtitleSelectItem, subtitleSelectBox),
```
123mpozzi, line 359: Do we know why this was commented out in the first place?

src/ts/components/TitleBar.ts
```diff
@@ -28,19 +28,24 @@ export class TitleBar extends Container<TitleBarConfig> {
...
+          cssClasses: ['titlebar-row']
+        }),
+        new Container({
+          components: [
             new MetadataLabel({ content: MetadataLabelContent.Description }),
           ],
+          cssClasses: ['titlebar-row']
+        }),
+      ],
     }, <TitleBarConfig>this.config);
```
123mpozzi, line 49: Nit: could be clearer:


```suggestion
    } as TitleBarConfig, this.config);
```

src/ts/components/TitleBar.ts
```diff
@@ -28,19 +28,24 @@ export class TitleBar extends Container<TitleBarConfig> {
...
       keepHiddenWithoutMetadata: false,
-      components: hasCustomComponents
-        ? config.components
-        : [
+      components: [
+        new Container({
+          components: [
             new MetadataLabel({ content: MetadataLabelContent.Title }),
+          ],
+          cssClasses: ['titlebar-row']
```
123mpozzi, line 40: Can we keep consistent behaviour between `cssClass` and `cssClasses`? Since there is only one element, I'd prefer to use `cssClass`.

Applies to all cases

src/scss/components/_title-bar.scss
```diff
@@ -19,6 +19,13 @@
     row-gap: .5em;
   }
 
+  .#{$prefix}-titlebar-row {
+    > .#{$prefix}-container-wrapper {
+      display: flex;
+      flex-direction: row;
+    }
+  }
+
```
123mpozzi, line 28: Did you also consider the vertical alignment of the inner elements?

For example, with 2 `titlebar-row`s, this is the default alignment: the button is at the top.

Image

123mpozzi (changes requested):
Design looks good 💪 , just left a few comments.

Unfortunately, the hiding mechanism for small-screen UI seems broken from 5e193d6, not sure why that didn't go through review - can we fix it here?
The issue is that UI is not displayed at all when you hover the player, or click inside it; not even the mouse pointer is shown.

src/ts/UIFactory.ts
stonko1994: Probably for initial testing

src/ts/components/TitleBar.ts
stonko1994: This is nothing new and has always been like that.

src/ts/components/TitleBar.ts
stonko1994: No we can't. `cssClass` will act as the only css class while `cssClasses` act as additional `cssClasses`.

src/scss/components/_title-bar.scss
stonko1994: I don't fully get the use-case here 🤔 Do you mean that the sapcer is taking too much space or what is the problem?

stonko1994:
> Design looks good 💪 , just left a few comments.
> 
> Unfortunately, the hiding mechanism for small-screen UI seems broken from [5e193d6](https://github.com/bitmovin/bitmovin-player-ui/commit/5e193d6718ee62b33d6d6b5774371e0526e5fae1), not sure why that didn't go through review - can we fix it here? The issue is that UI is not displayed at all when you hover the player, or click inside it; not even the mouse pointer is shown.

Upsi. Mistakes were made. I thought this mini change isn't worth waiting for a review 🫠 Thx for bringing up. Will fix it on the base branch. 

[ff54937](https://github.com/bitmovin/bitmovin-player-ui/commit/ff54937d8d24089c4b9b862b6bfc8c207398e847)

src/ts/components/TitleBar.ts
123mpozzi: Ah I see, it's cause it would remove `ui-container`. Thanks for the info

src/ts/components/TitleBar.ts
123mpozzi: Sure, was just a suggestion for improving the existing code 🙂

src/scss/components/_title-bar.scss
123mpozzi: There is no problem, just wondering if you considered the default vertical alignment of the buttons in this case - for example they could be aligned differently as well, with the button staying in the middle of the row (vertically).

But I believe it's better the way it currently is, with the button vertically aligned to the top of the row

123mpozzi (approved):

PR #744, opened 2025-08-26
Update mobile layout

stonko1994:
## Changes
- The small-screen UI is now also used on Web depending on the document width. Previously this was only used on mobile platforms. However, there is no reason to not also use it on web.
- Add missing PiP button
- Moving the safe-area-inset handling directly into the UI framwork. Previously this was directly applied within the iOS Player SDK. Moving it directly to the UI we also support 'full sized' players in mobile safari.


src/scss/components/overlays/_touch-control-overlay.scss
```diff
@@ -23,6 +23,7 @@
   opacity: 1;
   position: absolute;
   text-shadow: $color-shadow-seek-label;
+  z-index: 1;
```
stonko1994, line 26: The `::before` element is above the label. To fix this we apply the z-index of 1 here.

src/scss/components/settings/_settings-panel.scss
```diff
@@ -25,7 +25,7 @@
   overflow-y: auto;
   padding: 0;
   position: absolute;
-  right: 1em;
+  right: max(1em, env(safe-area-inset-right));
```
123mpozzi: What do you think about providing fallback values for `env` variables? So that in case the environment variable is not available, the `max` expression is still evaluated.

```suggestion
  right: max(1em, env(safe-area-inset-right, 0)); // env results 0 in case it is not available, so 1em is used
```

I've read about fallbacks for `env` here: https://developer.mozilla.org/en-US/docs/Web/CSS/env

src/ts/components/overlays/TouchControlOverlay.ts
```diff
@@ -37,6 +37,12 @@ export interface TouchControlOverlayConfig extends ContainerConfig {
    * Default: 15px
    */
   seekDoubleTapMargin?: number;
+
+  /**
+   * Time in milliseconds within which two consecutive taps are considered a double tap.
+   * Default: 200ms
+   */
+  seekDoubleTapTimeout?: number;
```
123mpozzi, line 45: Nice change 😎

src/scss/_variables.scss
```diff
@@ -17,8 +17,9 @@ $color-focus: transparentize($color-highlight, .8);
 $color-background-menu-title-overlay: rgba(0, 0, 0, .25) !default;
 $color-background-menu: #2e2e2e !default;
 $color-background-menu-hover-overlay: rgba(255, 255, 255, .07) !default;
-$color-background-seek-circle: rgba(124, 124, 124, .35) !default;
+$color-background-seek-circle: rgba(255, 255, 255, .35) !default;
```
123mpozzi, line 20: Nice, this makes it way more visible! 💪

src/scss/_variables.scss
```diff
@@ -17,8 +17,9 @@ $color-focus: transparentize($color-highlight, .8);
...
 $color-shadow-seek-label: 0 0 30px 0 rgba(0, 0, 0, .75) !default;
+$color-touch-control-overlay-background: rgba(0, 0, 0, .5) !default;
```
123mpozzi, line 22: Nit: I'd honestly make the overlay completely transparent. I see for example YouTube does that.
Cause I don't see the benefit of showing it to the user

```suggestion
$color-touch-control-overlay-background: rgba(0, 0, 0, 0) !default;
```

src/scss/components/overlays/_subtitle-overlay.scss
```diff
@@ -68,5 +68,10 @@
       bottom: 5em;
       transition: bottom $animation-duration-short ease-in;
     }
+
+    &:not(.#{$prefix}-controlbar-visible) {
+      top: env(safe-area-inset-top);
+      bottom: env(safe-area-inset-bottom);
```
123mpozzi: How do you feel about adding some additional padding? This would ensure the subtitles do not touch the absolute bottom of the player view, resulting in a more pleasant user experience in my opinion.

_Sorry, for some reason I could not do a code suggestion_
```scss
      top: max(1rem, env(safe-area-inset-top, 0));
      bottom: max(1rem, env(safe-area-inset-bottom, 0));
```


| Before | After |
|---|---|
| Image | Image |

123mpozzi (commented):
Very nice changes!
Just a few comments

src/scss/_variables.scss
stonko1994: The difference between touch input and non touch input devices is that the main pause/play toggle action is in the middle of the screen. The pause button is visible on mobile while it's not visible on web. When we have a very bright frame behind it, it's hard to see, hence I decided to add this background.

src/scss/components/overlays/_subtitle-overlay.scss
stonko1994: I'm on your side that the subtitles do not look good when positionied at the very edge of the screen. However, this is already the case in v3 and nothing new again. I'll note that down and maybe we'll get to it for v4.

src/scss/components/settings/_settings-panel.scss
stonko1994: [5504e88](https://github.com/bitmovin/bitmovin-player-ui/pull/744/commits/5504e88b421416fd2e79689aa8dd92ada66c545c)

src/scss/components/overlays/_subtitle-overlay.scss
123mpozzi: Ok, that would be awesome 🙂

123mpozzi (approved):

PR #746, opened 2025-08-29
Support spatial navigation for RecommendationOverlay

stonko1994:
## Description

Navigating the `RecommendationOverlay` isn't possible using our `SpatialNavigation` implementation.

### Changes
Introducing a `RecommendatdionOverlayNavigationGroup` to dynamically support navigating it.

### DIFF
--- a/src/ts/components/Container.ts
+++ b/src/ts/components/Container.ts
@@ -113,7 +113,7 @@ export class Container<Config extends ContainerConfig> extends Component<Config>
   /**
    * Updates the DOM of the container with the current components.
    */
-  protected updateComponents(): void {
+  updateComponents(): void {
     /* We cannot just clear the container to remove all elements and then re-add those that should stay, because
      * IE looses the innerHTML of unattached elements, leading to empty elements within the container (e.g. missing
      * subtitle text in SubtitleLabel).
--- a/src/ts/components/Container.ts
+++ b/src/ts/components/Container.ts
@@ -112,6 +112,10 @@ export class Container<Config extends ContainerConfig> extends Component<Config>
 
   /**
    * Updates the DOM of the container with the current components.
+   *
+   * This is called automatically after construction. However, when you dynamically
+   * add or remove components at runtime, you must call `updateComponents()` to
+   * re-render the container’s children.
    */
   updateComponents(): void {
     /* We cannot just clear the container to remove all elements and then re-add those that should stay, because
--- a/src/ts/components/overlays/RecommendationOverlay.ts
+++ b/src/ts/components/overlays/RecommendationOverlay.ts
@@ -12,13 +12,17 @@ import { RecommendationItem } from '../RecommendationItem';
 export class RecommendationOverlay extends Container<ContainerConfig> {
 
   private static readonly CLASS_HAS_RECOMMENDATIONS = 'recommendations';
-  private readonly replayButton: HugeReplayButton;
-  private recommendationContainer: Container<ContainerConfig> | null;
+  readonly replayButton: HugeReplayButton;
+  readonly recommendationContainer: Container<ContainerConfig>;
 
   constructor(config: ContainerConfig = {}) {
     super(config);
 
     this.replayButton = new HugeReplayButton();
+    this.recommendationContainer = new Container({
+      components: [],
+      cssClasses: ['recommendation-overlay-row', 'recommendations-section']
+    });
 
     this.config = this.mergeConfig(config, {
       cssClass: 'ui-recommendation-overlay',
@@ -26,7 +30,8 @@ export class RecommendationOverlay extends Container<ContainerConfig> {
       components: [
         new Container({
           components: [
-            this.replayButton
+            this.replayButton,
+            this.recommendationContainer,
           ],
           cssClasses: ['recommendation-overlay-row', 'replay-section'],
         }),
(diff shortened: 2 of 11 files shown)


src/ts/spatialnavigation/NavigationGroup.ts
```diff
@@ -54,7 +54,7 @@ export class NavigationGroup {
   }
 
   // Dynamically resolve all components within this group respecting FocusableContainers.
-  protected get components(): Focusable[] {
+  protected getComponents(): Focusable[] {
```
stonko1994, line 57: I always got a typescript compile error when trying to use `super.components` in the new `RecommendationOverlayNavigationGroup`. I did some research but wasn't able to solve it. As workaround I changed this from a computed property to a function. Let me know if anyone has an idea how to keep that a computed property and override it and calling super.

src/ts/components/overlays/RecommendationOverlay.ts
```diff
@@ -12,21 +12,26 @@ import { RecommendationItem } from '../RecommendationItem';
...
 
     this.config = this.mergeConfig(config, {
       cssClass: 'ui-recommendation-overlay',
       hidden: true,
       components: [
         new Container({
           components: [
-            this.replayButton
+            this.replayButton,
+            this.recommendationContainer,
```
stonko1994, line 34: We now always add the `recommendationContainer` to have it available initially for spatial navigation.

src/ts/components/Container.ts
```diff
@@ -113,7 +113,7 @@ export class Container<Config extends ContainerConfig> extends Component<Config>
   /**
    * Updates the DOM of the container with the current components.
    */
-  protected updateComponents(): void {
+  updateComponents(): void {
```
stonko1994, line 120: The `add-` and `removeComponent` functions are already available. However, without the `updateComponents` added/removed Components will never update. This is needed to dynamically add/remove recommendations.

123mpozzi: How about adding the intended use-case for manual calls? Adding docs here would help distinguish between automatic and manual calls


```suggestion
/**
  * Updates the DOM of the container with the current components.
  * 
  * This is called automatically after construction. However, when you dynamically
  * add or remove components at runtime you must call `updateComponents()` to
  * re-render the container’s children.
  * Use only in those cases, it is not needed during normal UI lifecycle.
  */
  updateComponents(): void {
```

src/ts/spatialnavigation/FocusableContainer.ts
```diff
@@ -10,7 +10,7 @@ import { isFocusable } from './TypeGuards';
  */
 export class FocusableContainer {
   readonly container: AnyContainer;
-  readonly primaryComponent: AnyComponent;
+  readonly _primaryComponent: AnyComponent;
```
123mpozzi: Typescript supports `private`, what do you think about changing this field into using that?

```suggestion
  private readonly primaryComponentOverride?: AnyComponent;
```

src/ts/spatialnavigation/NavigationGroup.ts
123mpozzi: Not sure how we can achieve that.
But we could keep the computed property as well by splitting the logic?

```ts
/**
 * Computed list of all components within this group respecting FocusableContainers.
 * Prefer reading this property.
 * Subclasses should override `getComponents()` instead of this accessor.
 */
protected get components(): Focusable[] {
  return this.getComponents();
}

/**
 * Subclass hook for customizing which components are considered focusable.
 * Call `super.getComponents()` to start from the base set.
 */
protected getComponents(): Focusable[] {
  let componentsToConsider: Focusable[] = [];
  
  ...

  return componentsToConsider;
}
```

123mpozzi (commented):
I was able to lose focus by seeking to the end and having no elements focused. I used mouse, but maybe this can be achieved by spatial nav as well.
It happens when you have no recommendation items and the result is that you cannot focus any element, effectively losing the spatial navigation capabilities.

Is that something we should take care of, or do you think this state is not reachable by only spatial navigation interactions?

https://github.com/user-attachments/assets/fb142698-1b1f-4c92-85ad-2ceb1e1a9ab8

src/ts/components/Container.ts
stonko1994: [cb31f17](https://github.com/bitmovin/bitmovin-player-ui/pull/746/commits/cb31f176e40d451717e646f1c8101b1daf8131ee)

src/ts/spatialnavigation/FocusableContainer.ts
stonko1994: [3236db5](https://github.com/bitmovin/bitmovin-player-ui/pull/746/commits/3236db50a4660155e2a0d9033cde7a3474d40fcc)

stonko1994:
> I was able to lose focus by seeking to the end and having no elements focused. I used mouse, but maybe this can be achieved by spatial nav as well. It happens when you have no recommendation items and the result is that you cannot focus any element, effectively losing the spatial navigation capabilities.
> 
> Is that something we should take care of, or do you think this state is not reachable by only spatial navigation interactions?
> 
>  spatialNavRecommendations.mov

I was not able to reproduce this with pure spatial navigation. So I'd ignore this use case for now.

src/ts/spatialnavigation/NavigationGroup.ts
stonko1994: It would be possible but not sure it's worth the additional method. I also had a call with @Andr3wid about possible solutions and we decided to keep it as is.

123mpozzi (approved):

PR #747, opened 2025-09-01
Prepare CI/CD automation for v4 release

stonko1994:
## Description

This PR prepares our release autoisation for the upcoming v4 UI release. Our current release workflow isn't very flexible when it comes to releasing from specific branches. It will always release from development no matter where we add a new tag. This is a problem for the upcoming major version in case we want to release pre-release versions or back-ports for v3.

### Changes
- Allowing manual release triggers via GH actions.
- Passing in the release version for manual specific releases
- Supporting pre-releases (alpha, beta, rc)
- An option to run `dry-runs`
- Continues deployment is still enabled and uses the new release workflow now

### Test runs
- Successful relase: https://github.com/bitmovin/bitmovin-player-ui/actions/runs/17667635900
- Successful dry-run: https://github.com/bitmovin/bitmovin-player-ui/actions/runs/17668765379
- Successful CD merge:
  - With a minor version: https://github.com/bitmovin/bitmovin-player-ui/actions/runs/17669021670/job/50216365697
  - With a patch version: https://github.com/bitmovin/bitmovin-player-ui/actions/runs/17669110152/job/50216640826
  - With no release as no new changelog entry: https://github.com/bitmovin/bitmovin-player-ui/actions/runs/17669281444

- [x] `CHANGELOG` entry


.github/scripts/notifySlackTeam.js
```diff
@@ -15,16 +16,9 @@ fs.readFile(changelogPath, 'utf8', (err, fileContent) => {
...
 });
 
-function parseReleaseVersion(fileContent) {
-  const regex = /##\s\[(\d+\.\d+.\d+)\]/;
-  const releaseVersion = fileContent.match(regex);
-
-  return releaseVersion[1];
-}
 
 function parseChangelogEntry(fileContent) {
```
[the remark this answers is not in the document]
stonko1994: It is still used in line 18

publish.sh
```diff
@@ -65,14 +116,14 @@ set +e
...
 # after fixing the problem, the npm publish would fail the build.
-IS_PUBLISHED=$(npm view ${PACKAGE_NAME}@${VERSION} dist-tags)
+IS_PUBLISHED=$(npm view "${PACKAGE_NAME}@${VERSION_NUMBER}" dist-tags)
 set -e
 
 if [[ ${IS_PUBLISHED} ]]; then
-    echo "WARNING ${VERSION} is already published, skipping."
-    exit 0
+    echo "WARNING ${VERSION_NUMBER} is already published, skipping."
+    exit 1
```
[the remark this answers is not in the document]
stonko1994: IMO this should still be a failure as it's unexpected to release a version again

.github/actions/setup-dependencies/action.yml
```diff
@@ -0,0 +1,24 @@
+name: 'Setup Dependencies'
+description: 'Sets up Node.js, caches node modules, and installs dependencies'
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Set up node.js
+      uses: actions/setup-node@v4
+      with:
+        node-version-file: .nvmrc
```
rolandkakonyi, line 10: Why not use `cache:` here instead of the custom caching in the next step?
Docs: https://github.com/actions/setup-node?tab=readme-ov-file#usage

Probably this would do it:
```
cache: 'npm'
```

.github/workflows/ci.yml
rolandkakonyi, line 1: We should have a `concurrency` like this on the top level so runs are canceled if there is a push while it is still running.
Something like this:
```
concurrency:
  group: ci-${{ github.ref }}
  cancel-in-progress: true
```

.github/workflows/release.yml
```diff
@@ -1,84 +1,245 @@
 name: Release
-
-on: 
-  push:
-    tags:
-      - '!player/'
-      - 'v*'
+run-name: Release ${{ inputs.version_number }} ${{ inputs.release_type }}${{ inputs.dry_run && ' (dry run)' || '' }}
+
```
rolandkakonyi, line 3: Same here regarding concurrency, in case it gets re-triggered with same version, we should cancel!

.github/scripts/calculateVersionNumber.js
```diff
@@ -0,0 +1,70 @@
...
+    switch (releaseType) {
+      case 'alpha':
+        shortName = 'a';
+        break;
+      case 'beta':
+        shortName = 'b';
+        break;
+      case 'rc':
+        shortName = 'rc';
+        break;
```
felix-hoc: is this a common pattern on mobile? On web, pre-release versions are `beta`/`alpha`/`rc` (not shortened), e.g. `8.138.0-beta.2`

.github/actions/setup-dependencies/action.yml
stonko1994: I wasn't aware of this 🙌  [9f21622](https://github.com/bitmovin/bitmovin-player-ui/pull/747/commits/9f216229f69cd293cbec312f32ca0d4e5921a188)

.github/workflows/release.yml
stonko1994: [e55cc5e](https://github.com/bitmovin/bitmovin-player-ui/pull/747/commits/e55cc5e532980cb8e84ce506e8315a06f3b59a83)

.github/workflows/ci.yml
stonko1994: [e55cc5e](https://github.com/bitmovin/bitmovin-player-ui/pull/747/commits/e55cc5e532980cb8e84ce506e8315a06f3b59a83)

.github/actions/setup-dependencies/action.yml
stonko1994: The difference to before is that `npm ci` now always runs. The previous version was extracted from the existing workflow. I'm not super familiar with web development but I think running `npm ci` always is prefered anyways no?

rolandkakonyi: We should still run `npm ci` as this won't install project dependencies, only set up node + restore+save npm cache

rolandkakonyi (approved):

.github/scripts/calculateVersionNumber.js
stonko1994: Yes we use the shortened version on mobile. But I'm fine with being consistent to the Web Players convention for the UI.

felix-hoc: Ok, I'd vote for the full version then, to be consistent between web + web UI, and with our previous naming in the UI (https://github.com/bitmovin/bitmovin-player-ui/tags)

stonko1994: [17d8896](https://github.com/bitmovin/bitmovin-player-ui/pull/747/commits/17d8896bcb8455b87bbc7a69887b619ff32e7412)

.github/workflows/release.yml
```diff
@@ -1,84 +1,234 @@
...
+    secrets:
+      NPM_AUTH_TOKEN:
+        description: 'NPM authentication token for publishing'
+        required: true
+      GCS_CREDENTIALS:
+        description: 'Google Cloud Storage credentials for uploading docs'
+        required: true
+      SLACK_WEBHOOK_URL:
+        description: 'Slack webhook URL for notifications'
+        required: true
```
felix-hoc, line 54: why do we need these as inputs? We're anyway taking them from the secrets, no?

.github/workflows/release.yml
```diff
@@ -1,84 +1,234 @@
...
+
+concurrency:
+  group: release-${{ inputs.version_number }}-${{ inputs.release_type }}
+  cancel-in-progress: true
```
felix-hoc, line 58: do we really want to cancel in-progress releases? Just wondering if this could leave us in an inconsistent state

felix-hoc (commented):
LGTM in general 👍🏼  just a few more thoughts

.github/workflows/release.yml
stonko1994: To our knowledge you need to explicitly define them for the `workflow_call` trigger. Otherwise the workflow can't access secrets which are not explicitly defined.

.github/workflows/release.yml
stonko1994: Neither is ideal. It all depends on whether the `git command` was executed already or not. If they were not yet executed, the next run will just calculate the same version number again (e.g., if alpha, beta, or rc). If they were already executed, the next number will be taken or it will fail e.g. on a final release.

On the other hand, when we don't cancel in progress once, both runs could try to release the very same version at the same time (**a concurrency group is not a queue**).

.github/workflows/release.yml
felix-hoc: ok, wasn't aware of that 👍🏼

.github/workflows/release.yml
felix-hoc: Ok, yeah fair points. It anyway shouldn't happen that two releases are running at the same time.

felix-hoc (approved):

felix-hoc (approved):

PR #753, opened 2025-09-12
Modern UI: Fix accidental tap after scrolling in settings menu on mobile

123mpozzi:
## Description

An accidental tap is triggered when scrolling through the settings menu on mobile.

When scrolling through a list of `ui-settings-panel-item`s on mobile, the gesture incorrectly triggers a tap on the item where the scroll ends. This causes the user to be navigated into the item’s sub-menu (`ui-settings-panel-item-select-option`) even though no tap was made.

- [x] `CHANGELOG` entry - ` no need, this is a fix of an unreleased version of an upcoming UI`

### DIFF
--- a/src/ts/components/settings/InteractiveSettingsPanelItem.ts
+++ b/src/ts/components/settings/InteractiveSettingsPanelItem.ts
@@ -14,6 +14,13 @@ export class InteractiveSettingsPanelItem<Config extends SettingsPanelItemConfig
     onClick: new EventDispatcher<InteractiveSettingsPanelItem<Config>, NoArgs>(),
   };
 
+  private touchState = {
+    startX: 0,
+    startY: 0,
+    moved: false,
+    touchId: null as number | null,
+  };
+
   constructor(config: Config) {
     super(config);
   }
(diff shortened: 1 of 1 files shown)


src/ts/components/settings/InteractiveSettingsPanelItem.ts
```diff
@@ -27,7 +34,64 @@ export class InteractiveSettingsPanelItem<Config extends SettingsPanelItemConfig
       this.onClickEvent();
     };
 
-    this.getDomElement().on('click touchend', handleClickEvent);
```
stonko1994, line 30: Do you know why there is a `touchend` listener in first place? I remember now that I did something like this once to workaround a iOS performance issue. But this got fixed so this shouldn't be needed anymore.

Is there and issues when we simply just remove the `touchend` listener?

123mpozzi: By digging around it looks like it's a legacy workaround to handle the 300ms click delay that existed on iOS Safari and older Android browsers.

Before modern viewport meta tags, mobile browsers introduced a delay between `touchend` and `click` events to detect double-tap gestures. Developers used `touchend` to provide immediate response.

Modern mobile browsers eliminate 300 ms click delay if you set the viewport. So we don't need to use this workaround anymore I guess.

A few sources:
- https://developer.chrome.com/blog/300ms-tap-delay-gone-away
- https://stackoverflow.com/questions/12238587/eliminate-300ms-delay-on-click-events-in-mobile-safari

123mpozzi: So that's a great point!
In the end it looks like the fix to this issue is just about removing the legacy workaround - Address in 8d51da1f8be2f1692e1112d350530537efc5677a

stonko1994 (approved):

PR #763, opened 2025-09-18
Modern UI - Fix auto-hide timer not applied on Scrolling and when clicking buttons

123mpozzi:
## Description

Auto-hide timer should be applied on user interactions, but right now it is not during:
- button clicking (e.g. volume toggle button)
- scrolling (e.g. continuous scrolling in the renditions menu)

- [x] `CHANGELOG` entry - `I'd not include since this is a fix to an unreleased version`


stonko1994 (approved):
Works like a charm.

PR #764, opened 2025-09-18
Modern UI - Save and restore current settings menu state when auto-hiding

123mpozzi:
## Description

Improves the auto-hide behavior when the settings panel is open, addressing UX issues where the menu could vanish too quickly.

### Changes

- use a longer inactivity timeout if the settings panel is left open without interaction
- ensure that when controls reappear (on mouse move or interaction), the settings panel is restored in the same state it was left in
  - restore the last-used navigation menu
  - restore the scroll position

### UX Considerations

Three possible approaches for handling auto-hide with the settings panel were explored:
1. Keep auto-hide active with the same rules
 - ✅ consistent with the rest of the UI
 - ❌ frustrating if the menu disappears while browsing options
 - ❌ users may feel  "punished" for not moving quickly enough.
2. Disable auto-hide while settings panel is open
- ✅ removes frustration: panel stays until dismissed and user stays in control
- ❌ risk of clutter if the user opens the settings and moves away
3. Hybrid (chosen): longer auto-hide delay while interacting with settings and remember what was opened
- ✅ panel stays open while interacting
- ✅ longer inactivity timeout (5–10× normal) still prevents clutter
- ✅  opened menus can reappear on any interaction, with the same state
- ❌ slightly more complex, but best balance of usability + cleanliness

We went with hybrid because it balances both goals (no frustration with quick auto-hide, no cluttering with timer removed), delivering a smoother, more predictable user experience.

- [x] `CHANGELOG` entry


stonko1994 (commented):
I quite like the idea and the basic UX of this. However, a few feedback points:
- The Settings Panel resizes when moving the mouse out and in again
- The 10 seconds are too long in case you keep the mouse in the view. I'd go for something around 3-5 seconds.
- The 'stored' state should restore after X second when the UI isn't going to be visible again soon. Maybe after X seconds after the UI hides, if no interaction is happening again in that timeframe, we clear the state

123mpozzi:
Thanks for the feedback! 🙌 
 
> * The Settings Panel resizes when moving the mouse out and in again

Nice find! In b95d4d0007b39bb30fb182b7bde94b3d75a9f782 I implemented store/restore of panel width and height as well

> * The 10 seconds are too long in case you keep the mouse in the view. I'd go for something around 3-5 seconds.

I reduced the auto-hide delay to 4 seconds in 68873ab6238d1ecae82cd6d1bb9266a9bfd01600

> * The 'stored' state should restore after X second when the UI isn't going to be visible again soon. Maybe after X seconds after the UI hides, if no interaction is happening again in that timeframe, we clear the state

I agree. In 4cd8119b541ba8cbff6ee9184b9bec1b341015ba I apply a `15` seconds timeout after the component hides - after that timeout, the state is cleared

src/ts/components/settings/SettingsPanel.ts
```diff
@@ -459,4 +459,39 @@ export class SettingsPanel<Config extends SettingsPanelConfig> extends Container
...
+      this.getDomElement().css({
+        width: panelWidth + 'px',
+        height: panelHeight + 'px',
+      });
+    }
+
+    // Restore scroll positions after DOM is ready
+    setTimeout(() => {
+      this.getDomElement().get(0).scrollTop = scrollTop;
+      const wrapper = this.getDomElement().find('.bmpui-container-wrapper').get(0);
```
stonko1994: `bmpui` should not be hard-coded as it can be customized.

src/ts/components/settings/SettingsPanel.ts
```diff
@@ -459,4 +459,39 @@ export class SettingsPanel<Config extends SettingsPanelConfig> extends Container
...
+   * Restores the navigation state without triggering resetNavigation.
+   * Used to preserve the user's navigation position when reopening the panel.
+   */
+  public restoreNavigationState(
+    activePage: SettingsPanelPage,
+    navigationStack: SettingsPanelPage[],
+    scrollTop: number,
+    wrapperScrollTop: number,
+    panelWidth?: number,
+    panelHeight?: number,
```
stonko1994: Nit: Those are quite some properties already. Should we combine them into a helper struct? 🤔

src/ts/components/settings/SettingsPanelAutoHideManager.ts
```diff
@@ -0,0 +1,150 @@
...
+import { SettingsPanel, SettingsPanelConfig } from './SettingsPanel';
+import { SettingsPanelPage } from './SettingsPanelPage';
+import { ArrayUtils } from '../../utils/ArrayUtils';
+import { Component, ComponentConfig } from '../Component';
+import { Timeout } from '../../utils/Timeout';
+
+/**
+ * State interface for preserving settings panel navigation and scroll position
+ */
+interface SettingsPanelState {
```
stonko1994: Oh the interface already exists (follow up from comment above)

src/ts/components/settings/SettingsPanelAutoHideManager.ts
```diff
@@ -0,0 +1,150 @@
...
+export class SettingsPanelAutoHideManager {
+  private openSettingsPanels: SettingsPanel<SettingsPanelConfig>[] = [];
+  private lastOpenSettingsPanel: SettingsPanel<SettingsPanelConfig> | null = null;
+  private lastSettingsPanelState: SettingsPanelState | null = null;
+  private stateClearTimeout: Timeout;
+  private readonly stateClearDelay: number;
+
+  constructor(
+    private uimanager: UIInstanceManager,
+    stateClearDelay: number = 15000,
```
stonko1994: To me the 15 seconds feel quite high. I played around with the value and I think that around 5 seconds feel about right to me.

Nit: Ideally we put that on the `SettingsPanelConfig` as customization option.

src/ts/components/settings/SettingsPanelAutoHideManager.ts
```diff
@@ -0,0 +1,150 @@
...
+      // Cancel the state clear timeout since we're restoring the state
+      if (this.stateClearTimeout) {
+        this.stateClearTimeout.clear();
+      }
+
+      // Show the panel first
+      panel.show();
+
+      // Then restore the navigation state (this will override the resetNavigation call in onShow)
+      setTimeout(() => {
```
stonko1994: Why do we need this async after 0?

src/ts/components/settings/SettingsPanelAutoHideManager.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Returns the appropriate hide delay - uses panel's hideDelay if open, normal otherwise
+   */
+  public getExtendedDelay(baseDelay: number): number {
+    if (!this.hasOpenPanels()) return baseDelay;
+
+    const openPanel = this.openSettingsPanels[0];
+    return openPanel.getConfig().hideDelay;
+  }
```
stonko1994: This method feels a bit weird to me. It only checks if we have an open settings panel and us the `SettingsPanel's` hide delay. Though the name really doesn't indicate that.

What do you think about the following:
- rename this to something like `getOpenSettingsPanelHideDelay`
- return an optional number
- use the base delay on the caller side as fallback if returning nil here.

src/ts/components/UIContainer.ts
```diff
@@ -126,18 +130,28 @@ export class UIContainer extends Container<UIContainerConfig> {
...
         // Let subscribers know that they should reveal themselves
         uimanager.onControlsShow.dispatch(this);
         isUiShown = true;
       }
       // Don't trigger timeout while seeking (it will be triggered once the seek is finished) or casting
       if (!isSeeking && !player.isCasting() && !this.hidingPrevented()) {
+        // Use extended timer if settings panel is open, normal timer otherwise
+        const hideDelay = this.settingsPanelManager.getExtendedDelay(config.hideDelay);
+        this.uiHideTimeout.clear();
+        this.uiHideTimeout = new Timeout(hideDelay, this.hideUi);
```
stonko1994: I think assigning a new `Timeout` here is super dangerous. After this code path was executed for the first time we basically change the hide timout of our complete UI 👀 It is never reset to the original hide delay without an open settings panel.

src/ts/components/UIContainer.ts
```diff
@@ -327,6 +341,8 @@ export class UIContainer extends Container<UIContainerConfig> {
         this.uiHideTimeout.start();
       }
     });
+
+    // Settings panel auto-hide management is now handled by SettingsPanelAutoHideManager
```
stonko1994: This comment is unnecessary.

src/ts/components/UIContainer.ts
```diff
@@ -62,6 +63,8 @@ export class UIContainer extends Container<UIContainerConfig> {
   private userInteractionEvents: { name: string; handler: EventListenerOrEventListenerObject }[];
   private hidingPrevented: () => boolean;
 
+  private settingsPanelManager: SettingsPanelAutoHideManager;
```
stonko1994: Any particular reason you put that on the `UIContainer` and not on the `UIManager`? 🤔

stonko1994 (commented):
Manual testing outcome:
- The UX feels really really nice when tweaking the timeout.
- Issues:
  - I found an issue when resizing the window (basically when switching UIVariant) while a SettingsPanel is open. On my device the UI starts to flicker and no settings panel will open anymore.

https://github.com/user-attachments/assets/3149ddf8-e7f5-4cf3-91f1-436cd0a09f43

src/ts/components/settings/SettingsPanelAutoHideManager.ts
123mpozzi: Address in 84c24cdfec9b13f2d75696693ef671ce1a56bcaf

src/ts/components/UIContainer.ts
123mpozzi: We would still need to assign a new `TimeOut` since its `delay` is read-only by design.
In 84c24cdfec9b13f2d75696693ef671ce1a56bcaf I implemented the reset to base hide delay when no settings panel is opened

src/ts/components/UIContainer.ts
123mpozzi: Removed in 52e4a662ee4f7c71beedc610ae9b92f9c126695e

src/ts/components/settings/SettingsPanelAutoHideManager.ts
123mpozzi: Moved the property under the config in 52e4a662ee4f7c71beedc610ae9b92f9c126695e

src/ts/components/settings/SettingsPanelAutoHideManager.ts
123mpozzi: Ah right! Changed the code to use the interface in 6f1ff90337abe64133ea9beec8279a7dd12a156c

src/ts/components/settings/SettingsPanel.ts
123mpozzi: Address in 6f1ff90337abe64133ea9beec8279a7dd12a156c

src/ts/components/settings/SettingsPanel.ts
123mpozzi: Address in 6f1ff90337abe64133ea9beec8279a7dd12a156c

src/ts/components/settings/SettingsPanelAutoHideManager.ts
123mpozzi: Actually, `SettingsPanelConfig` is the wrong place.

`SettingsPanelAutoHideManager` manages multiple settings panel with one delay, and the delay is not about individual panels, but about how long the UI remembers state after hiding.
  
  I think the right place is `UIConfig` since `stateClearDelay` is a global UI-level behavior. It applies to every UI instance.
  
  --- 
  
  I moved it again in 17cdc2b390bf547de07b79d447998515019656aa

123mpozzi: > To me the 15 seconds feel quite high. I played around with the value and I think that around 5 seconds feel about right to me.

  I've changed the delay to 5 seconds in 7b324b404185b6049d5d3fa3ce124f9d3b9e9b2f

src/ts/components/UIContainer.ts
123mpozzi: I've initially placed it here because only `UIContainer` has auto-hide logic that needs extended delays and state restoration.

However, I see it make sense to move in `UIManager` since we follow a similar pattern for subtitle manager - Address in 1ea0918ff59058b30d74a2d4d75f896c275b58b0

src/ts/components/settings/SettingsPanelAutoHideManager.ts
123mpozzi: Good catch! `setTimeout(.., 0)` is a common pattern when you need to defer execution until after event handlers have fully completed. Without it, we'd be fighting against the panel's own `onShow` logic.

 It ensures the order of operations when showing the panel:

  1. `panel.show()` triggers the panel's `onShow` event handler
  2. The `SettingsPanel`'s `onShow` handler calls `resetNavigation(true)`, which:
    - Clears the navigation stack
    - Resets to the root page
    - Updates CSS classes and DOM dimensions

  If we called `panel.restoreNavigationState(state)` synchronously right after `panel.show()`, it would execute in the same call stack as the `resetNavigation()` call, potentially causing conflicts with DOM updates and animations that are still in progress.

  The `setTimeout(.., 0)` pushes the restoration to the next event loop tick.

  
  Sources:
  - https://imdeveshshukla.medium.com/settimeout-with-time-0-explained-967f3133e305
  - https://stackoverflow.com/a/779785
  - https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#delay

123mpozzi:
> Manual testing outcome:
> 
> * The UX feels really really nice when tweaking the timeout.
> * Issues:
>   
>   * I found an issue when resizing the window (basically when switching UIVariant) while a SettingsPanel is open. On my device the UI starts to flicker and no settings panel will open anymore.
> 
>  ui-settings-panel.mov

Thanks for the video! 🙌 

I see, issue is `SettingsPanelAutoHideManager` is tracking panels across UI variant switches and trying to restore panels from a different UI instance, and different UI variants have different default auto-hide delay for the settings panel in `UIFactory`

It should be fixed as of 3fbe45d061f1c00295db9d6c39fc72fb1481286a

src/ts/components/settings/SettingsPanelAutoHideManager.ts
stonko1994: Would it be possible to add another `.onShow` listener to the panel? This can be compared with an `DispatchQueue.async.main` in iOS which we try to only use when absolutely necessary. Code like this is very prone for race-conditions so we should try to avoid it at any costs.

src/ts/components/settings/SettingsPanel.ts
stonko1994: wouldn't it be easier to expose the `innerContainerElement` from `Container` instead of trying to 'find' it again? 🤔 It will be always there and it might be a bit cleaner on the usage side.

You could just write `this.innerContainerElement` here? (we just need to make the `innerContainerElement` `protected` instead of `private`). Then we can also revert all of those `getWrapperClassName` which feels wrong to me because it looks a bit like a hack (our component structure shouldn't need to do class matching to access stuff)

src/ts/components/UIContainer.ts
stonko1994: Comment here: https://github.com/bitmovin/bitmovin-player-ui/commit/84c24cdfec9b13f2d75696693ef671ce1a56bcaf#r167619568

src/ts/components/settings/SettingsPanel.ts
123mpozzi: Yes, that change makes a lot of sense! Address in 9bf4f077bc20778aa9f332abdc7234f687d32494

src/ts/components/UIContainer.ts
123mpozzi: Thanks for linking the comment! Address in 0b342a5b6e41f2858855ac95405b3ccc0faf02b3

src/ts/components/settings/SettingsPanelAutoHideManager.ts
123mpozzi: That's a really valid concern. Address in 751a664577bd8ba249e0055cb62b7ba224658c28

stonko1994 (commented):
As discussed offline let's change the approach to not try to apply the state retro-active but don't reset it in first place.

123mpozzi:
> As discussed offline let's change the approach to not try to apply the state retro-active but don't reset it in first place.

I've refactored in a new PR, #785 , since the changes were mostly independent

PR #765, opened 2025-09-18
Use 'latest' NPM tag to gather current version instead of pre-released versions

stonko1994:
## Description

The [first beta release](https://github.com/bitmovin/bitmovin-player-ui/actions/runs/17830558389) workflow failed as the latest version is fetched with the release type. There was no beta version first to the command returned `null` which caused a failure further down the line. 

Using `latest` will ensure that the latest stable version is used for the comparison. This will be needed for v3 back-port releases going forward.

- [x] `CHANGELOG` entry **Not needed for an ad-hoc workflow fix**

### DIFF
--- a/publish.sh
+++ b/publish.sh
@@ -130,7 +130,7 @@ fi
 echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" > ~/.npmrc
 chmod 0600 ~/.npmrc
 
-NPM_LATEST=$(npm view --json ${PACKAGE_NAME} dist-tags | jq -r ".${NPM_TAG}")
+NPM_LATEST=$(npm view --json ${PACKAGE_NAME} dist-tags | jq -r ".latest")
 echo "INFO latest npm version is $NPM_LATEST"
 
 # We always publish the package with the channel/latest tag because there is no way to publish a package without
--- a/publish.sh
+++ b/publish.sh
@@ -130,14 +130,19 @@ fi
 echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" > ~/.npmrc
 chmod 0600 ~/.npmrc
 
-NPM_LATEST=$(npm view --json ${PACKAGE_NAME} dist-tags | jq -r ".latest")
+NPM_LATEST=$(npm view --json ${PACKAGE_NAME} dist-tags | jq -r ".${NPM_TAG}")
 echo "INFO latest npm version is $NPM_LATEST"
 
 # We always publish the package with the channel/latest tag because there is no way to publish a package without
 # a tag (the default tag is always "latest"). If the published version is older that the currently tagged version,
 # we have to revert the tag afterwards to avoid version regressions.
-echo "INFO publishing ${VERSION_NUMBER} to npm with tag '${NPM_TAG}' (current tagged version is ${NPM_LATEST})"
+echo "INFO publishing ${VERSION_NUMBER} to npm with tag '${NPM_TAG}' (current tagged version is ${NPM_LATEST:-"none"})"
 npm publish --tag ${NPM_TAG} ${NPM_DRY_RUN_CMD}
+
+# If there is no previously tagged version (e.g. for alpha/beta/rc releases) we don't need to do anything
+if [[ "$NPM_LATEST" == "null" ]]; then
+  exit 0
+fi
 
 # Checks if one version is greater than the other
 # https://stackoverflow.com/a/24067243/370252
--- a/publish.sh
+++ b/publish.sh
@@ -130,7 +130,7 @@ fi
 echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" > ~/.npmrc
 chmod 0600 ~/.npmrc
 
-NPM_LATEST=$(npm view --json ${PACKAGE_NAME} dist-tags | jq -r ".${NPM_TAG}")
+NPM_LATEST=$(npm view --json ${PACKAGE_NAME} dist-tags | jq -r ".${NPM_TAG} // empty")
 echo "INFO latest npm version is $NPM_LATEST"
 
 # We always publish the package with the channel/latest tag because there is no way to publish a package without
@@ -140,7 +140,7 @@ echo "INFO publishing ${VERSION_NUMBER} to npm with tag '${NPM_TAG}' (current ta
 npm publish --tag ${NPM_TAG} ${NPM_DRY_RUN_CMD}
 
 # If there is no previously tagged version (e.g. for alpha/beta/rc releases) we don't need to do anything
-if [[ "$NPM_LATEST" == "null" ]]; then
+if [[ -z "$NPM_LATEST" ]]; then
   exit 0
 fi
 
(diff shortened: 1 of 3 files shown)


publish.sh
```diff
@@ -130,7 +130,7 @@ fi
 echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" > ~/.npmrc
 chmod 0600 ~/.npmrc
 
-NPM_LATEST=$(npm view --json ${PACKAGE_NAME} dist-tags | jq -r ".${NPM_TAG}")
+NPM_LATEST=$(npm view --json ${PACKAGE_NAME} dist-tags | jq -r ".latest")
```
123mpozzi: I see the original `jq` command returned `null` because there was no `beta` key in the `dist-tags` yet, since  it is the first time we use that channel.

Hard-coding `latest` avoids the `null` but changes behaviour: every prerelease (`beta`, `staging`, etc.) would start comparing itself against the stable `latest` tag.
That breaks the intention of the script, which is to keep each channel independent (so a beta compares against the current beta, not against stable).

The bigger risk is around the tag reversion logic later in the script:

```sh
if version_gt "$NPM_LATEST" "$VERSION_NUMBER"; then
    echo "INFO reverting '${NPM_TAG}' tag from the just published version ${VERSION_NUMBER} to the greater ${NPM_LATEST}"
    if [[ $NPM_DRY_RUN != 1 ]]; then
        sleep 10
        npm dist-tag add ${PACKAGE_NAME}@${NPM_LATEST} ${NPM_TAG}
    fi
fi
```

If we force everything to use `latest`, then:
- A beta or RC publish would compare itself to the stable release
- If the stable release is newer than the prerelease (which it usually is), the script would think a 'downgrade' happened
- It would then try to revert the prerelease tag back to the stable version, effectively overwriting or undoing the beta/staging tag we just published

That’s a version regression risk: consumers installing `@beta` or `@staging` could suddenly get the stable build instead of the prerelease.

A safer fix is to keep using the dynamic tag but handle the missing-key case explicitly:

```suggestion
NPM_LATEST=$(
  npm view --json "${PACKAGE_NAME}" dist-tags |
  jq -r ".${NPM_TAG} // empty"
)

if [[ -z "$NPM_LATEST" ]]; then
  echo "INFO no existing '${NPM_TAG}' tag found on npm"
  # sentinel lower than any real version so version_gt never triggers
  NPM_LATEST="0.0.0-zzzzz"
fi
```

This sentinel solves the first-time tag issue by acting as a 'fake' super-low version, so the comparison never falsely triggers a revert.

This way:
- Stable releases still compare against the current latest
- Pre-releases compare against their own channel
- First-ever channel publishes no longer fail (they just see "no prior version")
- We avoid false-positive 'downgrades' that would wrongly revert tags

---

With the new `support/v3.x` branch, this becomes even more important: if we keep using the same tags (`latest`, `beta`, `staging`), v3 backports will always be compared against v4's tags:

E.g.:
- v4 is on `4.0.1`
- v3 needs to publish backport `3.103.1`
- `version_gt` comparison sees `4.0.1 > 3.103.1` and reverts the tag -> the v3 publish is ignored.

In the future, we might want to introduce separate tags (`latest-v3`, `beta-v3`, ..) so each major line can evolve independently without interfering with each other.

123mpozzi (changes requested):

publish.sh
stonko1994: I think the differnt tracks for v3 and v4 are not an issue and over the top for now as we don't plann to actively do feature development on both. In general pre-release version are rare in the UI. Anyway I improved the handling in case there is no previous tagged version:

[c5bb7cb](https://github.com/bitmovin/bitmovin-player-ui/pull/765/commits/c5bb7cb15a4d22a9182191d5aa75898fb39c6989)

stonko1994: Test run: https://github.com/bitmovin/bitmovin-player-ui/actions/runs/17856901800

publish.sh
```diff
@@ -136,9 +136,14 @@ echo "INFO latest npm version is $NPM_LATEST"
...
 # we have to revert the tag afterwards to avoid version regressions.
-echo "INFO publishing ${VERSION_NUMBER} to npm with tag '${NPM_TAG}' (current tagged version is ${NPM_LATEST})"
+echo "INFO publishing ${VERSION_NUMBER} to npm with tag '${NPM_TAG}' (current tagged version is ${NPM_LATEST:-"none"})"
 npm publish --tag ${NPM_TAG} ${NPM_DRY_RUN_CMD}
 
+# If there is no previously tagged version (e.g. for alpha/beta/rc releases) we don't need to do anything
+if [[ "$NPM_LATEST" == "null" ]]; then
+  exit 0
+fi
+
```
123mpozzi, line 146: Nit: since this guard is to just prevent the reverting, we could just skip the revert logic instead of exiting.
This way, if someone adds code after the revert block later, it'll not be accidentally skipped.

```suggestion
# Only attempt a revert if we have a real prior version
if [[ -n "$NPM_LATEST" ]]; then
# revert logic
fi
```

publish.sh
```diff
@@ -136,9 +136,14 @@ echo "INFO latest npm version is $NPM_LATEST"
 # We always publish the package with the channel/latest tag because there is no way to publish a package without
 # a tag (the default tag is always "latest"). If the published version is older that the currently tagged version,
 # we have to revert the tag afterwards to avoid version regressions.
-echo "INFO publishing ${VERSION_NUMBER} to npm with tag '${NPM_TAG}' (current tagged version is ${NPM_LATEST})"
+echo "INFO publishing ${VERSION_NUMBER} to npm with tag '${NPM_TAG}' (current tagged version is ${NPM_LATEST:-"none"})"
```
123mpozzi, line 139: This won't print `none`, but `null` - as can be seen in the [test run](https://github.com/bitmovin/bitmovin-player-ui/actions/runs/17856901800/job/50777984483).

That's because

123mpozzi, line 139: This won't print `none`, but `null` - as can be seen in the [test run](https://github.com/bitmovin/bitmovin-player-ui/actions/runs/17856901800/job/50777984483).

That's because `${NPM_LATEST:-"none"}` only uses the default if the variable is unset/empty. But the string "null" is non-empty, so it will print `null`, not `none`.

This snippet should fix it by returning an empty string in case `jq` does not find the tag:
```sh
# Get current version for this channel; return empty if missing
NPM_LATEST=$(npm view --json "${PACKAGE_NAME}" dist-tags | jq -r ".${NPM_TAG} // empty")

echo "INFO publishing ${VERSION_NUMBER} to npm with tag '${NPM_TAG}' (current tagged version is ${NPM_LATEST:-"none"})"
```

Testing for emptiness should also be safer since right now we're relying on `jq -r` returning exactly `null`.

123mpozzi (approved):
Good job on fixing the first-time tag case! 💪 

Just left a few comments - right now `none` will never be printed in the `echo` statement

publish.sh
stonko1994: [5bf56c3](https://github.com/bitmovin/bitmovin-player-ui/pull/765/commits/5bf56c3d6d02fae5d798baeb5f5dfcece3bbd461)

PR #767, opened 2025-09-19
Improve SCSS variable naming/documentation

stonko1994:
## Description

Our SCSS variables have no documentation which makes it hard to understand where they are used exactly. We mention some of them in our migration guide, and to simplify the customization of if this PR adds documentation to them. In addition, I renamed some of them to better fit the usage.

- [x] `CHANGELOG` entry ****


src/scss/_variables.scss
```diff
@@ -1,44 +1,134 @@
...
 $color-background-menu-hover-overlay: rgba(255, 255, 255, .07) !default;
+/**
+ * Background color for quick seek indicator
+ */
 $color-background-seek-circle: rgba(255, 255, 255, .35) !default;
+/**
+ * Background color for dimmed overlay effects
+ */
 $color-background-dimmed: rgba(0, 0, 0, .5) !default;
-$color-shadow-seek-label: 0 0 30px 0 rgba(0, 0, 0, .75) !default;
```
stonko1994: We don't need this shadow effect and the CSS property value was invalid anyways.

src/scss/_variables.scss
```diff
@@ -1,44 +1,134 @@
...
+ */
 $font-size: 1.125rem !default; /* 18px */
+/**
+ * Large font size variant
+ */
 $font-size-large: calc($font-size * 1.25) !default;
+/**
+ * Small font size variant
+ */
 $font-size-small: calc($font-size * 0.75) !default;
```
felix-hoc, line 81: what about the SCSS linting issues - should we fix them or maybe update the rules?

felix-hoc (approved):
cool 🙌🏼

src/scss/_variables.scss
stonko1994: They will be gone after #769

PR #771, opened 2025-09-25
Use webpack instead of gulp

dweinber:
## Description
**Please run `npm ci` before testing this branch, or after this was merged**

This PR removes gulp and all of its dependencies in favor of Webpack 5.
All relevant actions are now available as npm scripts (like `npm run start`, `npm run build`, `npm run build:prod`).

For the output files, there should be little to no difference compared to before, except for one change:
The output contains now `dist/js/bitmovinplayer-ui.js` and `dist/css/bitmovinplayer-ui.css` (which are minified and source maps are provided along with them), but NO `dist/js/bitmovinplayer-ui.min.js` and `dist/css/bitmovinplayer-ui.min.css` anymore.
As a consequence, the non-minifed built files are not distributed anymore.
The reason is that webpack does minification and so on as part of the process (if in `production` mode), not after the actual build process as it was with gulp.

previously:
```
|---css/
|     |-bitmovinplayer-ui.css // non-minified
|     |-bitmovinplayer-ui.min.css // minified
|     |-bitmovinplayer-ui.min.css.map // source map to view the minifed as unminified
|
|---js/
|     |-bitmovinplayer-ui.js // non-minified
|     |-bitmovinplayer-ui.min.js // minified
|     |-bitmovinplayer-ui.min.js.map // source map to view the minifed as unminified
```
now:
```
|---css/
|     |-bitmovinplayer-ui.css // minified
|     |-bitmovinplayer-ui.min.css.map // source map to view the minifed as unminified
|
|---js/
|     |-bitmovinplayer-ui.js // minified
|     |-bitmovinplayer-ui.min.js.map // source map to view the minifed as unminified
```

I don't think there's any concern with this change but happy to discuss if needed.

- [x] `CHANGELOG` entry


.github/workflows/ci.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         uses: ./.github/actions/setup-dependencies
 
       - name: Build
-        run: npx gulp build
+        run: npm run build
```
stonko1994, line 41: Can't comment in lin 27 but should we change that to `npm run test`? is there a difference between using `npm run` and without `run`?

package.json
```diff
@@ -2,10 +2,21 @@
   "name": "bitmovin-player-ui",
   "version": "4.0.0-alpha.1",
   "description": "Bitmovin Player UI Framework",
-  "main": "./dist/js/framework/main.js",
+  "main": "./dist/js/bitmovinplayer-ui.js",
```
stonko1994: I guess that's why the JS files were in there? as it links to the `main.js`?

package.json
```diff
@@ -2,10 +2,21 @@
...
   "types": "./dist/js/framework/main.d.ts",
   "scripts": {
     "test": "npx jest",
+    "start": "webpack-dev-server --mode development --open",
+    "build": "npm run clean; npx webpack --mode development",
+    "build:prod": "npm run clean; npm run lint && npx webpack --mode production && npm run copy-json && npm run docs",
+    "lint": "npm run lint-ts && npm run lint-sass",
+    "format": "npx prettier --write *",
+    "docs": "npx typedoc",
+    "clean": "rm -r dist",
```
stonko1994: Nit: just to be sure it always works

```suggestion
    "clean": "rm -rf dist",
```

README.md
```diff
@@ -26,23 +26,24 @@ The UI framework is also available in the NPM repository and comes with all sour
...
 
-- `gulp` to build project into `dist` directory
-- `gulp watch` to develop and rebuild changed files automatically
-- `gulp serve` to open test page in browser, build and reload changed files automatically
-- `gulp lint` to lint TypeScript and SASS files
-- `gulp build-prod` to build project with minified files into `dist` directory
+- `npm start` to open test page in browser, build and reload changed files automatically
+- `npm run build` to build project into `dist` directory without minification
+- `npm run lint` to lint TypeScript and SASS files
+- `npm run build:prod` to build project with minified files into `dist` directory
```
stonko1994: Nit: I'd swap the two lines to have the build commands next to each other:

```suggestion
- `npm run build:prod` to build project with minified files into `dist` directory
- `npm run lint` to lint TypeScript and SASS files
```

README.md
```diff
@@ -26,23 +26,24 @@ The UI framework is also available in the NPM repository and comes with all sour
...
+- `npm test` to run the test suite
+- `npm run format` to format all files with prettier
+- `npm run docs` to generate the HTML documentation
 
-To take a look at the project, run `gulp serve`. For changes, check our [CHANGELOG](CHANGELOG.md). This UI framework version is for player v8. The UI framework for player v7 can be found in the `support/v2.x` branch.
+To take a look at the project, run `npm run start`. For changes, check our [CHANGELOG](CHANGELOG.md). This UI framework version is for player v8. The UI framework for player v7 can be found in the `support/v2.x` branch.
```
stonko1994:
```suggestion
To take a look at the project, run `npm run start`. For changes, check our [CHANGELOG](CHANGELOG.md). This UI framework version is for player v8.
```

webpack.config.js
```diff
@@ -0,0 +1,128 @@
+const path = require('path');
+
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+
+/** Modify this to your needs */
+const outputnames = {
+  globalNamespace: process.env.OUTPUT_NAMESPACE || 'bitmovin.playerui',
+  filename: process.env.OUTPUT_FILENAME || 'bitmovinplayer-ui',
+  cssPrefix: process.env.OUTPUT_CSS_PREFIX || 'bmpui',
```
stonko1994, line 12: Customizing the `cssPrefix` doesn't work as the bmpui value in the `_variables.scss` isn't replaced. As discussed offline I think we can change the value to `{{PREFIX}}` but also need to add the string replacmenent for `scss` files.

stonko1994, line 12: How is the `process.env.OUTPUT_CSS_PREFIX` working? I couldn't manage to pass the values from the cli

.github/workflows/ci.yml
dweinber: TL;DR:
Both, `npm run test` and `npm test` work.

NPM has defined [lifecycle scripts](https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-operation-order). Only those can be run without the `run`. So you can do [`npm test`](https://docs.npmjs.com/cli/v8/commands/npm-test) or [`npm start`](https://docs.npmjs.com/cli/v8/commands/npm-start), but you can't do `npm build`, for example. That would trigger a `Unknown command` error.

package.json
dweinber: Ah 😅 you might be right! I can revert this change and bring the JS files back?

package.json
dweinber: Shouldn't be necessary, but sure, why not 😄 done in a54c3e19

README.md
dweinber: Good point 😅 thanks!

webpack.config.js
dweinber: You pass them as environment variables for node, so e.g.:
```
OUTPUT_NAMESPACE="my.player.ui" OUTPUT_FILENAME="myui" OUTPUT_CSS_PREFIX="myplayerui" npm run build
```

dweinber: Thanks, missed that again 😬 looking into it.

dweinber: We could also change it a bit so that it works like this:
```
npm run build -- --env=cssPrefix=myplayerui --env=outputFilename=myui --env=namespace=my.player.ui
```

I've no strong opinion

dweinber: Fixed in abb7b28f

dweinber: Added a comment in abb7b28f to the webpack.config.js on how to do this, and changed the comment in _variables.scss to point to webpack.config.js for details.

package.json
stonko1994: I don't know what's the difference 🤔 Maybe the CMD opening of types works different then?

stonko1994 (approved):

package.json
dweinber: I took a closer look at this setup, and looks like if you use npm and a bundler to build your app, imports from the `bitmovin-player-ui` always take the individual, required framework/*.js files. The bitmovinplayer-ui.js file might be mainly used for jsdelivr and similar things. 
Changes are in c1f6a4a2

PR #772, opened 2025-09-25
Remove subtitletogglebutton

dweinber:
## Description
Removes the SubtitleToggleButton again, as it's currently not fully working as expected and won't make it to the upcoming release.

No changelog entry, I believe this was never added there anyway.

- [x] `CHANGELOG` entry


src/ts/utils/SubtitleUtils.ts
```diff
@@ -4,7 +4,11 @@ import { PlayerAPI, SubtitleEvent, SubtitleTrack } from 'bitmovin-player';
 import { i18n } from '../localization/i18n';
 import { StorageUtils } from './StorageUtils';
 import { prefixCss } from '../components/DummyComponent';
-import { StoredSubtitleLanguage } from '../components/buttons/SubtitleToggleButton';
+
+export interface StoredSubtitleLanguage {
```
stonko1994: Let's also remove the `setSubtitleLanguageStorage` for now. We can always add it again when we implement this.

dweinber: Sounds good, done in c89b4144

stonko1994 (approved):

PR #775, opened 2025-09-30
Add subtitle-only UI

dweinber:
## Description
Adds a new method to the `UIFactory`, `UIFactory.buildSubtitleUI`, that only contains the `SubtitleOverlay` to render subtitles (and what's needed to be able to customize them). This should make it easier for customers to get subtitle cues being displayed with custom (non-Bitmovin) UIs.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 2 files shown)


luckygoyal-bitmovin:
Thanks @dweinber . I tested this change with RN test app and it works very well to render subtitles only view.

stonko1994 (approved):
As discussed offline let's tackle the invisible SettingsPanel requirement later 👍 Manual testing looks good.

PR #776, opened 2025-10-01
Fix JS UI framework files

dweinber:
## Description
In the generated JS files of the UI Framework (dist/js/framework/), the placeholder for the (CSS) prefix was not correctly replaced, leading to errors. This is fixed now in a webpack postprocessing plugin.

- [x] `CHANGELOG` entry


webpack.config.js
```diff
@@ -135,3 +147,14 @@ module.exports = (env, { mode }) => {
     target: ['web', 'es5'],
   };
 };
+
+async function createJavascriptUiFrameworkFilesWithReplacedPrefix() {
+  await exec('npx tsc'); // run the typescript compiler to generate the dist-tsc folder
+  const options = {
+    templatingOn: false, // disable liquid syntax to avoid issues with {{ and }}
```
stonko1994, line 154: Is there a way to use this built-in feature for this kind of replacement? But I assume that we use this approach on other places as well it could cause some issues there 🤔

package.json
```diff
@@ -8,12 +8,11 @@
     "test": "npx jest",
     "start": "webpack-dev-server --mode development --open",
     "build": "npm run clean; npx webpack --mode development",
-    "build:prod": "npm run clean; npm run lint && npx webpack --mode production && npx tsc && npm run copy-json && npm run docs",
+    "build:prod": "npm run clean; npm run lint && npx webpack --mode production && npm run docs",
     "lint": "npm run lint-ts && npm run lint-sass",
     "format": "npx prettier --write *",
     "docs": "npx typedoc",
-    "clean": "rm -rf dist",
-    "copy-json": "mkdir -p dist/js/framework/localization/languages && cp src/ts/localization/languages/*.json dist/js/framework/localization/languages/",
```
stonko1994, line 16: Where is the `copy-json` now handled? I can't spot the replacement.

dweinber: This is done automatically and directly by the `npx tsc` command (in `webpack.config.js#createJavascriptUiFrameworkFilesWithReplacedPrefix`).

webpack.config.js
dweinber: Do you mean the liquid templating? Given our very little use of templates, I felt it's not worth looking into this further.

webpack.config.js
```diff
@@ -135,3 +147,14 @@ module.exports = (env, { mode }) => {
     target: ['web', 'es5'],
   };
 };
+
+async function createJavascriptUiFrameworkFilesWithReplacedPrefix() {
+  await exec('npx tsc'); // run the typescript compiler to generate the dist-tsc folder
```
felix-hoc: is this only required for creating the output folder? If so, could we not just exec `mkdir -p <OUTPUT_ROOT_DIRECTORY>`?

dweinber: This invokes the typescript compiler (`tsc`) to create the individual JS files in the output directory (`dist/js/framework/`), which are used when you use the npm package with `import`. Webpack only creates the bundled JS files (plus CSS etc) like `dist/js/bitmovinplayer-ui.js`

dweinber: I guess the comment is a bit misleading... What about this?

```suggestion
  await exec('npx tsc'); // run the typescript compiler to generate the individual UI Framework JS files in `dist/js/framework/`
```

felix-hoc: Yes, that sounds good :)

felix-hoc (approved):

stonko1994 (approved):

PR #782, opened 2025-10-08
Fix UI v4 causing JS exception on Android

123mpozzi:
## Description

UI v4 causes a crash on Android: 

```log
Uncaught TypeError: e.isLinearAdActive is not a function
```

This is due to player interface API differences between the platforms: `isLinearAdActive` does not exist for Android.

### Changes

- Fix the crash in `Seekbar.ts` by adding a safeguard for the method existance
- Add a safeguard on the other usage of `isLinearAdActive` as well

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/seekbar/SeekBar.ts
+++ b/src/ts/components/seekbar/SeekBar.ts
@@ -129,9 +129,6 @@ export class SeekBar extends Component<SeekBarConfig> {
   private player: PlayerAPI;
   private uiManager: UIInstanceManager;
 
-  // We need to track the current ad state manually since our mobile SDKs don't expose an API to check if an ad is active.
-  private isAdPlaying: boolean = false;
-
   protected seekBarType: SeekBarType;
 
   protected isUiShown: boolean;
@@ -147,6 +144,8 @@ export class SeekBar extends Component<SeekBarConfig> {
   private pausedTimeshiftUpdater: Timeout;
 
   private isUserSeeking = false;
+  // We need to track the current ad state manually since our mobile SDKs don't expose an API to check if an ad is active.
+  private isAdPlaying: boolean = false;
 
   private seekBarEvents = {
     /**
--- a/src/ts/components/seekbar/SeekBar.ts
+++ b/src/ts/components/seekbar/SeekBar.ts
@@ -129,7 +129,7 @@ export class SeekBar extends Component<SeekBarConfig> {
   private player: PlayerAPI;
   private uiManager: UIInstanceManager;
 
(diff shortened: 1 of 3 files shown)


stonko1994 (commented):
This fixes the JS exception but it doesn't fix the underlying issue. After fixing the exception there is still a problem that the seekbar doesn't updated after the UI 'hides' (when the non seekbar elements are hidden). See this screen recording from iOS (where it also happens) and note that the seekbar stops updating once it slides down.

https://github.com/user-attachments/assets/d8a523f4-1b95-4957-9522-e3fa84ec8c2a

123mpozzi:
> This fixes the JS exception but it doesn't fix the underlying issue. After fixing the exception there is still a problem that the seekbar doesn't updated after the UI 'hides' (when the non seekbar elements are hidden). See this screen recording from iOS (where it also happens) and note that the seekbar stops updating once it slides down.
> 
>  ScreenRecording_10-09-2025.08-51-41_1.mov

Fix in 8584b3f2c97d5116320f9806d5c4d740a97419b8

src/ts/components/seekbar/SeekBar.ts
```diff
@@ -129,6 +129,9 @@ export class SeekBar extends Component<SeekBarConfig> {
   private player: PlayerAPI;
   private uiManager: UIInstanceManager;
 
+  // Necessary since player.ads?.isLinearAdActive?.() does not exist on mobile SDKs
+  private isAdPlaying: boolean = false;
```
stonko1994: Nit: I'd move this down closer to `isUserSeeking` to 'group' state variables together. Here it's between readonly objects which are not meant to be changed.

src/ts/components/seekbar/SeekBar.ts
```diff
@@ -129,6 +129,9 @@ export class SeekBar extends Component<SeekBarConfig> {
   private player: PlayerAPI;
   private uiManager: UIInstanceManager;
 
+  // Necessary since player.ads?.isLinearAdActive?.() does not exist on mobile SDKs
```
stonko1994: Nit:

```suggestion
  // We need to track the current ad state manually since our mobile SDKs don't expose an API to check if an ad is active.
```

stonko1994 (approved):

src/ts/components/seekbar/SeekBar.ts
123mpozzi: Good point! Address in 8e692e6ca10fd9b231d56b6695186fa602475477

PR #785, opened 2025-10-14
Modern UI - Add state restoration to SettingsPanel

123mpozzi:
## Description

This update improves how auto-hide behaves when the settings panel is open. Previously, the menu could disappear too quickly, which made navigating settings frustrating.

The new logic makes the experience smoother and more predictable.

### What's Changed
- The settings panel now temporarily remembers its state (last page, navigation path, and scroll position) when it's hidden and shown again shortly after
- The remembered state is automatically cleared after a short delay (can be configured) or when the panel is explicitly reset
- Auto-hide behavior has been refined to avoid hiding the UI while the settings panel is open - can be configured
- The default auto-hide delay has been increased to make interaction with settings smoother and more forgiving
- Fix auto-hide timeout being mistakenly applied when on mobile

### UX Considerations

Three possible approaches for handling auto-hide with the settings panel were explored:
1. Keep auto-hide active with the same rules
 - ✅ consistent with the rest of the UI
 - ❌ frustrating if the menu disappears while browsing options
 - ❌ users may feel  "punished" for not moving quickly enough.
2. Disable auto-hide while settings panel is open
- ✅ removes frustration: panel stays until dismissed and user stays in control
- ❌ risk of clutter if the user opens the settings and moves away
3. Hybrid (chosen): longer auto-hide delay while interacting with settings and remember what was opened
- ✅ panel stays open while interacting
- ✅ longer inactivity timeout (5–10× normal) still prevents clutter
- ✅  opened menus can reappear on any interaction, with the same state
- ❌ slightly more complex, but best balance of usability + cleanliness

The hybrid approach offers the best of both worlds: it feels responsive without punishing users for taking their time, and it still keeps the UI tidy

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/UIContainer.ts
+++ b/src/ts/components/UIContainer.ts
@@ -40,10 +40,12 @@ export interface UIContainerConfig extends ContainerConfig {
    */
   hideImmediatelyOnMouseLeave?: boolean;
 
-  // whether the ui containers hide delay should be used, or the hide delay of the settings panel
-  hideDelayTimeoutIfSettingsPanelOpen?: boolean;
-  // detect if a settings panel gets opened, and keep track of it
-  // 
+  /**
+   * When true, suspend the UIContainer's hide timer while a SettingsPanel is open,
+   * and resume it when the panel closes.
+   * Default: true
+   */
+  deferUiHideWhileSettingsOpen?: boolean;
 }
 
 /**
@@ -83,6 +85,7 @@ export class UIContainer extends Container<UIContainerConfig> {
         ariaLabel: i18n.getLocalizer('player'),
         hideDelay: 2000,
         hideImmediatelyOnMouseLeave: true,
+        deferUiHideWhileSettingsOpen: true,
       },
       this.config,
     );
@@ -129,6 +132,24 @@ export class UIContainer extends Container<UIContainerConfig> {
     let isFirstTouch = true;
     let playerState: PlayerUtils.PlayerState;
 
+    if (config.deferUiHideWhileSettingsOpen) {
+      uimanager.onComponentShow.subscribe((component: Component<ComponentConfig>) => {
+        if (component instanceof SettingsPanel) {
+          isSettingsPanelShown = true;
+        }
+      });
+      uimanager.onComponentHide.subscribe((component: Component<ComponentConfig>) => {
+        if (component instanceof SettingsPanel) {
+          isSettingsPanelShown = false;
+
+          if (hideUiPending) {// TODO: find better name for var
+            this.hideUi(true);
+            hideUiPending = false;
+          }
+        }
+      });
+    }
+
     this.hidingPrevented = (): boolean => {
       return config.hidePlayerStateExceptions && config.hidePlayerStateExceptions.indexOf(playerState) > -1;
     };
@@ -147,23 +168,6 @@ export class UIContainer extends Container<UIContainerConfig> {
       }
     };
 
-    uimanager.onComponentShow.subscribe((component: Component<ComponentConfig>) => {
-      if (component instanceof SettingsPanel) {
-        isSettingsPanelShown = true;
-      }
-    });
-
-    uimanager.onComponentHide.subscribe((component: Component<ComponentConfig>) => {
-      if (component instanceof SettingsPanel) {
-        isSettingsPanelShown = false;
-
-        if (hideUiPending) {
-          this.hideUi(true);
-          hideUiPending = false;
-        }
-      }
-    });
-
     this.hideUi = (force: boolean = false) => {
       // Hide the UI only if it is shown, and if not casting
       if (isUiShown && !player.isCasting()) {
@@ -173,7 +177,7 @@ export class UIContainer extends Container<UIContainerConfig> {
           return;
         }
 
-        if (isSettingsPanelShown) {
+        if (config.deferUiHideWhileSettingsOpen && isSettingsPanelShown) {
           hideUiPending = true
           return
         }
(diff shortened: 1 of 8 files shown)


src/ts/components/UIContainer.ts
```diff
@@ -121,11 +133,31 @@ export class UIContainer extends Container<UIContainerConfig> {
...
+        }
+      });
+    }
+
     this.hidingPrevented = (): boolean => {
       return config.hidePlayerStateExceptions && config.hidePlayerStateExceptions.indexOf(playerState) > -1;
     };
 
     this.showUi = () => {
+      hideUiPending = false;
```
felix-hoc: nit: naming

```suggestion
      isHideUiPending = false;
```

src/ts/components/settings/SettingsPanel.ts
```diff
@@ -66,6 +85,11 @@ export class SettingsPanel<Config extends SettingsPanelConfig> extends Container
   // navigation handling
   private activePage: SettingsPanelPage;
   private navigationStack: SettingsPanelPage[] = [];
+  
+  currentState: SettingsPanelState = null;
```
felix-hoc: Please explicitly state the visibility, in this case it seems to be only used internally so it can be `private`


```suggestion
  private currentState: SettingsPanelState = null;
```

src/ts/components/settings/SettingsPanel.ts
```diff
@@ -341,6 +402,81 @@ export class SettingsPanel<Config extends SettingsPanelConfig> extends Container
...
+  }
+
+  private resetState(): void {
+    this.activePage = this.getRootPage();
+    this.navigationStack = [];
+    this.currentState = null;
+    this.resetStateTimerId = null;
+
+    if (this.isHidden()) {
+      this.getDomElement().css({ width: '', height: '' });
```
felix-hoc: why do we need to set the width and height? shouldn't the hiding be sufficient?

src/ts/components/settings/SettingsPanel.ts
```diff
@@ -28,6 +28,25 @@ export interface SettingsPanelConfig extends ContainerConfig {
    * Default: true
    */
   pageTransitionAnimation?: boolean;
+
+  /**
+   * The delay in milliseconds after hiding the settings panel before its internal state
+   * (e.g., navigation stack and scroll position) is reset.
+   * Default: 5 seconds (5000)
+   */
+  stateResetDelay?: number;
```
felix-hoc, line 38: Is there also a way to disable it - like for `hideDelay`? Do we want that?

felix-hoc (commented):
Nice, generally LGTM and works well 🙌🏼  had a few comments, happy to discuss :)

src/ts/components/settings/SettingsPanel.ts
123mpozzi: No, there was no way. But makes a lot of sense to have it!
I've added it in dad1575047b6f111b9b5a359c50ac3a3fd39faf1 🙂  (used the same mechanism we have for `hideDelay`)

src/ts/components/settings/SettingsPanel.ts
123mpozzi: It's because different settings panel pages have different dimensions.

For example, the videoQuality page in the sample page pushes the panel to full height because it has many items.
If we hide and then restore it without resetting width/height, the root menu would appear with an oversized panel.

image

---

Here however this `isHidden` check was to immediately reset the component dimensions to auto since it was already hidden (and hence not visually noticeable for the user) - but I see that's tricky way of dealing with this.

I placed the dimension handling for both _shown_ (via `restoreNavigationState`) and _hidden_ components in the same place in f5db4269870f606847a561b4f64938a49b5b82e7

src/ts/components/settings/SettingsPanel.ts
123mpozzi: Ah good catch!

src/ts/components/UIContainer.ts
123mpozzi: Good idea - address in 493f095c8e80f52e46f713c359d9efbc3ea0219d

felix-hoc:
I've noticed two issues:
1. If the player dimensions are small, every time a panel is restored, it gets narrower, until it reaches the minimum width. Maybe a rounding error?
2. There is a width scaling animation getting played after the `stateResetDelay` times out - if the previously stored width was narrower that the main settings menu, it will get a "scale up" animation. It's not that big of a deal, just a bit unexpected. Maybe some state is not reset properly?


https://github.com/user-attachments/assets/6237cb08-fc15-4ec9-937d-7fde5aa1ccd4

felix-hoc (changes requested):
(see comment 👆🏼 )

123mpozzi:
> I've noticed two issues:
> 
> 1. If the player dimensions are small, every time a panel is restored, it gets narrower, until it reaches the minimum width. Maybe a rounding error?
> 2. There is a width scaling animation getting played after the `stateResetDelay` times out - if the previously stored width was narrower that the main settings menu, it will get a "scale up" animation. It's not that big of a deal, just a bit unexpected. Maybe some state is not reset properly?
> 
>  Screen.Recording.2025-10-20.at.07.50.15.mov

Thanks for reporting and for the detailed video 🙌 

I could not reproduce locally unfortunately: in my case all the settings panel pages have the same width even with a smaller player 🙁 

But I think I have fixed the issues:
1.  Address in bbe000d4a7ed4a603785eeaeb96a3a3dbf79fd94 by letting CSS handle the panel dimension calculations itself
2. Address in 161949dfcad683067b2ec9f97e9baaa19fc6f080 by resetting the dimensions when the panel is not visible instead of on show

felix-hoc (approved):
nice, LGTM :)

PR #788, opened 2025-11-05
Enable prefer-const linter rule

felix-hoc:
## Description

We are using `let` for non-reassigned variables in a bunch of places. The linter rule to prefer `const` is disabled, which allows new usages to slip into the code base.

=> Enabling the linter rule and ran the eslint autofixer to replace `let` with `const` wherever possible.

- / `CHANGELOG` entry - no, just a chore


stonko1994 (approved):
thx 🙌

PR #792, opened 2025-11-20
remove aria-label from Icon

oberhamsi:
## Description
see internal ticket #28261

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/Icon.ts
+++ b/src/ts/components/Icon.ts
@@ -1,35 +1,22 @@
 import { Component, ComponentConfig } from './Component';
-import { i18n, LocalizableText } from '../localization/i18n';
 import { DOM } from '../DOM';
 
-export interface IconConfig extends ComponentConfig {
-  /**
-   * WCAG20 standard for defining info about the component (usually the name)
-   */
-  ariaLabel?: LocalizableText;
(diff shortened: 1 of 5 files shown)


src/ts/components/Icon.ts
```diff
@@ -1,35 +1,22 @@
...
-  /**
-   * Alternative text for the icon, used for accessibility (e.g., screen readers).
-   */
-  altText?: LocalizableText;
-}
-
-export class Icon extends Component<IconConfig> {
-  constructor(config: IconConfig) {
+export class Icon extends Component<ComponentConfig> {
+  constructor(config: ComponentConfig) {
```
stonko1994: Nit: I was thinking if we should set a default value here to not need the empty object on the caller side 🤔 


```suggestion
  constructor(config: ComponentConfig = {}) {
```

stonko1994 (approved):
Thank you for your contribution 🙌 Really appreciate it.

Added an optional comment. Feel free to address or ignore it.

src/ts/components/Icon.ts
oberhamsi: makes sense, i'll clean it up!

PR #796, opened 2025-11-21
don't stop propagation unless using component wants it

oberhamsi:
afaiu this change is only relevant for key-handling in [InteractiveSettingsPanelItem](https://github.com/bitmovin/bitmovin-player-ui/blob/75d84c8010f4430df49db0b9f5625082e0c13475/src/ts/components/settings/InteractiveSettingsPanelItem.ts#L43)

and i think this still works with this PR? if you can give me a case where it doesn't work, i will continue working on the PR. maybe look into blur/focus or see if we can pass the new stopPropagationFocus down to where it's needed.

## Description
stopping propagation on click and focusin/focusout should only happen if the parent component wants it.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Removed
+
+- Button no longer calls stopPropagation on its click, focusin and focusout Events
+
 ## [4.4.0] - 2025-11-27
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


oberhamsi:
hey - still waiting for feedback on this, thanks

stonko1994:
Hey @oberhamsi, thanks for opening the PR. It's on my list to look at. I hope to get to it this week.

My initial though is that we can remove the `stopPropagation` completely again as we no longer have a use-case where we need this. But I still need to test a few use-cases to verify that.

oberhamsi:
alright! Note how i also changed it for the focus events where i am less sure it still works as intended

stonko1994:
Hi @oberhamsi I tested it again and IMO we can remove the `stopPropagation` completely again. (For click and focus). No need to add the config.

Please add a changelog entry stating that this is removed (In a `### Removed` section)

stonko1994 (commented):
See https://github.com/bitmovin/bitmovin-player-ui/pull/796#issuecomment-3595484408

oberhamsi:
thanks! I will do your suggestions later this week

stonko1994 (approved):
Thank you

PR #808, opened 2025-12-16
Add option to use Shadow DOM rendering of the UI

stonko1994:
## Description

In v4, the CSS selectors changed, resulting in a problem where specificity is not as strong as before. Therefore, we wanted to add an option to render our UI inside a Shadow DOM to eliminate potential CSS conflicts with an enclosing page.

### Changes
To enable that, this PR adds a ShadowDomConfig to the `UIConfig`, which can be used to enable and configure the ShadowDOM rendering.

### Manual Testing
You can add some conflicting CSS styles to the index.html and check the result w/ and w/o Shadow DOM rendering enabled:
```
<style>
  .container div > span {
    font-size: 1rem;
    line-height: 1.6;
    margin-bottom: 1rem;
    margin-top: 0.5rem;
    color: rgb(80 95 121);
  }
</style>
```

- [x] `CHANGELOG` entry

### DIFF
--- a/spec/utils/ShadowDomManager.spec.ts
+++ b/spec/utils/ShadowDomManager.spec.ts
@@ -1,0 +1,93 @@
+import { ShadowDomManager } from '../../src/ts/utils/ShadowDomManager';
+import { ShadowDomConfig } from '../../src/ts/UIConfig';
+import { DOM } from '../../src/ts/DOM';
+
+describe('ShadowDomManager', () => {
+  const appendLink = (href: string) => {
+    const link = document.createElement('link');
+    link.rel = 'stylesheet';
+    link.href = href;
+    document.head.appendChild(link);
+    return link;
+  };
+
+  beforeEach(() => {
+    document.head.innerHTML = '';
+    document.body.innerHTML = '';
+  });
+
+  describe('isShadowDomSupported', () => {
+    it('returns true when attachShadow is available', () => {
+      expect(ShadowDomManager.isShadowDomSupported()).toBe(true);
+    });
+
+    it('returns false and warns when Shadow DOM APIs are missing', () => {
+      const originalAttachShadow = (HTMLElement.prototype as any).attachShadow;
+      const originalShadowRoot = (global as any).ShadowRoot;
+      delete (HTMLElement.prototype as any).attachShadow;
+      (global as any).ShadowRoot = undefined;
+
+      expect(ShadowDomManager.isShadowDomSupported()).toBe(false);
+
+      if (originalAttachShadow) {
+        (HTMLElement.prototype as any).attachShadow = originalAttachShadow;
+      }
+      (global as any).ShadowRoot = originalShadowRoot;
+    });
+  });
+
+  describe('initialize', () => {
+    let manager: ShadowDomManager;
+    let container: DOM;
+
+    beforeEach(() => {
+      manager = new ShadowDomManager();
+      container = new DOM('div', {});
+      document.body.appendChild(container.get(0));
+    });
+
+    it('attaches a shadow root and injects main stylesheet by filename', () => {
+      appendLink('https://cdn.example.com/bitmovinplayer-ui.css');
+      const config: ShadowDomConfig = { enabled: true, uiStylesheetName: 'bitmovinplayer-ui' };
+
+      manager.initialize(container, config);
+
+      const shadowRoot = manager.getShadowRoot();
+      expect(shadowRoot).toBeDefined();
+      const links = shadowRoot.querySelectorAll('link[rel="stylesheet"]');
+      expect(links.length).toBe(1);
+      expect((links[0] as HTMLLinkElement).href).toContain('bitmovinplayer-ui');
+    });
+
+    it('injects additional stylesheets', () => {
+      appendLink('https://cdn.example.com/bitmovinplayer-ui.css');
+      const config: ShadowDomConfig = {
+        enabled: true,
+        uiStylesheetName: 'bitmovinplayer-ui',
+        additionalStylesheets: ['https://cdn.example.com/custom.css', 'other.css'],
+      };
+      appendLink('https://example.com/other.css');
+
+      manager.initialize(container, config);
+
+      const links = manager.getShadowRoot().querySelectorAll('link[rel="stylesheet"]');
+      const hrefs = Array.from(links).map(link => (link as HTMLLinkElement).href);
+      expect(hrefs).toEqual(
+        expect.arrayContaining([
+          expect.stringContaining('bitmovinplayer-ui.css'),
+          'https://cdn.example.com/custom.css',
+          expect.stringContaining('other.css'),
+        ]),
+      );
+    });
+
+    it('releases host and shadow root', () => {
+      appendLink('https://cdn.example.com/bitmovinplayer-ui.css');
+      manager.initialize(container, { enabled: true, uiStylesheetName: 'bitmovinplayer-ui' });
+
+      manager.release();
+
+      expect(manager.getShadowRoot()).toBeUndefined();
+    });
+  });
+});
--- a/spec/utils/ShadowDomManager.spec.ts
+++ b/spec/utils/ShadowDomManager.spec.ts
@@ -46,7 +46,7 @@ describe('ShadowDomManager', () => {
 
     it('attaches a shadow root and injects main stylesheet by filename', () => {
       appendLink('https://cdn.example.com/bitmovinplayer-ui.css');
-      const config: ShadowDomConfig = { enabled: true, uiStylesheetName: 'bitmovinplayer-ui' };
+      const config: ShadowDomConfig = { enabled: true, uiStylesheet: 'bitmovinplayer-ui' };
 
       manager.initialize(container, config);
 
@@ -61,7 +61,7 @@ describe('ShadowDomManager', () => {
       appendLink('https://cdn.example.com/bitmovinplayer-ui.css');
       const config: ShadowDomConfig = {
         enabled: true,
-        uiStylesheetName: 'bitmovinplayer-ui',
+        uiStylesheet: 'bitmovinplayer-ui',
         additionalStylesheets: ['https://cdn.example.com/custom.css', 'other.css'],
       };
       appendLink('https://example.com/other.css');
@@ -81,7 +81,7 @@ describe('ShadowDomManager', () => {
 
     it('releases host and shadow root', () => {
       appendLink('https://cdn.example.com/bitmovinplayer-ui.css');
-      manager.initialize(container, { enabled: true, uiStylesheetName: 'bitmovinplayer-ui' });
+      manager.initialize(container, { enabled: true, uiStylesheet: 'bitmovinplayer-ui' });
 
       manager.release();
 
--- a/spec/utils/ShadowDomManager.spec.ts
+++ b/spec/utils/ShadowDomManager.spec.ts
@@ -29,9 +29,7 @@ describe('ShadowDomManager', () => {
 
       expect(ShadowDomManager.isShadowDomSupported()).toBe(false);
 
-      if (originalAttachShadow) {
-        (HTMLElement.prototype as any).attachShadow = originalAttachShadow;
-      }
+      (HTMLElement.prototype as any).attachShadow = originalAttachShadow;
       (global as any).ShadowRoot = originalShadowRoot;
     });
   });
--- a/src/ts/UIConfig.ts
+++ b/src/ts/UIConfig.ts
@@ -172,11 +172,12 @@ export interface UIConfig {
   /**
    * Configure Shadow DOM rendering.
    * Enable it with:
-   * `shadowDomConfig: { enabled: true }`
+   * `shadowDom: true`
+   * `shadowDom: { enabled: true }` (especially if you need to set more configuration options)
    *
-   * Default: undefined (Shadow DOM disabled).
+   * Default: false
    */
-  shadowDomConfig?: ShadowDomConfig;
+  shadowDom?: ShadowDomConfig | boolean;
 }
 
 export interface ShadowDomConfig {
--- a/src/ts/UIConfig.ts
+++ b/src/ts/UIConfig.ts
@@ -175,6 +175,9 @@ export interface UIConfig {
    * `shadowDom: true`
    * `shadowDom: { enabled: true }` (especially if you need to set more configuration options)
    *
+   * Enable ShadowDom rendering to prevent CSS from an enclosing website to interfere with the UI styles. Check
+   * the availability here: https://caniuse.com/shadowdomv1.
+   *
    * Default: false
    */
   shadowDom?: ShadowDomConfig | boolean;
--- a/src/ts/UIConfig.ts
+++ b/src/ts/UIConfig.ts
@@ -193,14 +193,16 @@ export interface ShadowDomConfig {
    */
   enabled: boolean;
   /**
-   * Filename of the UI stylesheet to inject into the Shadow DOM.
+   * Filename or URL of the UI stylesheet to inject into the Shadow DOM.
    *
-   * When the Shadow DOM is created, the UI looks for a linked stylesheet whose `href` contains this
+   * If a non-URL value is provided, the UI looks for a linked stylesheet whose `href` contains this
    * value and reuses it inside the shadow root so the default styling is available.
+   *
+   * If a URL is provided, the value is used as it is.
    *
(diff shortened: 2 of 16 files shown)


src/ts/UIConfig.ts
```diff
@@ -169,4 +169,41 @@ export interface UIConfig {
...
+   * Configure Shadow DOM rendering.
+   * Enable it with:
+   * `shadowDomConfig: { enabled: true }`
+   *
+   * Default: undefined (Shadow DOM disabled).
+   */
+  shadowDomConfig?: ShadowDomConfig;
+}
+
+export interface ShadowDomConfig {
```
stonko1994, line 186: I was undecided with introducing a new `ShadowDomConfig` or just adding those options to the `UIConfig`. I went with the extra Config to group them together.

Also, I wasn't sure what our naming is on the web for nested configs. I saw that on the Player we don't suffix them with `Config`. Let me know if I should remove the Config suffix. I was also considering changing the type to `shadowDom?: boolean | ShadowDomConfig` to enable a short-hand syntax of just `shadowDom: true`. Let me know if this is something we want to have.

src/ts/UIConfig.ts
```diff
@@ -169,4 +169,41 @@ export interface UIConfig {
...
    */
   includeWatermark?: boolean;
+  /**
+   * Configure Shadow DOM rendering.
+   * Enable it with:
+   * `shadowDomConfig: { enabled: true }`
+   *
+   * Default: undefined (Shadow DOM disabled).
+   */
+  shadowDomConfig?: ShadowDomConfig;
```
dweinber:
```suggestion
  shadowDom?: ShadowDomConfig | boolean;
```

See reasoning below

src/ts/UIConfig.ts
dweinber: I think the grouping into `ShadowDomConfig` makes sense. 

Regarding the naming: For the player, we typically suffix the _types_ with `Config`, but not the properties. See https://cdn.bitmovin.com/player/web/8/docs/interfaces/Core.PlayerConfig.html#playback for a few examples.

And yes, the shorthand boolean would be nice 🙂

src/ts/UIConfig.ts
```diff
@@ -169,4 +169,41 @@ export interface UIConfig {
    * Default: false
    */
   includeWatermark?: boolean;
+  /**
+   * Configure Shadow DOM rendering.
+   * Enable it with:
+   * `shadowDomConfig: { enabled: true }`
+   *
+   * Default: undefined (Shadow DOM disabled).
```
dweinber: Why do we want to have this disabled by default? 🤔

src/ts/UIConfig.ts
```diff
@@ -169,4 +169,41 @@ export interface UIConfig {
    * Default: false
    */
   includeWatermark?: boolean;
+  /**
+   * Configure Shadow DOM rendering.
```
dweinber, line 173: Is it worth adding any comment regarding browser/platform support? https://caniuse.com/shadowdomv1 shows pretty good support, except for older TVs/STBs/etc

src/ts/UIConfig.ts
```diff
@@ -169,4 +169,41 @@ export interface UIConfig {
...
+   * Enable it with:
+   * `shadowDomConfig: { enabled: true }`
```
dweinber: If you change the shorthand to be supported, let's show both here?


```suggestion
   * `shadowDom: true`
   * `shadowDom: { enabled: true }` (especially if you need to set more configuration options)
```

src/ts/utils/FocusVisibilityTracker.ts
```diff
@@ -45,13 +52,13 @@ export class FocusVisibilityTracker {
 
   private registerEventListeners(): void {
     for (const event in this.eventHandlerMap) {
-      document.addEventListener(event, this.eventHandlerMap[event], true);
+      this.uiWrapperElement.on(event, this.eventHandlerMap[event], true);
```
dweinber, line 55: In the case of no shadow DOM, there is now a change from `document.addEventListener` to `this.uiWrapperElement.on` 🤔 
Is this intentional? Why? Can there be any problems with this change? Would be good to get insights from someone who was part of the initial implementation of the `FocusVisibilityTracker` to understand why `document` was used in the first place (thinking of some werid TV issues or whatnot)

src/ts/components/settings/SelectBox.ts
```diff
@@ -176,11 +178,11 @@ export class SelectBox extends ListSelector<ListSelectorConfig> {
 
     clearTimeout(this.dropdownCloseListenerTimeoutId);
 
-    DocumentDropdownClosedEvents.forEach(event => document.addEventListener(event, this.onDropdownClosed, true));
+    DocumentDropdownClosedEvents.forEach(event => this.uiWrapperElement.on(event, this.onDropdownClosed, true));
```
dweinber, line 181: In the case of no Shadow DOM, this is changed from `document.addEventListener` to just the UI's/player's parent `div`. Is this intentional? Do you know why we listened for the whole `document` previously and whether that's an issue we're (re-)introducing?

src/ts/UIManager.ts
```diff
@@ -249,6 +252,10 @@ export class UIManager {
       this.uiContainerElement = new DOM(player.getContainer());
     }
 
+    if (this.config.shadowDomConfig && this.config.shadowDomConfig.enabled && ShadowDomManager.isShadowDomSupported()) {
```
dweinber: Enable the shorthand option as discussed in some comment

```suggestion
    if ((this.config.shadowDom === true || this.config.shadowDom.enabled) && ShadowDomManager.isShadowDomSupported()) {
```

src/ts/utils/ShadowDomManager.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  public static isShadowDomSupported(): boolean {
+    const shadowSupported =
+      typeof ShadowRoot !== 'undefined' &&
+      typeof HTMLElement !== 'undefined' &&
+      typeof HTMLElement.prototype.attachShadow === 'function';
+
+    if (!shadowSupported) {
+      console.warn('Shadow DOM is not supported in this environment. Falling back to classic UI rendering.');
+    }
```
dweinber: NIT: IMO this warning should not belong in this class, the caller should handle this (it's good to have it!). This class (and even less so its static method) doesn't decide whether the UI rendering falls back to the classic option, or fails, or does something else.

src/ts/utils/ShadowDomManager.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    return shadowSupported;
+  }
+
+  initialize(containerElement: DOM, shadowDomConfig: ShadowDomConfig) {
```
dweinber, line 26:
```suggestion
  initialize(containerElement: DOM, shadowDom: ShadowDomConfig) {
```

src/ts/utils/ShadowDomManager.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    const stylesheetName = shadowDomConfig.uiStylesheetName || ShadowDomManager.DEFAULT_SHADOW_STYLESHEET_NAME;
+    const mainHref = resolveStylesheetHref(stylesheetName);
+
+    if (mainHref) {
+      appendStylesheet(this.shadowRoot, mainHref);
+    } else {
+      console.warn(
+        'Shadow DOM is enabled but no stylesheet was found. ' +
+          'Provide `shadowDomCss` (URL or filename) or include bitmovinplayer-ui.css on the page.',
```
dweinber: Shouldn't we use the `{{FILENAME}}` here, too?
And `shadowDomCss` is not a valid config property.

```suggestion
          'Provide `uiStylesheetName` (URL or filename) or include {{FILENAME}} on the page.',
```

spec/utils/ShadowDomManager.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    it('returns false and warns when Shadow DOM APIs are missing', () => {
+      const originalAttachShadow = (HTMLElement.prototype as any).attachShadow;
+      const originalShadowRoot = (global as any).ShadowRoot;
+      delete (HTMLElement.prototype as any).attachShadow;
+      (global as any).ShadowRoot = undefined;
+
+      expect(ShadowDomManager.isShadowDomSupported()).toBe(false);
+
+      if (originalAttachShadow) {
```
dweinber: Why is this check needed for `attachShadow` but not for the `ShadowRoot`?

src/ts/UIConfig.ts
```diff
@@ -169,4 +169,41 @@ export interface UIConfig {
...
+  enabled: boolean;
+  /**
+   * Filename of the UI stylesheet to inject into the Shadow DOM.
+   *
+   * When the Shadow DOM is created, the UI looks for a linked stylesheet whose `href` contains this
+   * value and reuses it inside the shadow root so the default styling is available.
+   *
+   * Default: 'bitmovinplayer-ui'
+   */
+  uiStylesheetName?: string;
```
dweinber: Can a full URL be passed as well? e.g. not linking it in the document at all and pass the URL here, so that it's _only_ injected into the `ShadowRoot`?

dweinber (changes requested):
Nice improvement, a few comments and discussion points below

src/ts/UIConfig.ts
stonko1994: To not break existing integrations. It would break additional CSS stylings.

src/ts/UIConfig.ts
stonko1994: Hm what makes this special to other APIs where we also don't add this information? 🤔 Also how can I find out older TVs support? 🤔 

I added a note and link to it: [e6e1d29](https://github.com/bitmovin/bitmovin-player-ui/pull/808/commits/e6e1d29230f7e2f45e9a393950a30c2ca2817392)

src/ts/UIConfig.ts
stonko1994: [90750c1](https://github.com/bitmovin/bitmovin-player-ui/pull/808/commits/90750c15b99a52cbf477c6e5f37ba21fa9834804)

src/ts/UIConfig.ts
stonko1994: [90750c1](https://github.com/bitmovin/bitmovin-player-ui/pull/808/commits/90750c15b99a52cbf477c6e5f37ba21fa9834804)

src/ts/UIManager.ts
stonko1994: [90750c1](https://github.com/bitmovin/bitmovin-player-ui/pull/808/commits/90750c15b99a52cbf477c6e5f37ba21fa9834804)

src/ts/utils/ShadowDomManager.ts
stonko1994: [90750c1](https://github.com/bitmovin/bitmovin-player-ui/pull/808/commits/90750c15b99a52cbf477c6e5f37ba21fa9834804)

src/ts/utils/FocusVisibilityTracker.ts
stonko1994: This is an intentional change because IMO having it on the document was a 'shortcut' in first place. (I'm familiar with the implementation).

There is this in line 40:
```
isBitmovinUi(element, this.bitmovinUiPrefix)
```

So we are anyway just interested in focused elements within our UI to add the correct focus css class for our styling.

src/ts/utils/ShadowDomManager.ts
```diff
@@ -0,0 +1,124 @@
+import { DOM } from '../DOM';
+import { ShadowDomConfig } from '../UIConfig';
+import { prefixCss } from '../components/DummyComponent';
+
+/**
+ * Encapsulates Shadow DOM initialization.
+ */
+export class ShadowDomManager {
```
dweinber, line 8: Is there any reason to export (or to not export) the new `ShadowDomManager`  from `main.ts`, so that it can be used/accessed like other components?

src/ts/utils/FocusVisibilityTracker.ts
dweinber: Makes sense 👍 Thanks for double-checking and the explanation

src/ts/components/settings/SelectBox.ts
dweinber: Seems to be fine, too

src/ts/UIConfig.ts
dweinber: Fair question on what's different in this case. 

My reasoning was that typically whatever we use is simply being used by the UI, and we need to make sure it's supported on all platforms we support, or it's properly transpiled to something that works everywhere, or we have polyfills (in the UI or the player) to make up for it.

In this case though, it's a decision the customer needs to make, so I thought it might be good to give them all the information needed to make a good decision. However, as there's anyway an automatic fallback, it may not be as necessary on a second thought. So I'm fine either way.

Regarding older TV support:

The safest option is typically simply testing, but especially with more exotic STBs and similar devices this is not feasible. 

Samsung Tizen 2016 TVs (the oldest we support) use `Webkit r152340` which is a bit tricky to map to versions in caniuse; 2017 uses Chromium 47: https://developer.samsung.com/smarttv/develop/specifications/web-engine-specifications.html

LG WebOS 2016 TVs (again the oldest we support) use Chromium 38: https://webostv.developer.lge.com/develop/specifications/web-api-and-web-engine

src/ts/utils/ShadowDomManager.ts
stonko1994: Agree: [ceb4c9a](https://github.com/bitmovin/bitmovin-player-ui/pull/808/commits/ceb4c9a5144437fb4e6ff9705ca5c45f9692af93)

src/ts/utils/ShadowDomManager.ts
stonko1994: [20fd717](https://github.com/bitmovin/bitmovin-player-ui/pull/808/commits/20fd7178f50221e475bdf626b557875cac6869b5)

spec/utils/ShadowDomManager.spec.ts
stonko1994: [de815de](https://github.com/bitmovin/bitmovin-player-ui/pull/808/commits/de815deac8953f006f8a551d643e51bc35570ea4)

src/ts/UIConfig.ts
stonko1994: Yes that is supported. I renamed the config and updated the documentation: [f8b0637](https://github.com/bitmovin/bitmovin-player-ui/pull/808/commits/f8b063714dca0a85b7bb83d2dd769eae7333369c)

src/ts/utils/ShadowDomManager.ts
stonko1994: I considered this as an internal helper class and not part of the public API of the framework. Do you think there will be a use-case to use from the outside?

dweinber: I couldn't think of one really, hence my question 😄 
Maybe the `isShadowDomSupported()` to check upfront if it's supported and will be used?

src/ts/utils/ShadowDomManager.ts
```diff
@@ -0,0 +1,119 @@
...
+              'Check the URL or ensure the stylesheet is linked on the page.',
+          );
+        }
+      });
+    }
+  }
+}
+
+function resolveStylesheetHref(stylesheetName: string): string | null {
+  const isUrl = /^https?:\/\//.test(stylesheetName) || stylesheetName.startsWith('/');
```
dweinber: Hm this actually means we don't support relative URLs. Is this intentional/considered?

stonko1994: [295f476](https://github.com/bitmovin/bitmovin-player-ui/pull/808/commits/295f476b01d8d0c91908dcbcdf8d9a6f1bd79836)

src/ts/utils/ShadowDomManager.ts
stonko1994: I guess that could be use-full: [455e554](https://github.com/bitmovin/bitmovin-player-ui/pull/808/commits/455e554488b8f1cb3556e5f7be99ce95c6bc3723)

src/ts/utils/ShadowDomManager.ts
dweinber: Nice, thanks

dweinber (approved):
Nice feature 👍 Looks good now!

PR #811, opened 2026-01-02
Fix uncaught errors with live streams after player destruction 

jmsn:
### Issue

The `TimelineMarkersHandler` and its polling logic for live streams are only reset based on the player's `sourceunloaded` event.

Destroying the player would cause uncaught JS errors, as the UI's polling logic would operate on a destroyed player.

### Fix

Add resetting `TimelineMarkersHandler` on player destroy

- [x] `CHANGELOG` entry


zenzi-deluxe (approved):
👍

stonko1994 (approved):

PR #814, opened 2026-01-13
fix(slider): enhanced UX increasing hit area sizes

francoismassart:
## Description
The limited `height` of the **progress bar** and of the **volume** slider makes it difficult to to be clicked (even using a mouse).
The same goes for the knobs on both of these sliders.

This MR uses some CSS magic to inject a bigger hit area via `:after` pseudo selector.
It looks the same,  but it is now **way easier** to use these sliders even on touch screen devices 🎉.

### Before
before

### After
after

NB: the semi-transparent `red` and `lime` colors are not rendered by default 😇, for debugging purposes only.

- [x] `CHANGELOG` entry


stonko1994 (approved):

stonko1994:
Thank you very much for this contribution. Really appreciate it.

PR #816, opened 2026-02-05
Update labels on duration changed (Seekbar + PlabackTime)

testcenter:
## Description
When casting VOD streams >1h the duration will be displayed without hours.

This is because the duration is not yet available with the initial "ready" event (`duration=0.0`).
As the duration gets available with a following cast event, the time format for the `SeekBarLabel` and `PlaybackTimeLabel` are already fixed and will not change.

With this change we are listening to `onDurationChanged` events in the `SeekBarLabel` and `PlaybackTimeLabel` and will update the time format accordingly

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-- Time format will be updated when an `onDurationChanged` event is received
+- Time-related labels not properly updating the time format if the duration changes from below 1 hour to greater than 1 hour.
 
 ## [4.8.0] - 2026-01-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- Time format will be updated when an `onDurationChanged` event is received
+
 ## [4.8.0] - 2026-01-29
 
 ### Added
--- a/src/ts/components/labels/PlaybackTimeLabel.ts
+++ b/src/ts/components/labels/PlaybackTimeLabel.ts
@@ -173,6 +173,7 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
       updateTimeFormatBasedOnDuration();
     };
     uimanager.getConfig().events.onUpdated.subscribe(init);
+    player.on(player.exports.PlayerEvent.DurationChanged, init);
 
     init();
   }
--- a/src/ts/components/labels/PlaybackTimeLabel.ts
+++ b/src/ts/components/labels/PlaybackTimeLabel.ts
@@ -151,17 +151,6 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
     });
     liveStreamDetector.detect(); // Initial detection
 
-    player.on(player.exports.PlayerEvent.TimeChanged, playbackTimeHandler);
-    player.on(player.exports.PlayerEvent.Ready, updateTimeFormatBasedOnDuration);
-    player.on(player.exports.PlayerEvent.Seeked, playbackTimeHandler);
-
-    player.on(player.exports.PlayerEvent.TimeShift, updateLiveTimeshiftState);
-    player.on(player.exports.PlayerEvent.TimeShifted, updateLiveTimeshiftState);
-    player.on(player.exports.PlayerEvent.Playing, updateLiveTimeshiftState);
-    player.on(player.exports.PlayerEvent.Paused, updateLiveTimeshiftState);
-    player.on(player.exports.PlayerEvent.StallStarted, updateLiveTimeshiftState);
-    player.on(player.exports.PlayerEvent.StallEnded, updateLiveTimeshiftState);
-
     const init = () => {
       // Reset min-width when a new source is ready (especially for switching VOD/Live modes where the label content
       // changes)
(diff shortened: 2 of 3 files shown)


src/ts/components/labels/PlaybackTimeLabel.ts
```diff
@@ -173,6 +173,7 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
       updateTimeFormatBasedOnDuration();
     };
     uimanager.getConfig().events.onUpdated.subscribe(init);
+    player.on(player.exports.PlayerEvent.DurationChanged, init);
```
stonko1994: Is there a reason that we need to call `init`? Why not using the `updateTimeFormatBasedOnDuration` similar to the `Ready` event?

stonko1994: Can we group the player event listening to others on top?

src/ts/components/seekbar/SeekBarLabel.ts
```diff
@@ -61,26 +61,31 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
...
-          : StringUtils.FORMAT_MMSS;
-      // Set initial state of title and thumbnail to handle sourceLoaded when switching to a live-stream
-      this.setTitleText(null);
-      this.setThumbnail(null);
-    };
-
-    uimanager.getConfig().events.onUpdated.subscribe(init);
-    init();
+    uimanager.getConfig().events.onUpdated.subscribe(this.init);
+    player.on(PlayerEvent.DurationChanged, this.init);
```
stonko1994: We must not use `PlayerEvent` directly in the UI as it will bundle the whole player into the UI which results in a duplicated player. (Known TS issue)

Instead we need to use `player.exports.PlayerEvent` as in the other component.

src/ts/components/seekbar/SeekBarLabel.ts
```diff
@@ -61,26 +61,31 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
     );
   }
 
+  private init = () => {
```
stonko1994: No need to keep that as an arrow function if you move it here: + naming

```suggestion
  private initializeTimeFormat(): void {
```

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- Time format will be updated when an `onDurationChanged` event is received
```
stonko1994: This sounds more like an added/changed entry and not like something which needed fixing.

I'd propose to either add this as changed:

```
### Changed

- Time format of duration-related labels will be updated when an `onDurationChanged` event is received
```

or if you want to indicate the fix:

```
### Fixed

- Time-related labels not properly updating the time format if the duration changes from below 1 hour to greater than 1 hour.
```

src/ts/components/seekbar/SeekBarLabel.ts
```diff
@@ -251,5 +256,6 @@ export class SeekBarLabel extends Container<SeekBarLabelConfig> {
     super.release();
 
     this.uiManager.onSeekPreview.unsubscribe(this.handleSeekPreview);
+    this.player.off(PlayerEvent.DurationChanged, this.init);
```
stonko1994: Same as above. Please use `player.exports`

stonko1994 (changes requested):

src/ts/components/labels/PlaybackTimeLabel.ts
testcenter: As it resets the css style. Otherwise (when only calling `updateTimeFormatBasedOnDuration`) the css is not updated and so some parts of the label are truncated (or overdrawn) by the seekbar.

src/ts/components/seekbar/SeekBarLabel.ts
testcenter: Unfortunately eslint doesn't allow this.
(`Avoid referencing unbound methods which may cause unintentional scoping of 'this'.`)

src/ts/components/seekbar/SeekBarLabel.ts
testcenter: https://github.com/bitmovin/bitmovin-player-ui/pull/816/commits/a928ab0fbf73b7afeb2ccfe0aef579915b01d0f3

src/ts/components/seekbar/SeekBarLabel.ts
testcenter: Thanks for letting me know! :bow: 
Fixed with https://github.com/bitmovin/bitmovin-player-ui/pull/816/commits/a928ab0fbf73b7afeb2ccfe0aef579915b01d0f3

src/ts/components/labels/PlaybackTimeLabel.ts
testcenter: Restructured with https://github.com/bitmovin/bitmovin-player-ui/pull/816/commits/79654f8bbe01d890a134b159af4dddd09a9b8ffe
(had to move the `init` block as well)

CHANGELOG.md
testcenter: I kept it as `Fix`.
Thanks for finding a better wording :+1: 
https://github.com/bitmovin/bitmovin-player-ui/pull/816/commits/5173071a799e1fe3d7bf5026bd25fe6bd1e5cedf

src/ts/components/seekbar/SeekBarLabel.ts
stonko1994: ok. can we still do the renaming please. Having a function just called `init` might be confusing

stonko1994 (commented):
https://github.com/bitmovin/bitmovin-player-ui/pull/816#discussion_r2768337473

src/ts/components/seekbar/SeekBarLabel.ts
testcenter: renamed the function: https://github.com/bitmovin/bitmovin-player-ui/pull/816/commits/7386d5212c3875b73fac8d99643523ae61140381

stonko1994 (approved):

PR #817, opened 2026-02-05
Prevent subtitle/audio track selection feedback loop

stonko1994:
## Description

We have a feedback loop in audio and subtitle track selection that is triggered by the Player without any user interaction.

The UI needs to assume that Player events are asynchronous. Currently, the UI re-enables the subtitle track via a SubtitleEnabled event, which may override another API call to the player. This was particularly an issue on iOS around the initial subtitle selection. We encountered an issue where the UI overrides the native subtitle selection, which was handled in the `SourceLoaded` (which is synchronous) event listener.

### Changes
To prevent this feedback loop, this PR introduced dedicated methods to distinguish player updates from user selections. Therefore, the UI won't set the subtitle track again after the player emits its event.

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/settings/PlaybackSpeedSelectBox.ts
+++ b/src/ts/components/settings/PlaybackSpeedSelectBox.ts
@@ -30,9 +30,8 @@ export class PlaybackSpeedSelectBox extends SelectBox {
 
     this.addDefaultItems();
 
-    this.onItemSelected.subscribe((sender: PlaybackSpeedSelectBox, value: string) => {
+    this.onItemSelectionChanged.subscribe((sender: PlaybackSpeedSelectBox, value: string) => {
       player.setPlaybackSpeed(parseFloat(value));
-      this.selectItem(value);
     });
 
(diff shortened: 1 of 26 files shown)


src/ts/components/settings/PlaybackSpeedSelectBox.ts
```diff
@@ -30,9 +30,8 @@ export class PlaybackSpeedSelectBox extends SelectBox {
 
     this.addDefaultItems();
 
-    this.onItemSelected.subscribe((sender: PlaybackSpeedSelectBox, value: string) => {
+    this.onItemSelectionChanged.subscribe((sender: PlaybackSpeedSelectBox, value: string) => {
       player.setPlaybackSpeed(parseFloat(value));
-      this.selectItem(value);
```
stonko1994, line 35: This is now explicitly done in `dispatchItemSelectionChanged` already.

felix-hoc (approved):
Nice, tested locally and works on web 👍🏼

PR #820, opened 2026-02-06
[PW-16803] Match key and label for proper audio tracks synchronization

zenzi-deluxe:
## Description

Issue: https://bitmovin.atlassian.net/browse/PW-16803

(From the ticket) Audio tracks are mixed up in the main content audio track selector when multiple audio tracks are loaded across streams (e.g. pre-roll then content). 

The issue appears when audio tracks with the same key are matched in the synchronization function. When this happens, older audio tracks are retained, causing inconsistencies with the tracks prompted by the `playerAPI`. A simple fix would be to check for matching keys **and labels** when adding new tracks.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Time-related labels not properly updating the time format if the duration changes from below 1 hour to greater than 1 hour.
-- Audio tracks synchronization failing to update matching audio tracks with same key but different label.
+- Audio tracks not properly updating after CSAI ad.
 
 ## [4.8.0] - 2026-01-29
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Time-related labels not properly updating the time format if the duration changes from below 1 hour to greater than 1 hour.
+- Audio tracks synchronization failing to update matching audio tracks with same key but different label.
```
wasp898: We need something more customer facing, like `Audio tracks not correctly updated after CSAI ad.`

wasp898 (commented):
We could not find a better fix, could you please double check? 🙏  @stonko1994

CHANGELOG.md
zenzi-deluxe: Good point! I updated the `CHANGELOG` here: 1162857b7472583a998a10a1e3e499d583215de6.

wasp898 (changes requested):
Looks like some player events might be missing, so let's check that first.

zenzi-deluxe:
As mentioned by @wasp898, the issue was in the missing `AudioRemoved` and `AudioAdded` events, which are now properly dispatched on source loaded and unloaded. See [PR](https://github.com/bitmovin-engineering/bitdash/pull/8552) for more info.

PR #829, opened 2026-02-20
Automatically release every week

stonko1994:
## Description

Since we merged the v4 branch to dev we paused the continious deployment. This PR enables it again by adding a weekly release train workflow

- [x] `CHANGELOG` entry **Not added since this is non UI related**

### DIFF
--- a/.github/workflows/tag-release-version.yml
+++ b/.github/workflows/tag-release-version.yml
@@ -3,7 +3,7 @@ run-name: Weekly release train${{ github.event_name == 'workflow_dispatch' && ' 
 
 on:
   schedule:
-    - cron: '0 10 * * 5' # Every Friday at 10:00 UTC
+    - cron: '0 12 * * 4' # Every Thursday at 12:00 UTC
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/tag-release-version.yml
+++ b/.github/workflows/tag-release-version.yml
@@ -1,46 +1,47 @@
-name: Trigger release on merge
-run-name: Starting release for ${{ github.actor }} PR merge
+name: Weekly Release Train
+run-name: Weekly release train${{ github.event_name == 'workflow_dispatch' && ' (manual)' || '' }}
+
 on:
-  pull_request_target:
-    types:
-      - closed
-    branches:
-      - develop
-      - support/*
+  schedule:
+    - cron: '0 10 * * 5' # Every Friday at 10:00 UTC
+  workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-release-version.yml
```diff
@@ -1,46 +1,47 @@
...
-          echo "CHANGELOG.md has not been updated, skipping release"
-        fi
-      env:
-        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+            try {
+              detectReleaseLevel(core);
+              core.setOutput('has_unreleased', 'true');
+            } catch (error) {
+              core.info(`No unreleased changes found, skipping release: ${error.message}`);
+              core.setOutput('has_unreleased', 'false');
```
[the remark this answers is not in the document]
stonko1994: That's the point that we don't release in case there is no Unreleased section

.github/workflows/tag-release-version.yml
```diff
@@ -1,46 +1,47 @@
-name: Trigger release on merge
-run-name: Starting release for ${{ github.actor }} PR merge
+name: Weekly Release Train
+run-name: Weekly release train${{ github.event_name == 'workflow_dispatch' && ' (manual)' || '' }}
```
123mpozzi:
```suggestion
run-name: Weekly release train ${{ github.event_name == 'workflow_dispatch' && ' (manual)' || '' }}
```

Missing whitespace: would be `Weekly release train(manual)`

.github/workflows/tag-release-version.yml
```diff
@@ -1,46 +1,47 @@
...
-        else
-          echo "CHANGELOG.md has not been updated, skipping release"
-        fi
-      env:
-        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+            try {
+              detectReleaseLevel(core);
+              core.setOutput('has_unreleased', 'true');
+            } catch (error) {
+              core.info(`No unreleased changes found, skipping release: ${error.message}`);
```
123mpozzi, line 37: Nit, no need to change, but note for the future:

One thing to consider here is we treat any error as "no unreleased changes", which is fine by looking at when `detectReleaseLevel.js` throws explicitly.

However, there are some possible implicit throws as well since we use `fs.readFileSync` in that dependency, and we could consider any unexpected runtime error on the runner. For this reason we could differentiate between 'expected' throws (no unreleased entry, fine to skip release) vs unexpected throws (runtime error, release could still be good, but the error prevented it)

123mpozzi (approved):

PR #833, opened 2026-02-25
[PW-20545] Clean up the UI after destroying the player

zenzi-deluxe:
## Description

Issue: https://bitmovin.atlassian.net/browse/PW-20545
Also related to: 
- https://bitmovin.atlassian.net/browse/PW-6794
- https://bitmovin.atlassian.net/browse/PW-31280

This behavior is present in previous versions of the UI `v4` (and `v3`) as well. It generates from components of the UI (such as the `SeekBar`) calling the player APIs after the player has been destroyed. 
The proposed solution is to listen to the `PlayerEvent.Destroy` to release the UI and the event handlers in time.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- UI not released and still showing after calling `player.destroy()`
+
 ## [4.9.1] - 2026-02-24
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


wasp898 (changes requested):
As far as I know we already listen to the `Destroy` event and call release in:
https://github.com/bitmovin-engineering/bitdash/blob/463c39cc674904686f3c4899586c33f6d188363b/src/ui/ModuleDefinition.ts#L13-L17

Also I assume the SeekBar is released by the UIManager, so there shouldn't be a need to add any handling there.

The problem should be that `Destroy` itself is async, and so by the time it gets consumed, the player is most likely already destroyed. So my assumption is we should shift focus to the player side to find a solution, rather than UI...

zenzi-deluxe:
In my opinion, the player should not handle the disposal of the UI. Why not letting the UI manage itself?

wasp898:
> In my opinion, the player should not handle the disposal of the UI. Why not letting the UI manage itself?

As it came up on Slack, this is currently a feature: only if player manages the UI it should also release it. That wouldn't be the case if UI reacted to `PlayerEvent.Destroy`.

PR #835, opened 2026-03-02
Feature/fix fcc window color

LudoMichaud:
## Description

Add "!important" tag to the "window" styling (subtitle overlay). 

The Window styling for FCC subtitles was getting overridden probably here : 
https://github.com/bitmovin/bitmovin-player-ui/blob/751da29a4e3d1882c87317abdd65ebbb93ec5f6e/src/scss/components/overlays/_subtitle-overlay.scss#L23

- [ ] `CHANGELOG` entry

### DIFF
--- a/src/scss/components/settings/subtitlesettings/_subtitle-overlay-settings.scss
+++ b/src/scss/components/settings/subtitlesettings/_subtitle-overlay-settings.scss
@@ -70,9 +70,7 @@
     @each $opacity-name, $opacity-value in $opacities {
       &.#{$prefix}-windowcolor-#{$color-name}#{$opacity-name} {
         .#{$prefix}-subtitle-region-container {
-          .#{$prefix}-ui-subtitle-label {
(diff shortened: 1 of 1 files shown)


stonko1994 (commented):
Hi @LudoMichaud, thanks for the PR. I don't really see the original issue. can you elaborate how to reproduce the issue to identify whether the fix is appropriate?

LudoMichaud:
@stonko1994 sorry about that. I fixed this https://bitmovin.atlassian.net/browse/PW-28067

stonko1994:
Ok thank you I was able to reproduce the issue. However, I can still reproduce the issue on this branch 🤔 

> [!NOTE]
> In general we should not use `!important` anywhere in our UI as much as possible. 

As @felix-hoc already mentioned on the internal ticket, the root problem is that the `container-wrapper` has no size, and also the `!important` isn't changing that.

123mpozzi:
Work has continued here: https://github.com/bitmovin/bitmovin-player-ui/pull/848

PR #836, opened 2026-03-05
Handle multiple ad breaks at the same position

dweinber:
## Summary
- Adds `AdBreakTracker` utility to count current and total ads across ad breaks that share the same schedule time
- Updates `AdCounterLabel` to use `AdBreakTracker` so the ad counter reflects the total ad count when multiple breaks play at the same position
- Resets the tracker index when the last break in a group finishes

## Test plan
- [x] Ad counter label shows correct total ad count when multiple ad breaks are scheduled at the same time
- [x] Ad counter resets after the last break in a group finishes
- [x] Unit tests pass for `AdBreakTracker` and `AdCounterLabel`

## NOTE
This will only properly work with the Bitmovin Player Web 8.249.0+.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-### Fixed
+### Added
 
-- `AdCounterLabel` now correctly shows the ad position across multiple ad breaks scheduled at the same time (e.g. `Ad 2 of 3` instead of `Ad 1 of 1` for each)
+- `AdCounterLabel` now shows the ad position across multiple ad breaks scheduled at the same time (e.g. `Ad 2 of 3` instead of `Ad 1 of 1` for each)
 
 ## [4.9.1] - 2026-02-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- `AdCounterLabel` now correctly shows the ad position across multiple ad breaks scheduled at the same time (e.g. `Ad 2 of 3` instead of `Ad 1 of 1` for each)
+
 ## [4.9.1] - 2026-02-24
 
 ### Fixed
--- a/src/ts/components/ads/AdCounterLabel.ts
+++ b/src/ts/components/ads/AdCounterLabel.ts
@@ -41,10 +41,7 @@ export class AdCounterLabel extends Label<AdCounterLabelConfig> {
     this.player = player;
 
     this.adBreakTracker = uimanager.getConfig().adBreakTracker;
-
-    this.adBreakTracker.onAdCountChanged.subscribe((_, adBreakTrackerEvent: AdBreakTrackerAdCountChangedArgs) => {
-      this.setAdCounterFromAdBreakTracker(adBreakTrackerEvent.currentAdIndex, adBreakTrackerEvent.totalNumberOfAds);
-    });
+    this.adBreakTracker.onAdCountChanged.subscribe(this.adBreakTrackerAdCountChangedHandler);
 
     // An ad break may already be ongoing when configure is called, in this case the onAdCountChanged event was missed
     // and the label is set here
@@ -54,6 +51,8 @@ export class AdCounterLabel extends Label<AdCounterLabelConfig> {
   }
 
   release(): void {
+    this.adBreakTracker?.onAdCountChanged.unsubscribe(this.adBreakTrackerAdCountChangedHandler);
+
     this.adBreakTracker = undefined;
     this.player = undefined;
 
@@ -65,6 +64,13 @@ export class AdCounterLabel extends Label<AdCounterLabelConfig> {
       this.setAdCounterFromAdBreakTracker(this.adBreakTracker?.currentAdIndex, this.adBreakTracker?.totalNumberOfAds);
     }
   }
+
+  private readonly adBreakTrackerAdCountChangedHandler = (
+    _: AdBreakTracker,
+    adBreakTrackerEvent: AdBreakTrackerAdCountChangedArgs,
+  ) => {
+    this.setAdCounterFromAdBreakTracker(adBreakTrackerEvent.currentAdIndex, adBreakTrackerEvent.totalNumberOfAds);
+  };
 
   private setAdCounterFromAdBreakTracker(currentAdIndex?: number, totalNumberOfAds?: number) {
     if (currentAdIndex === 0 && totalNumberOfAds === 0) {
--- a/src/ts/components/ads/AdCounterLabel.ts
+++ b/src/ts/components/ads/AdCounterLabel.ts
@@ -3,7 +3,7 @@ import { UIInstanceManager } from '../../UIManager';
 import { LabelConfig, Label } from '../labels/Label';
 import { PlayerAPI } from 'bitmovin-player';
 import { StringUtils } from '../../utils/StringUtils';
-import { AdBreakTracker, AdBreakTrackerChangedArgs } from '../../utils/AdBreakTracker';
+import { AdBreakTracker, AdBreakTrackerAdCountChangedArgs } from '../../utils/AdBreakTracker';
 
 export interface AdCounterLabelConfig extends LabelConfig {
   /**
@@ -42,7 +42,7 @@ export class AdCounterLabel extends Label<AdCounterLabelConfig> {
 
     this.adBreakTracker = uimanager.getConfig().adBreakTracker;
 
-    this.adBreakTracker.onAdCountChanged.subscribe((_, adBreakTrackerEvent: AdBreakTrackerChangedArgs) => {
+    this.adBreakTracker.onAdCountChanged.subscribe((_, adBreakTrackerEvent: AdBreakTrackerAdCountChangedArgs) => {
       this.setText(
         StringUtils.replaceAdMessagePlaceholders(
           i18n.performLocalization(this.config.adCountOutOfTotal),
--- a/src/ts/components/ads/AdCounterLabel.ts
+++ b/src/ts/components/ads/AdCounterLabel.ts
@@ -7,7 +7,8 @@ import { AdBreakTracker, AdBreakTrackerChangedArgs } from '../../utils/AdBreakTr
 
 export interface AdCounterLabelConfig extends LabelConfig {
   /**
-   * Message displayed during the ad indicating which ad out of how many in the current ad break is currently playing.
+   * Message displayed during the ad indicating which ad out of how many is currently playing. It takes all ad breaks
+   * with the same schedule time into account.
    * Supported placeholders: look at {@link StringUtils.replaceAdMessagePlaceholders}
    */
   adCountOutOfTotal?: LocalizableText;
--- a/src/ts/components/ads/AdCounterLabel.ts
+++ b/src/ts/components/ads/AdCounterLabel.ts
@@ -54,34 +54,37 @@ export class AdCounterLabel extends Label<AdCounterLabelConfig> {
       );
     });
 
-    player.on(player.exports.PlayerEvent.AdBreakStarted, () => {
-      this.setText('');
-    });
-
-    player.on(player.exports.PlayerEvent.AdBreakFinished, () => {
-      this.setText('');
-    });
+    player.on(player.exports.PlayerEvent.AdBreakStarted, this.clearText);
+    player.on(player.exports.PlayerEvent.AdBreakFinished, this.clearText);
   }
 
+  clearText = () => {
+    super.clearText();
+  };
+
   release(): void {
+    this.player?.on(this.player.exports.PlayerEvent.AdBreakStarted, this.clearText);
+    this.player?.on(this.player.exports.PlayerEvent.AdBreakFinished, this.clearText);
+
     this.adBreakTracker?.release();
+
     this.adBreakTracker = undefined;
+    this.player = undefined;
+
     super.release();
   }
 
   protected onLanguageChanged(): void {
-    if (this.player?.ads?.isLinearAdActive?.()) {
-      if (this.adBreakTracker) {
-        this.setText(
-          StringUtils.replaceAdMessagePlaceholders(
-            i18n.performLocalization(this.config.adCountOutOfTotal),
-            this.player,
-            undefined,
-            this.adBreakTracker.currentAdIndex,
-            this.adBreakTracker.totalNumberOfAds,
-          ),
-        );
-      }
+    if (this.adBreakTracker?.currentAdIndex > 0 || this.player?.ads?.isLinearAdActive?.()) {
+      this.setText(
+        StringUtils.replaceAdMessagePlaceholders(
+          i18n.performLocalization(this.config.adCountOutOfTotal),
+          this.player,
+          undefined,
+          this.adBreakTracker?.currentAdIndex,
+          this.adBreakTracker?.totalNumberOfAds,
+        ),
+      );
     }
   }
 }
--- a/src/ts/components/ads/AdCounterLabel.ts
+++ b/src/ts/components/ads/AdCounterLabel.ts
@@ -43,16 +43,14 @@ export class AdCounterLabel extends Label<AdCounterLabelConfig> {
     this.adBreakTracker = uimanager.getConfig().adBreakTracker;
 
     this.adBreakTracker.onAdCountChanged.subscribe((_, adBreakTrackerEvent: AdBreakTrackerAdCountChangedArgs) => {
-      this.setText(
-        StringUtils.replaceAdMessagePlaceholders(
-          i18n.performLocalization(this.config.adCountOutOfTotal),
-          player,
-          undefined,
-          adBreakTrackerEvent.currentAdIndex,
-          adBreakTrackerEvent.totalNumberOfAds,
-        ),
-      );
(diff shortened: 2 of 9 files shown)


src/ts/utils/StringUtils.ts
```diff
@@ -107,9 +107,21 @@ export namespace StringUtils {
    * @param adMessage an ad message with optional placeholders to fill
    * @param player the player to get the time data from
    * @param skipOffset if specified, {remainingTime} will be filled with the remaining time until the ad can be skipped
+   * @param activeAdIndex if specified, {activeAdIndex} will be set to this value. Can be used to calculate the ad index
+   *   across multiple ad breaks which are scheduled for the same time. If not provided, the value will be calculated
+   *   for the current ad break only from the player API.
+   * @param totalNumberOfAds if specified, {totalAdsCount} wil be set to this value. Can be used to calculate the total
```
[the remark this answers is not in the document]
dweinber: Fixed in f4e62190

src/ts/utils/AdBreakTracker.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    if (
+      (this.adIndexOffsetOfPreviousBreaks > 0 && activeBreak.scheduleTime === this.groupScheduleTime) ||
+      siblings.length > 0
+    ) {
+      this.groupScheduleTime = activeBreak.scheduleTime;
+      this.currentAdIndexAcrossBreaks = withinBreakIndex + 1 + this.adIndexOffsetOfPreviousBreaks;
+      // Sibling ads arrays may not be populated yet (VAST manifests load lazily), so count
+      // sibling breaks and assume 1 ad each. totalAdsOverride self-corrects on each subsequent
+      // AdStarted as real ad counts become known.
```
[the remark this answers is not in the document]
dweinber: Updated in 708db2e6

src/ts/utils/AdBreakTracker.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    if (adBreak.scheduleTime !== this.groupScheduleTime) {
+      return;
+    }
+
+    this.adIndexOffsetOfPreviousBreaks += this.numberOfAdsInCurrentAdBreak;
+    this.numberOfAdsInCurrentAdBreak = 0;
+
+    this.dispatchChanged();
+  };
```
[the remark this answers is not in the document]
dweinber: Fixed in 3b1ad0f3

src/ts/utils/AdBreakTracker.ts
```diff
@@ -0,0 +1,141 @@
...
+    return this.events.onChanged.getEvent();
+  }
+
+  /**
+   * Index of the currently playing ad across all co-scheduled breaks (1-based), or 0 when no ad
+   * is active.
+   */
+  get currentAdIndex(): number {
+    return this.currentAdIndexAcrossBreaks;
+  }
```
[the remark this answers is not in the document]
dweinber: Fixed in 3b1ad0f3

src/ts/utils/StringUtils.ts
```diff
@@ -107,9 +107,21 @@ export namespace StringUtils {
    * @param adMessage an ad message with optional placeholders to fill
    * @param player the player to get the time data from
    * @param skipOffset if specified, {remainingTime} will be filled with the remaining time until the ad can be skipped
+   * @param activeAdIndex if specified, {activeAdIndex} will be set to this value. Can be used to calculate the ad index
+   *   across multiple ad breaks which are scheduled for the same time. If not provided, the value will be calculated
+   *   for the current ad break only from the player API.
+   * @param totalNumberOfAds if specified, {totalAdsCount} will be set to this value. Can be used to calculate the total
+   *   number of ads across multiple ad breaks which are scheduled for the same time. If not provided, the value will
+   *   be calculated for the current ad break only from the player API.
```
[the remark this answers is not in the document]
dweinber: Fixed in 3d09f046

src/ts/utils/StringUtils.ts
```diff
@@ -152,14 +164,14 @@ export namespace StringUtils {
           return formatNumber(0, formatString);
         }
```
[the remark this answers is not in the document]
dweinber: Changed in 67fe9834

src/ts/utils/StringUtils.ts
```diff
@@ -152,14 +164,14 @@ export namespace StringUtils {
...
+        if (formatString.includes('activeAdIndex')) {
+          const adIndex =
+            activeAdIndex ??
+            ads.findIndex(ad => (activeAd.id != null && ad.id != null ? ad.id === activeAd.id : ad === activeAd)) + 1;
+          return formatNumber(adIndex, formatString);
         }
 
-        return formatNumber(ads.length, formatString);
+        return formatNumber(totalNumberOfAds ?? ads.length, formatString);
       }
```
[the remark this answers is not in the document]
dweinber: Should be covered by 67fe9834 as well

src/ts/utils/AdBreakTracker.ts
```diff
@@ -0,0 +1,146 @@
...
+      return;
+    }
+
+    // Note: `player.ads.list()` provides all ad breaks except past ad breaks or the currently active ad break
+    const siblings = (this.player.ads?.list?.() ?? []).filter(b => b.scheduleTime === activeBreak.scheduleTime);
+
+    this.numberOfAdsInCurrentAdBreak = activeBreak.ads?.length ?? 0;
+
+    const activeAd = this.player.ads?.getActiveAd?.();
+    const ads = activeBreak.ads;
```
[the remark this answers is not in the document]
dweinber: Changed in a84cca9a

src/ts/utils/AdBreakTracker.ts
```diff
@@ -0,0 +1,147 @@
...
+    return this.events.onChanged.getEvent();
+  }
+
+  /**
+   * Index of the currently playing ad across all co-scheduled breaks (1-based), or 0 when no ad
+   * is active.
+   */
+  get currentAdIndex(): number {
+    return this.currentAdIndexAcrossBreaks;
+  }
```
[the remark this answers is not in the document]
dweinber: Fixed in 938bdaf7

src/ts/components/ads/AdCounterLabel.ts
```diff
@@ -46,20 +40,48 @@ export class AdCounterLabel extends Label<AdCounterLabelConfig> {
     super.configure(player, uimanager);
     this.player = player;
 
-    const clearText = () => {
+    this.adBreakTracker = new AdBreakTracker(player);
```
stonko1994: What do you think about moving that `AdBreakTracker` out of the `AdCounterLabel` into the `UIManager`? Having it here means that we can't easily use it from any other component (where we might need it in the future).

src/ts/components/ads/AdCounterLabel.ts
```diff
@@ -46,20 +40,48 @@ export class AdCounterLabel extends Label<AdCounterLabelConfig> {
...
+
+    this.adBreakTracker.onChanged.subscribe((_, adBreakTrackerEvent: AdBreakTrackerChangedArgs) => {
```
stonko1994: The API of the tracker isn't quite easy to follow. This code change came frist in the PR review and just seeing this without looking at the implementation I have no idea what actually changes in `adBreakTracker.onChanged`. Can we rename that to something more indicating what changes?

src/ts/components/ads/AdCounterLabel.ts
```diff
@@ -46,20 +40,48 @@ export class AdCounterLabel extends Label<AdCounterLabelConfig> {
...
+  release(): void {
+    this.adBreakTracker?.release();
+    this.adBreakTracker = undefined;
+    super.release();
   }
 
   protected onLanguageChanged(): void {
     if (this.player?.ads?.isLinearAdActive?.()) {
-      this.updateLabelText();
+      if (this.adBreakTracker) {
```
stonko1994: Is there any use-case where we not have an `adBreakTracker` but we show the label? I guess not, but should we fallback to the previous implementation if any?

src/ts/utils/AdBreakTracker.ts
```diff
@@ -0,0 +1,151 @@
...
+    player.on(player.exports.PlayerEvent.AdStarted, this.handleAdStarted);
+    player.on(player.exports.PlayerEvent.AdBreakFinished, this.handleAdBreakFinished);
+  }
+
+  get onChanged(): Event<AdBreakTracker, AdBreakTrackerChangedArgs> {
+    return this.events.onChanged.getEvent();
+  }
+
+  /**
+   * Index of the currently playing ad across all co-scheduled breaks (1-based), or 0 when no ad
```
stonko1994: Can we use a unified term for those ad breaks? Right now I see:
- co-scheduled
- group of co-scheduled
- Sibling

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- `AdCounterLabel` now correctly shows the ad position across multiple ad breaks scheduled at the same time (e.g. `Ad 2 of 3` instead of `Ad 1 of 1` for each)
```
stonko1994: IMO this is a new feature and not a fix. It worked as designed.

src/ts/utils/AdBreakTracker.ts
```diff
@@ -0,0 +1,151 @@
...
+  // Number of ads from co-scheduled ad breaks that have already finished.
+  private adIndexOffsetOfPreviousBreaks: number = 0;
+  // Index of the currently playing ad across all co-scheduled ad breaks (1-based)
+  private currentAdIndexAcrossBreaks: number = 0;
+  // Total ad count across the group of co-scheduled ad breaks
+  private totalNumberOfAdsAcrossBreaks: number = 0;
+  // scheduleTime shared by the current group of co-scheduled ad breaks, or undefined when not in a group.
+  private groupScheduleTime: number | undefined = undefined;
+  // Ad count of the currently active break
+  private numberOfAdsInCurrentAdBreak: number = 0;
```
stonko1994: Playing devil's advocate here. Do we need all of those states? Would it be possible to 'lazily' compute the necessary information without those states?

To me this feels like we only need this becuase the player removes old AdBreaks from the schedule array right? What if we track those on ourselves instead and use the `player.activeAd` as before?

stonko1994 (commented):
Did you test this changes on the mobile SDKs as well? Especially on versions where we don't support `player.ads.list` yet?

src/ts/utils/AdBreakTracker.ts
dweinber: 100% agree, I'm just not sure what the best term is 😅 "group of co-scheduled" is probably the clearest meaning, but it doesn't read very well, a but clumsy. Any thoughts or suggestions?

CHANGELOG.md
dweinber: From a customer perspective, it's not, but I get your point. Happy to change.

dweinber: a931d379

src/ts/components/ads/AdCounterLabel.ts
dweinber: I considered it, but as currently no other component uses it I thought it's fine to keep changes isolated. But I guess we might also want to add a total ad break duration somewhere or so...

src/ts/components/ads/AdCounterLabel.ts
dweinber: This comes mainly from the fact that we have several components where we just have a `onChanged` event 😄 but reading those again, it makes more sense in those cases than here, admittedly.

I'd go for `onAdCountsChanged` unless you have a better suggestion?
What triggers the event is 
a) an update of the current ad index counter (i.e. new ad starts)
b) the total count of ads across co-scheduled ad breaks (typically if not all ad details of a co-scheduled break is known yet, so we may not know how many ads are in that break)

dweinber: Alternative option that considers also e.g. total ad break group duration in the future could be `onAdBreakDetailsChanged`.

src/ts/components/ads/AdCounterLabel.ts
dweinber: I can't really think of a case tbh.

src/ts/components/ads/AdCounterLabel.ts
stonko1994: Could be a good use case and would qualify for moving it to a more general place.

src/ts/components/ads/AdCounterLabel.ts
stonko1994: `onAdCountChanged` sounds good (without the `s`) 👍

src/ts/utils/AdBreakTracker.ts
stonko1994: Hm what about `mergedAdBreaks` or even `subsequentAdBreaks`?

Actually this triggered another use-case in my head. What if multiple seeked-over ad breaks are playing back-to-back? I guess the current implementation wouldn't merge them together? (not in scoep - just asking)

src/ts/components/ads/AdCounterLabel.ts
dweinber: I thought the same and the change is already in progress :slightly_smiling_face:

src/ts/components/ads/AdCounterLabel.ts
dweinber: The `s` was on purpose to signify that multiple values may have changed, but I agree it sounds slightly strange. So fine for me

src/ts/utils/AdBreakTracker.ts
dweinber: Hm `subsequentAdBreaks` sounds good IMO.

Good use case, but that's not covered. It only works for ad breaks where the `scheduleTime` matches

src/ts/components/ads/AdCounterLabel.ts
dweinber: Changed to `onAdCountChanged ` in eb0cd5ee

src/ts/components/ads/AdCounterLabel.ts
dweinber: Moved it to the UIManager in 95734236, using a similar approach as we already use for the `VolumeController`.

src/ts/utils/AdBreakTracker.ts
dweinber: Renamed in d4d5e39c

src/ts/utils/AdBreakTracker.ts
dweinber: I think we need some state in either case mostly the player removes past and current ad breaks from the API. 

I tried a few different approaches, including storing the past ad breaks, but I'm not sure it gets better... just different 😅 

What also plays into this topic is that events are async, so I've seen that in the first `AdBreakFinished`, the next ad break is already removed from `player.ads.list` and the `AdBreakStarted` of the next is coming just milliseconds after the first `AdBreakFinished`. This leads easily to race conditions 😬 

I'd suggest we discuss offline (tomorrow) what the best approach would be.

dweinber: Changed the approach in b1d58c8b to keep track of past and the active ad break and calculate other details on the fly as needed.

src/ts/components/ads/AdCounterLabel.ts
```diff
@@ -46,20 +40,49 @@ export class AdCounterLabel extends Label<AdCounterLabelConfig> {
...
+  }
+
+  release(): void {
+    this.adBreakTracker?.release();
+
+    this.adBreakTracker = undefined;
+    this.player = undefined;
+
+    super.release();
   }
```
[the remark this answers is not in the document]
dweinber: Good catch, was a left over of the refactoring of moving the AdBreakTracker creation to the UI manager. Fixed in 2ab1e4ef and bc356493

src/ts/utils/AdBreakTracker.ts
```diff
@@ -0,0 +1,172 @@
...
+
+      // Add the active break if it's not already retained (new break in the group)
+      if (!this.groupBreaks.includes(activeBreak)) {
+        this.groupBreaks.push(activeBreak);
+      }
+    } else {
+      // Single ad break, not part of a group
+      this.groupBreaks = [activeBreak];
+      this.groupScheduleTime = undefined;
+    }
```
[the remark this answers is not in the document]
dweinber: Sounds like a valid concern, although I think that's rather an edge case. I've applied a different fix for this problem in 685017ed

src/ts/utils/AdBreakTracker.ts
```diff
@@ -0,0 +1,177 @@
...
+          return offset + activeAdIndex + 1;
+        }
+      }
+
+      offset += adCount;
+    }
+
+    // Active ad not found in any retained break — fall back to offset + 1
+    return offset + 1;
+  }
```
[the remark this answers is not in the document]
dweinber: Fixed in a4cb209d

stonko1994 (approved):
I tested it on mobile and it doesn't work there. However, I just figured out that on the small screen UI we don't even show the `Ad x from y` label at all. There we show the currentTime of the Ad. Propably because it didn't work and we don't have the `.schedule` API exposed to the UI bridge yet.

Just a fyi to keep in mind in case necessary also for mobile.

Edit: I was hoping that the implementation just fallsback to `Ad 1 of 1` on mobile until we properly expose it but it doesn't apparently.

dweinber:
I noticed that we don't show the AdCounterLabel in the SmallScreenUI, I thought it's just because of the available screen size. As we don't show it, let's postponed this to a future PR.

PR #838, opened 2026-03-23
Fix AdStatusOverlay blocking clicks on HugePlaybackToggleButton

dweinber:
## Summary

- Disable `pointer-events` on `AdStatusOverlay` and its descendants so clicks pass through to the `HugePlaybackToggleButton` underneath on small player sizes
- Re-enable `pointer-events` on `AdSkipButton` so it remains clickable

## Test plan

- [x] Load a small-sized player (e.g. 400x225) with an ad
- [x] Pause the ad and verify the `HugePlaybackToggleButton` is clickable
- [x] Verify the `AdSkipButton` still works when the ad becomes skippable

### DIFF
--- a/src/scss/components/ads/_ad-skip-button.scss
+++ b/src/scss/components/ads/_ad-skip-button.scss
@@ -3,6 +3,7 @@
 %ui-button-ad-skip {
   @extend %ui-button;
 
+  pointer-events: auto;
   background-color: transparentize($color-black, 0.25);
   border-radius: 20px;
   padding: 0.5em 1em;
--- a/src/scss/components/ads/_ad-skip-button.scss
+++ b/src/scss/components/ads/_ad-skip-button.scss
@@ -23,7 +23,7 @@
     }
   }
 
-  &:hover {
+  &:not(.#{$prefix}-disabled):hover {
     background-color: $color-black;
   }
 }
--- a/src/scss/components/ads/_ad-status-overlay.scss
+++ b/src/scss/components/ads/_ad-status-overlay.scss
@@ -8,10 +8,14 @@
   box-sizing: border-box;
   padding: 1em 1em 0.5em;
   bottom: 5em;
+  pointer-events: none;
+
+  * {
+    pointer-events: none;
+  }
 
   .#{$prefix}-bar {
     > .#{$prefix}-container-wrapper {
-      pointer-events: none;
       display: flex;
       margin: 0.5em 0;
     }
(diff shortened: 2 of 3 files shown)


src/scss/components/ads/_ad-status-overlay.scss
```diff
@@ -8,10 +8,14 @@
   box-sizing: border-box;
   padding: 1em 1em 0.5em;
   bottom: 5em;
+  pointer-events: none;
+
+  * {
+    pointer-events: none;
+  }
```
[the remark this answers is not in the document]
dweinber: Changed in 4b643189

src/scss/components/ads/_ad-skip-button.scss
```diff
@@ -3,6 +3,7 @@
 %ui-button-ad-skip {
   @extend %ui-button;
 
+  pointer-events: auto;
```
stonko1994, line 6: Looks like this slightly changes the UX as the skip button now always gets the hover state even though it's disabled before it is actually skippable.

Is this intentional?

src/scss/components/ads/_ad-status-overlay.scss
```diff
@@ -8,10 +8,18 @@
...
+
+  .#{$prefix}-ui-button-ad-skip {
+    pointer-events: auto;
+  }
```
stonko1994, line 19: Follow up from above. Maybe it's also this

src/scss/components/ads/_ad-skip-button.scss
dweinber: Not intentional, I didn't notice it to be honest. The disabled button doesn't get the "hand" pointer though.

In the current production version, the disabled button doesn't show the hover effect and clicking on it doesn't do anything. So this is what we want to keep?

dweinber: To be honest I don't fully understand why the hover effect wasn't shown before these changes. If we don't want the hover animation to show up, then we should also make sure that it's only enabled if the button is active. I've changed that now in 1d392809

src/scss/components/ads/_ad-status-overlay.scss
```diff
@@ -8,10 +8,18 @@
   box-sizing: border-box;
   padding: 1em 1em 0.5em;
   bottom: 5em;
+  pointer-events: none;
+
+  * {
+    pointer-events: none;
+  }
+
+  .#{$prefix}-ui-button-ad-skip {
```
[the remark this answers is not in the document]
dweinber: This is on purpose, the button should always capture the click/touch event, no matter if enabled or disabled.

stonko1994 (approved):

PR #846, opened 2026-04-03
Improve CEA captions legibility for small player sizes

123mpozzi:
## Description

Issue: CEA captions are not legible on very small players

**Before**

image

**After**

image

## Changes
- CEA-608 small-player legibility: captions on players rendered at or below 360px in height now use a proportional 80% safe area instead of fixed em margins, and the controlbar pushup is suppressed. Threshold is configurable via `UIConfig.cea608SmallPlayerHeightThreshold`.
- CEA grid stale cache: CEA-608 captions no longer keep stale sizing after CEA subtitle rendering is disabled and re-enabled, which could make captions appear incorrectly scaled or positioned.

## Manual testing

Use this quick sample to test the rendering on various player sizes:

```html
<!doctype html>
<html lang="en">
  <head>
    <title>Bitmovin Small Player Test</title>
    <meta charset="UTF-8" />
    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />

    <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/bitmovin-player@8/bitmovinplayer.js"></script>
    <script src="js/bitmovinplayer-ui.js"></script>
    <link rel="stylesheet" href="css/bitmovinplayer-ui.css" />

    <style>
      body {
        margin: 0;
        padding: 24px;
        font-family: Arial, sans-serif;
        background: #101114;
        color: #fff;
      }

      a {
        color: #8cc8ff;
      }

      .page {
        display: flex;
        flex-direction: column;
        gap: 16px;
      }

      .controls {
        display: flex;
        align-items: center;
        gap: 12px;
        flex-wrap: wrap;
      }

      .controls label {
        font-size: 13px;
        color: #aaa;
      }

      .controls select,
      .controls input {
        background: #1e1e1e;
        border: 1px solid #444;
        color: #fff;
        padding: 4px 8px;
        font-size: 13px;
      }

      .controls button {
        background: #2a2a2a;
        border: 1px solid #555;
        color: #fff;
        padding: 4px 12px;
        font-size: 13px;
        cursor: pointer;
      }

      .controls button:hover {
        background: #3a3a3a;
      }

      .size-indicator {
        font-size: 13px;
        color: #e6b31e;
        font-family: monospace;
      }

      .player-wrapper {
        box-shadow: 0 0 24px rgba(0, 0, 0, 0.45);
        transition: width 0.2s, height 0.2s;
      }

      #player {
        width: 100%;
        height: 100%;
      }

      .meta {
        max-width: 720px;
        font-size: 13px;
        line-height: 1.6;
        color: #aaa;
      }

      .test-actions {
        display: flex;
        gap: 8px;
        flex-wrap: wrap;
      }

      .font-controls {
        display: flex;
        align-items: center;
        gap: 8px;
        flex-wrap: wrap;
      }
    </style>
  </head>
  <body>
    <div class="page">
      <div><a href="index.html">&lt;- Back to playground</a></div>

      <div class="controls">
        <label>Preset:</label>
        <select id="preset">
          <option value="320x180">320×180 — well below threshold</option>
          <option value="400x225">400×225 — below threshold</option>
          <option value="427x240">427×240 — below threshold</option>
          <option value="480x270">480×270 — below threshold</option>
          <option value="640x360">640×360 — at threshold (≤360)</option>
          <option value="1280x720">1280×720 — above threshold</option>
        </select>

        <label>or custom W×H:</label>
        <input id="custom-w" type="number" value="320" min="100" max="1920" style="width:60px" />
        <span style="color:#aaa">×</span>
        <input id="custom-h" type="number" value="180" min="56" max="1080" style="width:60px" />
        <button id="apply-custom">Apply</button>

        <span class="size-indicator" id="size-indicator">320×180</span>
      </div>

      <div class="test-actions">
        <button id="run-cea608-row-test">Run CEA-608 All Rows Test</button>
        <button id="run-cea608-long-text-test">Run Long Text Test</button>
        <button id="clear-cea608-row-test">Clear Test Cues</button>
      </div>

      <div class="font-controls">
        <label for="cea608-font-size">CEA font size:</label>
        <select id="cea608-font-size">
          <option value="100">100%</option>
          <option value="150">150%</option>
          <option value="200">200%</option>
        </select>
        <button id="apply-cea608-font-size">Apply Font Size</button>
      </div>

      <div class="player-wrapper" id="player-wrapper">
        <div id="player"></div>
      </div>

      <div class="meta">
        Uses the same synthetic CEA-608 row pattern as the main playground, but in a constrained player so you can verify row placement across all 15 rows.
      </div>
    </div>

    <script type="text/javascript">
      var wrapper = document.getElementById('player-wrapper');
      var indicator = document.getElementById('size-indicator');
      var activeTestCues = [];

      function applySize(w, h) {
        wrapper.style.width = w + 'px';
        wrapper.style.height = h + 'px';
        indicator.textContent = w + '×' + h;
      }

      document.getElementById('preset').addEventListener('change', function () {
        var parts = this.value.split('x');
        var w = parseInt(parts[0]);
        var h = parseInt(parts[1]);
        document.getElementById('custom-w').value = w;
        document.getElementById('custom-h').value = h;
        applySize(w, h);
      });

      document.getElementById('apply-custom').addEventListener('click', function () {
        var w = parseInt(document.getElementById('custom-w').value);
        var h = parseInt(document.getElementById('custom-h').value);
        applySize(w, h);
      });

      applySize(320, 180);

      var playerConfig = {
        key: '',
        ui: false,
        playback: {
          autoplay: true,
          muted: true,
        },
      };

      var source = {
        // <your source with CEA captions>
      };

      window.player = new bitmovin.player.Player(document.getElementById('player'), playerConfig);
      window.uiManager = bitmovin.playerui.UIFactory.buildUI(window.player, { forceSubtitlesIntoViewContainer: true });

      function getUiPlayer() {
        return window.uiManager.currentUi.playerWrapper.getPlayer();
      }

      function cueEnter(text, row, column) {
        var cue = {
          text: text,
          position: {
            row: row,
            column: column,
          },
        };

        getUiPlayer().fireEventInUI(bitmovin.player.PlayerEvent.CueEnter, cue);
        return cue;
      }

      function cueExit(cue) {
        getUiPlayer().fireEventInUI(bitmovin.player.PlayerEvent.CueExit, cue);
      }

      function clearCea608RowTest() {
        activeTestCues.forEach(function (cue) {
          cueExit(cue);
        });
        activeTestCues = [];
      }

      function runCea608RowTest() {
        clearCea608RowTest();

        for (var row = 0; row < 15; row++) {
          var text = 'Test' + (row + 1);
          activeTestCues.push(cueEnter(text, row, 0));
          activeTestCues.push(cueEnter(text, row, 15 - text.length / 2 + 1));
          activeTestCues.push(cueEnter(text, row, 31 - text.length + 1));
        }
      }

      function runCea608LongTextTest() {
        var longText = 'XOXOX XOXO OXOX XOXOXOXOXOX XOXO';

        clearCea608RowTest();
        activeTestCues.push(cueEnter(longText.substring(0, 32), 0, 0));
        activeTestCues.push(cueEnter(longText.substring(0, 24), 7, 4));
        activeTestCues.push(cueEnter(longText.substring(0, 18), 14, 14));
      }

      function applyCea608FontSize() {
        var value = document.getElementById('cea608-font-size').value;
        window.uiManager.getSubtitleSettingsManager().fontSize.value = value;
      }

      document.getElementById('run-cea608-row-test').addEventListener('click', runCea608RowTest);
      document.getElementById('run-cea608-long-text-test').addEventListener('click', runCea608LongTextTest);
      document.getElementById('clear-cea608-row-test').addEventListener('click', clearCea608RowTest);
      document.getElementById('apply-cea608-font-size').addEventListener('click', applyCea608FontSize);

      window.player.load(source).catch(function (error) {
        console.error('Failed to load test stream', error);
      });
    </script>
  </body>
</html>
```

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- New `UIConfig.ceaCaptionPushupMinHeight` option (default `360`) to configure the rendered player height threshold below which small-player CEA-608 caption adjustments are applied
+
 ### Changed
 
-- CEA-608 captions now render within a centered 80% safe area and keep their row-based placement when the control bar is shown
+- CEA-608 captions on players shorter than `ceaCaptionPushupMinHeight` now render within a centered 80% safe area and keep their row-based placement when the control bar is shown
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-### Fixed
+### Changed
 
-- CEA captions not legible on very small player sizes
+- Improve CEA captions legibility when the rendered player size has a small height
 
 ## [4.11.0] - 2026-04-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
-- Improve CEA captions legibility when the rendered player size has a small height
+- CEA-608 captions now render within a centered 80% safe area and keep their row-based placement when the control bar is shown
+
+### Fixed
+
+- Reset cached CEA-608 grid sizing when CEA subtitle rendering is disabled so the next CEA session recalculates its layout correctly
 
 ## [4.11.0] - 2026-04-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
-- CEA-608 captions on players shorter than `ceaCaptionPushupMinHeight` now render within a centered 80% safe area and keep their row-based placement when the control bar is shown
+- CEA-608 captions on players with a rendered height below 360 CSS pixels (configurable via `ceaCaptionPushupMinHeight`) now render within a centered 80% safe area and keep their row-based placement when the control bar is shown
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- CEA captions not legible on very small player sizes
+
 ## [4.11.0] - 2026-04-03
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,15 +9,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- New `UIConfig.ceaCaptionPushupMinHeight` option (default `360`) to configure the rendered player height threshold below which small-player CEA-608 caption adjustments are applied
+- New `UIConfig.cea608SmallPlayerHeightThreshold` option (default `360`) to configure the rendered player height threshold at or below which small-player CEA-608 caption adjustments are applied
 
 ### Changed
 
-- CEA-608 captions on players with a rendered height below 360 CSS pixels (configurable via `ceaCaptionPushupMinHeight`) now render within a centered 80% safe area and keep their row-based placement when the control bar is shown
+- CEA-608 captions on players with a rendered height at or below 360 CSS pixels (configurable via `cea608SmallPlayerHeightThreshold`) now stay readable by using a centered 80% safe area and avoiding control-bar pushup that can crowd caption rows on small players
 
 ### Fixed
 
-- Reset cached CEA-608 grid sizing when CEA subtitle rendering is disabled so the next CEA session recalculates its layout correctly
+- CEA-608 captions could keep stale sizing after CEA subtitle rendering was disabled and re-enabled, causing captions to appear incorrectly scaled or positioned
 
 ## [4.11.0] - 2026-04-03
 
--- a/src/scss/components/overlays/_subtitle-overlay-cea608.scss
+++ b/src/scss/components/overlays/_subtitle-overlay-cea608.scss
@@ -4,10 +4,15 @@
   --cea608-row-height: math.div(100%, 15);
 
   &.#{$prefix}-cea608 {
-    bottom: 2em;
-    left: 3em;
-    right: 3em;
-    top: 2em;
+    // CEA-608 is rendered inside its centered 80% safe area.
+    bottom: math.percentage(math.div(1 - 0.8, 2));
+    left: math.percentage(math.div(1 - 0.8, 2));
+    right: math.percentage(math.div(1 - 0.8, 2));
+    top: math.percentage(math.div(1 - 0.8, 2));
+
+    &.#{$prefix}-controlbar-visible {
+      bottom: math.percentage(math.div(1 - 0.8, 2));
+    }
 
     .#{$prefix}-subtitle-region-container {
       height: var(--cea608-row-height);
@@ -17,11 +22,9 @@
       text-align: left;
 
       // Define positions for all 15 rows.
-      // --cea608-grid-offset shifts the grid upward when a minimum font size causes the full
-      // 15-row grid to exceed the overlay height, keeping bottom rows in view.
       @for $i from 0 through 14 {
         &.#{$prefix}-subtitle-position-cea608-row-#{$i} {
-          top: calc(var(--cea608-row-height) * #{$i} - var(--cea608-grid-offset, 0px));
+          top: calc(var(--cea608-row-height) * #{$i});
         }
       }
     }
--- a/src/scss/components/overlays/_subtitle-overlay-cea608.scss
+++ b/src/scss/components/overlays/_subtitle-overlay-cea608.scss
@@ -2,7 +2,7 @@
 
 // Inset applied on all four sides when the player is below the height threshold.
 // Keeps captions within the broadcast-standard 80% safe area at small sizes.
-$cea-safe-area-inset: math.percentage(math.div(1 - 0.8, 2));
+$cea608-safe-area-inset: math.percentage(math.div(1 - 0.8, 2));
 
 .#{$prefix}-ui-subtitle-overlay {
   --cea608-row-height: math.div(100%, 15);
@@ -13,20 +13,16 @@ $cea-safe-area-inset: math.percentage(math.div(1 - 0.8, 2));
     right: 3em;
     top: 2em;
 
-    // Below the rendered-height threshold the overlay is constrained to an 80% safe area and the controlbar pushup
-    // is disabled. Pushing captions further up at small player sizes reduces legibility, and the safe area keeps
-    // them away from the edges without relying on fixed em values that do not scale with player size.
-    &.#{$prefix}-cea-pushup-disabled {
-      bottom: $cea-safe-area-inset;
-      left: $cea-safe-area-inset;
-      right: $cea-safe-area-inset;
-      top: $cea-safe-area-inset;
+    // Constrains the overlay to the 80% safe area and suppresses the controlbar pushup at small player sizes.
+    &.#{$prefix}-cea608-pushup-disabled {
+      bottom: $cea608-safe-area-inset;
+      left: $cea608-safe-area-inset;
+      right: $cea608-safe-area-inset;
+      top: $cea608-safe-area-inset;
 
-      // Suppresses the controlbar pushup from _subtitle-overlay.scss. Both selectors compile to 0,3,0 specificity;
-      // this rule wins via source order (this file is imported after _subtitle-overlay.scss). Kept explicit for
-      // clarity and resilience against import-order changes.
+      // Higher-specificity override of the controlbar pushup from _subtitle-overlay.scss.
       &.#{$prefix}-controlbar-visible {
-        bottom: $cea-safe-area-inset;
+        bottom: $cea608-safe-area-inset;
       }
     }
 
--- a/src/scss/components/overlays/_subtitle-overlay-cea608.scss
+++ b/src/scss/components/overlays/_subtitle-overlay-cea608.scss
@@ -10,7 +10,9 @@
     right: math.percentage(math.div(1 - 0.8, 2));
     top: math.percentage(math.div(1 - 0.8, 2));
 
-    &.#{$prefix}-controlbar-visible {
+    // Below the small-player height threshold the controlbar pushup is disabled: captions stay anchored to the safe
+    // area regardless of controlbar state, since pushing them further up at small sizes reduces legibility.
+    &.#{$prefix}-cea-pushup-disabled.#{$prefix}-controlbar-visible {
       bottom: math.percentage(math.div(1 - 0.8, 2));
     }
 
--- a/src/scss/components/overlays/_subtitle-overlay-cea608.scss
+++ b/src/scss/components/overlays/_subtitle-overlay-cea608.scss
@@ -1,9 +1,5 @@
 @use 'sass:math';
 
-// Inset applied on all four sides when the player is below the height threshold.
-// Keeps captions within the broadcast-standard 80% safe area at small sizes.
-$cea608-safe-area-inset: math.percentage(math.div(1 - 0.8, 2));
-
 .#{$prefix}-ui-subtitle-overlay {
   --cea608-row-height: math.div(100%, 15);
 
@@ -15,6 +11,10 @@ $cea608-safe-area-inset: math.percentage(math.div(1 - 0.8, 2));
 
     // Constrains the overlay to the 80% safe area and suppresses the controlbar pushup at small player sizes.
     &.#{$prefix}-cea608-pushup-disabled {
+      // Inset applied on all four sides when the player is at or below the height threshold.
+      // Keeps captions within the broadcast-standard 80% safe area at small sizes.
+      $cea608-safe-area-inset: math.percentage(math.div(1 - 0.8, 2));
+
       bottom: $cea608-safe-area-inset;
       left: $cea608-safe-area-inset;
       right: $cea608-safe-area-inset;
--- a/src/scss/components/overlays/_subtitle-overlay-cea608.scss
+++ b/src/scss/components/overlays/_subtitle-overlay-cea608.scss
@@ -1,5 +1,9 @@
 @use 'sass:math';
 
+// Inset applied on all four sides when the player is below the height threshold.
+// Keeps captions within the broadcast-standard 80% safe area at small sizes.
+$cea-safe-area-inset: math.percentage(math.div(1 - 0.8, 2));
+
 .#{$prefix}-ui-subtitle-overlay {
   --cea608-row-height: math.div(100%, 15);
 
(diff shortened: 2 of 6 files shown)


jmsn (changes requested):
This is a rather pragmatic solution with some limitations worth discussing.

1. There are 15 lines in the CEA-608 grid. This approach only works if the captions are positioned on the lower 3 (?) lines. Any caption in lines above that would be cut off. Are you aware of this limitation? (I can share a test stream, if needed)
2. How were the 5% and 7% ratios chosen, eye-balled?
3. An alternative approach could be to switch off CEA-608 style rendering and render as "normal" subtitles if the player size is small. Maybe worth considering.

123mpozzi:
Those are all good considerations, thanks for bringing them up.

> 1. There are 15 lines in the CEA-608 grid. This approach only works if the captions are positioned on the lower 3 (?) lines. Any caption in lines above that would be cut off. Are you aware of this limitation? (I can share a test stream, if needed)

Good point, but then CEA are already broken due to the existing behavior, no? We have this feature to 'push' up captions when player gets focus and control-bar shows. This effectively compresses the rows vertically, which makes them not legible on smaller player sizes.

Given that CEA-608 defines a safe caption area of 80% of the video height, and the control bar alone can take up a significant portion of that (potentially up to ~40%), rendering all 15 rows legibly becomes impossible in such cases.

In my opinion, for CEA, row fidelity is the more correct behavior; I’m happy to open an internal discussion to align on this.

For now, I addressed this in 9892a34226e3a000320440e6c93fabfa606226a9 : CEA captions are now rendered inside their safe area and the pushup mechanism no longer applies to them.

Having a full-rows stream would be very helpful 🙏 , although I did find a sample online with captions in the upper rows, and the results looked good with these changes.

> 2. How were the 5% and 7% ratios chosen, eye-balled?

I got it from here: https://www.bbc.co.uk/accessibility/forproducts/guides/subtitles/#Authoring-font-size , they recommend a line height in the range 7% to 8%. It should result in a font size of slightly less, I guess between 5 to 6 % of the video height.

I picked 5 and 7 cause they looked really legible after some trials. I see, however, this is very arbitrary, so I removed the playerSize-dependent behavior in 9892a34226e3a000320440e6c93fabfa606226a9 


> 3. An alternative approach could be to switch off CEA-608 style rendering and render as "normal" subtitles if the player size is small. Maybe worth considering.

Definitely worth considering, but would still be broken with the control-bar pushup behavior. I'd try making it work with CEA style for now

123mpozzi:
In b7a2611dd06f99c8a6ff097c48715b82cec37ff9 the adjustments have been put under a tweak, as we agreed on in our internal meeting

jmsn (commented):
**Manual-testing snippet is stale** — the embedded HTML references font-size tiers (`7%` / `5%` min font) that no longer exist in the implementation (which has only a single threshold toggle). Please update or remove.

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -840,19 +863,54 @@ function isCea608SubtitleCue(cue: SubtitleCueEvent): boolean {
 }
 
 function awaitTransitionEnd(domElement: DOM) {
```
jmsn, line 910: **Consider fixing at the call site instead of adding a fallback timer**

The new fallback timeout looks like a bandaid workaround that we should avoid. When `cea-pushup-disabled` is active the overlay's `bottom` no longer changes when the controlbar shows, so `transitionend` correctly never fires — there is no transition to wait for. The fallback still waits `duration + delay + 50ms` (~350ms for a 300ms CSS transition) before calling `ensureCea608GridSizeUpdated`, with no corresponding layout change during that window.

Instead, the callers of `awaitTransitionEnd` should check whether the pushup is disabled and, if so, skip the await and run the recalc immediately. Adding a `cea608PushupDisabled` private field or a method that determines this state would both be good options.

With this change `awaitTransitionEnd` can revert to its pre-PR form, and the new `getMaximumTransitionTimeoutMs` + `parseCssTimeToMs` helpers can be removed — we only call `awaitTransitionEnd` when a transition is actually going to run.

src/ts/UIConfig.ts
```diff
@@ -186,6 +186,28 @@ export interface UIConfig {
...
+   * At small player sizes fixed `em` margins take up a disproportionate share of the available space, and pushing
+   * captions up when the controlbar appears can make them illegible. The 80% safe area and the suppressed pushup
+   * together keep CEA-608 captions legible and correctly positioned at these sizes.
+   *
+   * Set to `0` to disable both adjustments for all player sizes (restoring the default large-player behaviour
+   * everywhere).
+   *
+   * Default: `360`
+   */
+  ceaCaptionPushupMinHeight?: number;
```
jmsn: Suggestion: rename to `cea608SmallPlayerHeightThreshold` — consistent with the `cea608` prefix used elsewhere, unambiguous about being a threshold rather than a minimum, and neutral about the specific effects.

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -33,6 +33,8 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
 
   private static readonly CLASS_CONTROLBAR_VISIBLE = 'controlbar-visible';
   private static readonly CLASS_CEA_608 = 'cea608';
+  private static readonly CLASS_CEA_PUSHUP_DISABLED = 'cea-pushup-disabled';
+  private static readonly DEFAULT_CEA_CAPTION_PUSHUP_MIN_HEIGHT = 360;
```
jmsn: Nit: align with the existing `cea608` prefix (e.g. `CLASS_CEA608_PUSHUP_DISABLED = 'cea608-pushup-disabled'`, `DEFAULT_CEA608_SMALL_PLAYER_HEIGHT_THRESHOLD`).

src/scss/components/overlays/_subtitle-overlay-cea608.scss
```diff
@@ -1,5 +1,9 @@
 @use 'sass:math';
 
+// Inset applied on all four sides when the player is below the height threshold.
+// Keeps captions within the broadcast-standard 80% safe area at small sizes.
+$cea-safe-area-inset: math.percentage(math.div(1 - 0.8, 2));
```
jmsn: Nit: align with the existing `cea608` prefix — `$cea608-safe-area-inset`, class `cea608-pushup-disabled`.

src/scss/components/overlays/_subtitle-overlay-cea608.scss
```diff
@@ -9,14 +13,31 @@
...
+    &.#{$prefix}-cea-pushup-disabled {
+      bottom: $cea-safe-area-inset;
+      left: $cea-safe-area-inset;
+      right: $cea-safe-area-inset;
+      top: $cea-safe-area-inset;
+
+      // Suppresses the controlbar pushup from _subtitle-overlay.scss. Both selectors compile to 0,3,0 specificity;
+      // this rule wins via source order (this file is imported after _subtitle-overlay.scss). Kept explicit for
+      // clarity and resilience against import-order changes.
+      &.#{$prefix}-controlbar-visible {
```
jmsn, line 24: The code comment above is overly verbose and the specificity is stated incorrectly — this rule compiles to (0,4,0), not (0,3,0). Suggest trimming to a one-liner, e.g. `// Higher-specificity override of the controlbar pushup from _subtitle-overlay.scss.`

src/scss/components/overlays/_subtitle-overlay-cea608.scss
```diff
@@ -9,14 +13,31 @@
     right: 3em;
     top: 2em;
 
+    // Below the rendered-height threshold the overlay is constrained to an 80% safe area and the controlbar pushup
+    // is disabled. Pushing captions further up at small player sizes reduces legibility, and the safe area keeps
+    // them away from the edges without relying on fixed em values that do not scale with player size.
+    &.#{$prefix}-cea-pushup-disabled {
```
jmsn: Suggest dropping the "Below the rendered-height threshold…" framing — the comment should state what the rule does, not under what circumstances it's applied.

123mpozzi: Yeah, that's a better way to document the style, address in 6481b78a7d17d5316d90511542be5aca0e3b86a1

src/scss/components/overlays/_subtitle-overlay-cea608.scss
123mpozzi: Good catch! Address in 6481b78a7d17d5316d90511542be5aca0e3b86a1

src/scss/components/overlays/_subtitle-overlay-cea608.scss
123mpozzi: Address in 6481b78a7d17d5316d90511542be5aca0e3b86a1

src/ts/components/overlays/SubtitleOverlay.ts
123mpozzi: Address in 598a5a2713690992a88a420b538d032a87e06af8

src/ts/UIConfig.ts
123mpozzi: Address in 6481b78a7d17d5316d90511542be5aca0e3b86a1

src/ts/components/overlays/SubtitleOverlay.ts
123mpozzi: Thanks for the detailed comment! 💪 
Address in 598a5a2713690992a88a420b538d032a87e06af8

123mpozzi:
> **Manual-testing snippet is stale** — the embedded HTML references font-size tiers (`7%` / `5%` min font) that no longer exist in the implementation (which has only a single threshold toggle). Please update or remove.

Thanks for letting me know 🙌 , I've updated it!

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -145,7 +154,8 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
       if (component instanceof ControlBar) {
         this.getDomElement().addClass(this.prefixCss(SubtitleOverlay.CLASS_CONTROLBAR_VISIBLE));
 
-        if (this.cea608Enabled && this.ensureCea608GridSizeUpdated) {
+        if (this.cea608Enabled && this.ensureCea608GridSizeUpdated &&
+            !this.getDomElement().hasClass(this.prefixCss(SubtitleOverlay.CLASS_CEA608_PUSHUP_DISABLED))) {
```
jmsn: please extract to const for readability

```
const isCea608PushupTransitionEnabled = !this.getDomElement().hasClass(this.prefixCss(SubtitleOverlay.CLASS_CEA608_PUSHUP_DISABLED))
```

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -155,7 +165,8 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
       if (component instanceof ControlBar) {
         this.getDomElement().removeClass(this.prefixCss(SubtitleOverlay.CLASS_CONTROLBAR_VISIBLE));
 
-        if (this.cea608Enabled && this.ensureCea608GridSizeUpdated) {
+        if (this.cea608Enabled && this.ensureCea608GridSizeUpdated &&
+            !this.getDomElement().hasClass(this.prefixCss(SubtitleOverlay.CLASS_CEA608_PUSHUP_DISABLED))) {
```
jmsn: same here, please extarct to const

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -164,6 +175,15 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
     this.configureCea608Captions(player, uimanager);
     // Init
     subtitleClearHandler();
+    this.updateCeaPushupClass(new DOM(player.getContainer()).height());
+  }
+
+  private updateCeaPushupClass(playerHeight: number): void {
```
jmsn: function name leaks implementation detail, suggest renaming to something like `updateCea608PushupFromPlayerHeight`

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -164,6 +175,15 @@ export class SubtitleOverlay extends Container<ContainerConfig> {
...
+    if (playerHeight < this.cea608SmallPlayerHeightThreshold) {
```
jmsn: NIT: Would it make sense to instead use `<=` here so that configuring `cea608SmallPlayerHeightThreshold: 360` treats a 360px-tall player as small?

If yes, also please update the JSDoc and CHANGELOG wording from "below" to "at or below" accordingly.

CHANGELOG.md
```diff
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
...
+
+- New `UIConfig.ceaCaptionPushupMinHeight` option (default `360`) to configure the rendered player height threshold below which small-player CEA-608 caption adjustments are applied
+
+### Changed
+
+- CEA-608 captions on players with a rendered height below 360 CSS pixels (configurable via `ceaCaptionPushupMinHeight`) now render within a centered 80% safe area and keep their row-based placement when the control bar is shown
+
+### Fixed
+
+- Reset cached CEA-608 grid sizing when CEA subtitle rendering is disabled so the next CEA session recalculates its layout correctly
```
jmsn: this description is rather low level, what was the UX implication of this? 🙏

CHANGELOG.md
```diff
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
...
 
+## [Unreleased]
+
+### Added
+
+- New `UIConfig.ceaCaptionPushupMinHeight` option (default `360`) to configure the rendered player height threshold below which small-player CEA-608 caption adjustments are applied
+
+### Changed
+
+- CEA-608 captions on players with a rendered height below 360 CSS pixels (configurable via `ceaCaptionPushupMinHeight`) now render within a centered 80% safe area and keep their row-based placement when the control bar is shown
```
jmsn: the config option name is outdated

src/scss/components/overlays/_subtitle-overlay-cea608.scss
```diff
@@ -1,5 +1,9 @@
 @use 'sass:math';
 
+// Inset applied on all four sides when the player is below the height threshold.
+// Keeps captions within the broadcast-standard 80% safe area at small sizes.
+$cea608-safe-area-inset: math.percentage(math.div(1 - 0.8, 2));
```
jmsn: This is a global scope variable.

Worth considering vowing it into the scope of `&.#{$prefix}-cea608-pushup-disabled {`

src/ts/components/overlays/SubtitleOverlay.ts
123mpozzi: abf62453c646a7e9d4afcb808721359bdad0607c

src/ts/components/overlays/SubtitleOverlay.ts
123mpozzi: good idea! address in abf62453c646a7e9d4afcb808721359bdad0607c

src/ts/components/overlays/SubtitleOverlay.ts
123mpozzi: Thanks for the naming suggestion 🙌 
- address in abf62453c646a7e9d4afcb808721359bdad0607c

src/ts/components/overlays/SubtitleOverlay.ts
123mpozzi: Good point, `360` should be in it, otherwise it might be misleading. Address in abf62453c646a7e9d4afcb808721359bdad0607c

CHANGELOG.md
123mpozzi: Address in abf62453c646a7e9d4afcb808721359bdad0607c so the readability issue is explicitly stated

CHANGELOG.md
123mpozzi: Good catch! Address in abf62453c646a7e9d4afcb808721359bdad0607c

src/scss/components/overlays/_subtitle-overlay-cea608.scss
123mpozzi: Address in abf62453c646a7e9d4afcb808721359bdad0607c

jmsn (approved):

PR #847, opened 2026-04-04
Add comparator funcs to control audio and subtitle tracks ordering

123mpozzi:
## Description

Currently there's no way to control the tracks ordering. For some devices, like iOS, tracks might not be ordered alphabetically due to asynchronous loading and addition of tracks.

This PR address this issue by introducing a comparator function for list selectors, allowing integrators to define the display order for audio and subtitle tracks.

### Changes

- Adds an optional comparator callback to `ListSelector`, allowing to customize the display order of its elements

- Adds `ListSelector.onItemsChanged` which fires when the effective item collection changes, including reordering and item updates, so UI consumers can rebuild from the current list state

- Extend `AudioTrackListBox` and `SubtitleListBox` to accept config objects (in addition to title-only), so that `comparator`, `filter`, and `translator` can be passed through

 - Keeps the built-in `Off` subtitle option pinned to the top when a subtitle comparator is used

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- `AudioTrackListBox` and `SubtitleListBox` now accept config objects that extend `ListBoxConfig` without `listSelector`
 - `ListSelectorConfig.comparator` to customize the display order of list-backed selection UIs such as `AudioTrackSelectBox`, `SubtitleSelectBox`, `AudioTrackListBox`, and `SubtitleListBox`. For subtitle selection UIs, the built-in `Off` option remains fixed at the top.
   Example:
   ```ts
@@ -18,6 +17,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
     comparator: (a, b) => String(a.label).localeCompare(String(b.label)), // A-Z
   });
   ```
+- `AudioTrackListBox` and `SubtitleListBox` now can accept config objects that extend `ListBoxConfig` without `listSelector`.
+  This enables configuring list-selector behavior such as `comparator`, `filter`, and `translator`, as well as list-box/settings-panel options like `title`, `hideDelay`, and related panel settings.
 
 ## [4.11.0] - 2026-04-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- `ListSelectorConfig.comparator` to customize the display order of list-backed selection UIs such as `AudioTrackSelectBox`, `SubtitleSelectBox`, `AudioTrackListBox`, and `SubtitleListBox`.
+- `AudioTrackListBox` and `SubtitleListBox` now accept config objects that extend `ListBoxConfig` without `listSelector`
+- `ListSelectorConfig.comparator` to customize the display order of list-backed selection UIs such as `AudioTrackSelectBox`, `SubtitleSelectBox`, `AudioTrackListBox`, and `SubtitleListBox`. For subtitle selection UIs, the built-in `Off` option remains fixed at the top.
   Example:
   ```ts
-    const subtitleListBox = new SubtitleListBox({
-      title: i18n.getLocalizer('settings.subtitles'),
-      comparator: (a, b) => String(a.label).localeCompare(String(b.label)), // A-Z
-    });
+  const subtitleListBox = new SubtitleListBox({
+    title: i18n.getLocalizer('settings.subtitles'),
+    comparator: (a, b) => String(a.label).localeCompare(String(b.label)), // A-Z
+  });
   ```
 
 ## [4.11.0] - 2026-04-03
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   ```
 - `AudioTrackListBox` and `SubtitleListBox` now can accept config objects that extend `ListBoxConfig` without `listSelector`.
   This enables configuring list-selector behavior such as `comparator`, `filter`, and `translator`, as well as list-box/settings-panel options like `title`, `hideDelay`, and related panel settings.
+- `ListSelector.onItemsChanged` event that fires whenever the effective item collection changes, including item additions, removals, reordering, and item property updates that affect rendering
 
 ## [4.11.0] - 2026-04-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Added
+
+- `UIConfig.audioTrackComparator` and `UIConfig.subtitleComparator` to control the display order of audio and subtitle tracks in the selection UI
+
 ## [4.11.0] - 2026-04-03
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- `ListSelectorConfig.comparator` to control the display order of list-backed selection UIs such as audio and subtitle track lists
+- `ListSelectorConfig.comparator` to customize the display order of list-backed selection UIs such as `AudioTrackSelectBox`, `SubtitleSelectBox`, `AudioTrackListBox`, and `SubtitleListBox`.
+  Example:
+  ```ts
+    const subtitleListBox = new SubtitleListBox({
+      title: i18n.getLocalizer('settings.subtitles'),
+      comparator: (a, b) => String(a.label).localeCompare(String(b.label)), // A-Z
+    });
+  ```
 
 ## [4.11.0] - 2026-04-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   ```
 - `AudioTrackListBox` and `SubtitleListBox` now can accept config objects that extend `ListBoxConfig` without `listSelector`.
   This enables configuring list-selector behavior such as `comparator`, `filter`, and `translator`, as well as list-box/settings-panel options like `title`, `hideDelay`, and related panel settings.
-- `ListSelector.onItemsChanged` event that fires whenever the effective item collection changes, including item additions, removals, reordering, and item property updates that affect rendering
+- `ListSelector.onItemsChanged` event that fires whenever the effective item collection changes, including item additions, removals, reordering, or updates to rendered item data such as `label` and `ariaLabel`.
+  This event should be used by consumers that rebuild the full list UI from `getItems()`, while `onItemAdded` and `onItemRemoved` remain strict membership-change events.
 
 ## [4.11.0] - 2026-04-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
     comparator: (a, b) => String(a.label).localeCompare(String(b.label)), // A-Z
   });
   ```
+  > **Note**: Requires building a custom UI. The default `UIFactory` presets do not expose this option.
 - `AudioTrackListBox` and `SubtitleListBox` now can accept config objects that extend `ListBoxConfig` without `listSelector`.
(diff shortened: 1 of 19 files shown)


src/ts/UIConfig.ts
```diff
@@ -186,6 +186,28 @@ export interface UIConfig {
...
+   * Comparator function to define a custom display order for subtitle tracks in the subtitle selection UI.
+   *
+   * When set, subtitle tracks returned by the player will be sorted using this comparator before being displayed.
+   * The function follows the same contract as {@link Array.prototype.sort}.
+   *
+   * @example
+   * // Sort subtitle tracks alphabetically by label
+   * subtitleComparator: (a, b) => a.label.localeCompare(b.label)
+   */
+  subtitleComparator?: (a: SubtitleTrack, b: SubtitleTrack) => number;
```
stonko1994: I think we should move this to the `ListSelector`. There we already have the `ListItemLabelTranslator` and `ListItemFilter` and this sorting falls into the same category.

To set those the `AudioTrackListBox` and `SubtitleListBox` also need to be adapted. Their constructor need to support the `ListBox`/`ListBoxConfig` constructor from its super class. The `ListBoxConfig` extends the `ListSelectorConfig` which will eventually enable setting those new 'sorting' config.

123mpozzi: I thought of that, but we lose the functionality of ordering via other fields that we have access to using Audio and SubtitleTrack.

However, I think it makes sense, it's more fitting and label names ordering are the intended use-case - address in the current version

CHANGELOG.md
```diff
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
...
+- `ListSelectorConfig.comparator` to customize the display order of list-backed selection UIs such as `AudioTrackSelectBox`, `SubtitleSelectBox`, `AudioTrackListBox`, and `SubtitleListBox`. For subtitle selection UIs, the built-in `Off` option remains fixed at the top.
+  Example:
+  ```ts
+  const subtitleListBox = new SubtitleListBox({
+    title: i18n.getLocalizer('settings.subtitles'),
+    comparator: (a, b) => String(a.label).localeCompare(String(b.label)), // A-Z
+  });
+  ```
+- `AudioTrackListBox` and `SubtitleListBox` now can accept config objects that extend `ListBoxConfig` without `listSelector`.
+  This enables configuring list-selector behavior such as `comparator`, `filter`, and `translator`, as well as list-box/settings-panel options like `title`, `hideDelay`, and related panel settings.
```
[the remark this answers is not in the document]
123mpozzi: Resolved with an updated PR description

src/ts/components/lists/ListSelector.ts
```diff
@@ -212,13 +231,37 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
...
+      this.selectedItem = null;
+    }
+
+    for (const key of removedKeys) {
+      this.onItemRemovedEvent(key);
+    }
+    for (const key of addedKeys) {
+      this.onItemAddedEvent(key);
+    }
   }
```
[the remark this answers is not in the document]
123mpozzi: Address in d88652f032793f6c6c66345d11996c95da575540

src/ts/components/lists/ListSelector.ts
```diff
@@ -116,33 +193,20 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
...
     }
 
-    // Apply translator function
-    if (this.config.translator) {
-      listItem.label = this.config.translator(listItem);
+    const existingIndex = this.getItemIndex(key);
+    if (existingIndex > -1) {
+      ArrayUtils.remove(this.items, this.items[existingIndex]);
+      this.onItemRemovedEvent(key);
     }
```
jmsn, line 211: why can't `this.removeItem()` be called here instead?

src/ts/components/lists/ListSelector.ts
```diff
@@ -209,22 +274,59 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
...
+    if (this.selectedItem !== null && !nextKeys.has(this.selectedItem)) {
+      this.selectedItem = null;
+    }
+
+    for (const key of removedKeys) {
+      this.onItemRemovedEvent(key);
+    }
+    for (const key of addedKeys) {
+      this.onItemAddedEvent(key);
+    }
```
jmsn, line 317: can this be done above when removing and adding keys? like 

```ts
    const removedKeys = this.items
      .filter(item => !nextKeys.has(item.key))
      .map(item => item.key)
      .forEach(key => this.onItemRemovedEvent(key)) // <<< add it here?
```

src/ts/components/lists/ListSelector.ts
```diff
@@ -90,6 +126,47 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
...
+      const previousItem = previousItems[i];
+      const nextItem = nextItems[i];
+
+      if (
+        previousItem.key !== nextItem.key ||
+        i18n.performLocalization(previousItem.label) !== i18n.performLocalization(nextItem.label) ||
+        previousItem.ariaLabel !== nextItem.ariaLabel
+      ) {
+        return true;
+      }
```
jmsn, line 150: Are these additional comparisons (beyond the key) really needed/relevant?

Is it correct to assume that `haveItemsChanged` aims to check if keys have been added or removed, is that correct?

src/ts/components/lists/ListSelector.ts
```diff
@@ -80,6 +102,20 @@ export abstract class ListSelector<Config extends ListSelectorConfig> extends Co
     this.items = this.config.items;
   }
 
+  private normalizeItem(listItem: ListItem): ListItem | null {
```
jmsn, line 108: NIT: short function doc would not hurt, as the name is pretty abstract

jmsn (commented):
review part 1, mainly `ListSelector.ts`

src/ts/utils/SubtitleUtils.ts
```diff
@@ -70,9 +74,26 @@ export class SubtitleSwitchHandler {
...
+
+    if (!this.hasComparator()) {
+      if (!this.listElement.hasItem(subtitle.id)) {
+        this.listElement.addItem(subtitle.id, subtitle.label);
+      }
+      return;
     }
+
+    const subtitles = this.player.subtitles.list();
+    const mergedSubtitles = subtitles.some(track => track.id === subtitle.id) ? subtitles : [...subtitles, subtitle];
```
jmsn: The "subtitle" vs. "subtitles" naming makes this line unnecessarily hard to read. Please rename one of the two variables.

src/ts/utils/SubtitleUtils.ts
```diff
@@ -70,9 +74,26 @@ export class SubtitleSwitchHandler {
...
+    const offListItem: ListItem = {
+      key: SubtitleSwitchHandler.SUBTITLES_OFF_KEY,
+      label: i18n.getLocalizer('off'),
+    };
+
+    this.listElement.synchronizeItems([
+      offListItem,
+      ...mergedSubtitles.map(track => ({ key: track.id, label: track.label })),
```
jmsn: NIT: you can re-use the `subtitleToListItem` function here

src/ts/utils/AudioTrackUtils.ts
```diff
@@ -44,11 +44,25 @@ export class AudioTrackSwitchHandler {
...
+
+    if (!this.hasComparator()) {
+      if (!this.listElement.hasItem(audioTrack.id)) {
+        this.listElement.addItem(audioTrack.id, i18n.getLocalizer(audioTrack.label), true);
+      }
+      return;
     }
+
+    const audioTracks = this.player.getAvailableAudio();
+    const mergedTracks = audioTracks.some(track => track.id === audioTrack.id) ? audioTracks : [...audioTracks, audioTrack];
```
jmsn: Same here, please make it more readable by renaming

src/ts/utils/AudioTrackUtils.ts
```diff
@@ -44,11 +44,25 @@ export class AudioTrackSwitchHandler {
...
+
+    this.listElement.synchronizeItems(mergedTracks.map(track => ({ key: track.id, label: track.label })));
```
jmsn: Use `audioTrackToListItem`

src/ts/components/lists/ListBox.ts
```diff
@@ -66,31 +71,42 @@ export class ListBox extends SettingsPanel<ListBoxConfig> {
...
+    };
+
+    this.listSelector.onItemsChanged.subscribe(onItemsChanged);
 
     this.settingsPanelPage.configure(player, uimanager);
     this.listSelector.configure(player, uimanager);
+
+    if (!initialBuildDone) {
+      rebuildItems();
+    }
```
jmsn: Claude tells me it's debatable whether this is necessary, and seeing any action -- maybe worth double-checking.

If this is kept, please add a short comment.

src/ts/UIConfig.ts
stonko1994: Would it be possible to add a generic field to the `ListItem` (and with that through all the `ListSelector` class) to enable that again?

src/ts/components/lists/ListSelector.ts
123mpozzi: Because calling `removeItem()` would fire `onItemsChanged` before the replacement item is inserted, which would expose an intermediate state and cause an extra rebuild (via `onItemsChangedEvent`)

src/ts/components/lists/ListSelector.ts
123mpozzi: Because that would fire `onItemRemoved` before `synchronizeItems()` has finished updating the selector state.
That ordering matters. It means any listener reacting to `onItemRemoved` or `onItemsChanged` sees the final synchronized state, not a half-finished one.

src/ts/utils/SubtitleUtils.ts
123mpozzi: Thanks for raising this, it was indeed hard to read.

Renamed the locals to remove the subtitle/subtitles ambiguity in a9df91cf49d9b030af87d4a55ce403cd30454a1a

src/ts/utils/SubtitleUtils.ts
123mpozzi: Address in a9df91cf49d9b030af87d4a55ce403cd30454a1a

src/ts/utils/AudioTrackUtils.ts
123mpozzi: Address in a9df91cf49d9b030af87d4a55ce403cd30454a1a

src/ts/utils/AudioTrackUtils.ts
123mpozzi: Address in a9df91cf49d9b030af87d4a55ce403cd30454a1a

src/ts/components/lists/ListBox.ts
123mpozzi: I've kept the guard and documented it in a9df91cf49d9b030af87d4a55ce403cd30454a1a. `listSelector.configure()` can synchronously trigger `onItemsChanged` via the track handlers, so this prevents a duplicate initial rebuild.

Without the guard, `configure()` would:
1. rebuild through the event
2. then fall through and rebuild again

src/ts/components/lists/ListSelector.ts
123mpozzi: That's a good idea! Added a short doc comment to clarify the normalization step in 49f68c023ff214ac5ce00a4db13fc607d1dd0e06

src/ts/components/lists/ListSelector.ts
123mpozzi: I would keep them, `haveItemsChanged` is an "effective rendered item changed" check. So `ariaLabel` belongs there for the same reason the localized label does: it affects what the UI exposes. If it changes, the rendered option may need to be rebuilt so accessibility text stays correct

src/ts/UIConfig.ts
123mpozzi: I had a quick look and I think this makes sense as a future extension, but I’d avoid adding it in this PR.

Adding generics through `ListItem` / `ListSelector` would expand the public API surface, so it’s not a small, risk-free change here.

A smaller alternative would be adding an opaque field like `data?: unknown` to `ListItem`, but then we’d need to define whether payload changes should affect `haveItemsChanged()` / `onItemsChanged()`.
Right now those semantics are intentionally limited to effective rendered-item changes (keys/order/labels/aria), so I’d keep this PR scoped and revisit payload support separately if we have a concrete use case.

src/ts/components/lists/ListSelector.ts
jmsn: Okay, this duality of `onItemsChanged` and  `onItemAdded`+`onItemRemoved` is another topic, seems a bit redundant as`onItemAdded`/`onItemRemoved` are probably not used a lot anymore.

I guess, not a blocker but worth noting.

jmsn (approved):
LGTM,  assuming [this concern](https://github.com/bitmovin/bitmovin-player-ui/pull/847#pullrequestreview-4073652651) is sorted

PR #849, opened 2026-04-09
Fix VTT background color applies color to window instead

123mpozzi:
## Description

Applying background color to VTT subtitles applies to color to the window instead.

## Changes

window color was applied to the generic subtitle container, which is not the visible cue box for plain WebVTT cues.

This branch adds separate VTT cue/container classes and applies window color to the actual cue box when there is no WebVTT region.

Additionally, it fixes cue updates when a WebVTT cue changes container type.

## Testing Instructions

On demo page, ensure that:
- For VTT: background color applies to background color (color behind text) and window color to window color (color whole row)
- For CEA: background color and window color still works as expected

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fixed subtitle window coloring for non-region WebVTT cues
-- Fix WebVTT subtitle background color applies window color instead
+- WebVTT subtitle background color applies styling to window color instead
 
 ## [4.11.0] - 2026-04-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Subtitle window coloring was not applied for non-region WebVTT cues
-- WebVTT subtitle background color applies styling to window color instead
+- WebVTT subtitle background color is applied to the window instead of the text background
 
 ## [4.11.0] - 2026-04-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fixed subtitle window coloring for non-region WebVTT cues
+- Fix WebVTT subtitle background color applies window color instead
 
 ## [4.11.0] - 2026-04-03
 
--- a/src/scss/components/overlays/_subtitle-overlay.scss
+++ b/src/scss/components/overlays/_subtitle-overlay.scss
@@ -63,6 +63,24 @@
       }
     }
 
+    .#{$prefix}-ui-subtitle-label.#{$prefix}-subtitle-vtt-cue,
+    .#{$prefix}-subtitle-vtt-region-container .#{$prefix}-ui-subtitle-label {
+      .#{$prefix}-ui-label-text {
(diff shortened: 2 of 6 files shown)


felix-hoc:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

src/scss/components/overlays/_subtitle-overlay.scss
```diff
@@ -63,6 +63,24 @@
...
+      }
+
+      &[style*='text-align: center'] .#{$prefix}-ui-label-text {
+        margin-left: auto;
+        margin-right: auto;
+      }
+
+      &[style*='text-align: right'] .#{$prefix}-ui-label-text,
+      &[style*='text-align: end'] .#{$prefix}-ui-label-text {
+        margin-left: auto;
```
felix-hoc: Per the [WebVTT cue text alignment spec](https://www.w3.org/TR/webvtt1/#webvtt-cue-text-alignment), `end` is a *logical* value (inline-end direction), while `right` is physical. For RTL cues (Arabic, Hebrew), `align: end` maps to the visual left — so `margin-left: auto` pushes the wrong way.

Split these into separate rules:
```scss
&[style*='text-align: right'] .#{}-ui-label-text {
  margin-left: auto;
}
&[style*='text-align: end'] .#{}-ui-label-text {
  margin-inline-start: auto; // logical: end = left in RTL, right in LTR
}
```

src/scss/components/overlays/_subtitle-overlay.scss
```diff
@@ -63,6 +63,24 @@
...
+      }
```
felix-hoc, line 88: `text-align: start` (the logical counterpart to `end`) is not handled. For RTL cues with `align: start`, the text should be at the visual right, but with no rule here the element stays at the CSS default (visual left). Add:
```scss
&[style*='text-align: start'] .#{}-ui-label-text {
  margin-inline-end: auto;
}
```
Ref: [WebVTT §cue-text-alignment](https://www.w3.org/TR/webvtt1/#webvtt-cue-text-alignment)

src/scss/components/overlays/_subtitle-overlay.scss
123mpozzi: Great points, thanks for the detailed replies!

Address in f281551c9c79b4d9f8d50b57c7d7eefbf629cc98

src/scss/components/overlays/_subtitle-overlay.scss
123mpozzi: Address in f281551c9c79b4d9f8d50b57c7d7eefbf629cc98

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Subtitle window coloring was not applied for non-region WebVTT cues
+- WebVTT subtitle background color applies styling to window color instead
```
felix-hoc: nit: sounds a bit strange, maybe we should rephrase?
```suggestion
- WebVTT subtitle background color is applied to the window instead of the text background
```

felix-hoc (approved):
LGTM, just a nit. Also not sure if the base branch is correct, please check before merging :)

CHANGELOG.md
123mpozzi: Yes, that sounds better, thanks for the suggestion!

PR #865, opened 2026-05-05
Fix `APINotAvailableError` on UI release

zenzi-deluxe:
## Description

Fixes a race condition where `player.isCasting()` (and other player API calls inside `hideUi`/`showUi` closures) could be called after the player was already destroyed.

### Root cause

`UIContainer` creates a `Timeout` that calls `hideUi` after `hideDelay` ms. When the timeout's delay elapses, the callback is placed in the JS macrotask queue. If the player is destroyed and `release()` is called in the same execution context, `clearTimeout` is a no-op because the timer has already fired. The callback is already queued and will execute after `release()` returns, hitting the destroyed player's API.

### This fix

Call `hideUi` at the very top of `release()`, causing the next `hideUi` to become a no-op, hence blocking all player API calls after the player has been torn down.

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/UIContainer.ts
+++ b/src/ts/components/UIContainer.ts
@@ -65,7 +65,6 @@ export class UIContainer extends Container<UIContainerConfig> {
   private static readonly CONTROLS_HIDDEN = 'controls-hidden';
 
   private uiHideTimeout: Timeout;
-  private isReleased: boolean = false;
   private playerStateChange: EventDispatcher<UIContainer, PlayerUtils.PlayerState>;
 
   private userInteractionEventSource: DOM;
@@ -157,9 +156,6 @@ export class UIContainer extends Container<UIContainerConfig> {
     };
 
     this.showUi = () => {
-      if (this.isReleased) {
-        return;
-      }
       isHideUiPending = false;
 
       if (!isUiShown) {
@@ -174,9 +170,6 @@ export class UIContainer extends Container<UIContainerConfig> {
     };
 
     this.hideUi = (force: boolean = false) => {
-      if (this.isReleased) {
-        return;
-      }
       // Hide the UI only if it is shown, and if not casting
       if (isUiShown && !player.isCasting()) {
         if (force) {
@@ -505,7 +498,8 @@ export class UIContainer extends Container<UIContainerConfig> {
   }
 
   release(): void {
-    this.isReleased = true;
+    // Hide the UI to make sure hideUi becomes a no-op and avoid race conditions with the hide timeout while releasing
+    this.hideUi(true);
 
     // Explicitly unsubscribe user interaction event handlers because they could be attached to an external element
     // that isn't owned by the UI and therefore not removed on release.
--- a/src/ts/components/UIContainer.ts
+++ b/src/ts/components/UIContainer.ts
@@ -7,7 +7,7 @@ import { CancelEventArgs, Event as UiEvent, EventDispatcher } from '../EventDisp
 import { PlayerAPI, PlayerResizedEvent } from 'bitmovin-player';
 import { i18n } from '../localization/i18n';
 import { Button, ButtonConfig } from './buttons/Button';
-import { TouchControlOverlay, TouchControlOverlayConfig } from './overlays/TouchControlOverlay';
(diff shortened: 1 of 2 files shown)


zenzi-deluxe:
Without the changes, this appears after running the following test:

Screenshot 2026-05-05 at 14 22 45

src/ts/components/UIContainer.ts
stonko1994, line 514: We already do this 🤔 why isn't that enough? Based on your PR description I'd assume this is enough? Otherwise should we move that before the `super.release()`.

zenzi-deluxe: I tried moving it up, but it did not help.
We cancel recurring timeouts, but apparently the callback functions (`hideUi`) still get executed once. That's probably a really unfortunate case, when the timeout already fires, adding the callback to the execution queue. By the time the callback gets executed, if the player gets destroyed in the same instant, the reference is null and the callback throws. I tried different approaches, but could not figure out a better way to handle this. 
If you have any idea, I can try it out.

wasp898: I was suggesting that we call `this.hideUi(true)` in `release`, but not sure if that will solve it. Would have to be tested 😄

stonko1994: I'm not super familiar how `setInterval` work in JS and what happens on `clearInterval`. Would it help if we wrap the callback we pass into the `setInterval` with a state check of `isActive` which we check right before we call the `callback`?

---

I'm not sure if calling `hideUi` in `release` helps as the player might be already destroyed regardless or? Which brings me to an idea actually: Should we add this into this component?

```ts
this.player.on(this.player.exports.PlayerEvent.Destroy, () => { this.uiHideTimeout.clear(); });
```

zenzi-deluxe: My understanding is that `clearInterval` should remove existing timeouts, so we would not have any more recurring executions of the callback. The issue happens when the timeout is cleared, but the callback has just been added to the tasks queue for execution. In that case, we clear the following timeout intervals, but cannot prevent the callback to be executed, since it's already set to be executed. That's why the `APINotAvailableError` is logged only once and the reason behind adding the guard inside the callback.

---

Regarding the second suggestion, I tried adding a listener for `PlayerEvent.Destroy`, but we incur in the same race condition. I noticed this issue happens with only few tests, which are executing and passing very fast. This is inline with my suspicion of `player.destroy()` and `Ui.release()` being called in the same execution context.

wasp898: `release` is called during the player teardown and it's not async so it should work if we hide the UI then. The `Destroy` event is still async so it would be worse.
I think if my suggestion works, we could just go with that. Release would hide the UI, so even if a pre-scheduled task runs asynchronously, it would lead to a no-op as the UI is already hidden.

zenzi-deluxe: Calling it also solves the issue, as the first condition short-circuits on the `isUiShown` being false after hiding the UI in the `release` and `player.isCasting()` is never checked. However, any event between that call and timer firing could call `showUi()`, setting `isUiShown = true` again. Also, it would not protect `showUi` from the `player.isCasting()` call. It might never happen, but that's a risk.

wasp898: I was trying to avoid introducing a flag, but if we prefer a more general solution, I'd be fine with that. 🙂

zenzi-deluxe: @stonko1994 what'd you say?

zenzi-deluxe: We can go with @wasp898's approach and if it happens again, introduce the flag and use the more general approach. Sounds good?

zenzi-deluxe: Added here: [e41764e](https://github.com/bitmovin/bitmovin-player-ui/pull/865/commits/e41764e4398b27d2f1378b5cd6e54938a77dbee3)

stonko1994 (approved):

src/ts/components/UIContainer.ts
stonko1994: sounds good to me 👍

wasp898: 👍 please update the description

PR #867, opened 2026-05-08
Add video stats overlay and right-click context menu

lukaskroepfl:
## Summary

Adds a YouTube-style "stats for nerds" debug overlay and a right-click context menu to the default player UI, plus the settings-panel entry point that makes both reachable on touch / TV / WebView platforms. Everything is gated behind a single `UIConfig` opt-out for production deployments.

### `DebugInfoOverlay`

Small "video stats" overlay in the player corner. Hidden by default; opened via the context menu / settings-panel row.

Lines rendered (only those the source actually exposes):

- Video: `1920×818@24 6.21 Mbps (avc1.4d4028)` + `↓ <downloaded bitrate>` when ABR has switched
- Audio: `130 kbps (mp4a.40.2)` + `↓ <downloaded bitrate>`
- Resolution: `<decoded> → <element size>` (catches up/down-scaling)
- Color: `bt709 / bt709` — parsed from AV1 / VP9 codec strings per ITU-T H.273 (omitted on AVC/HEVC, which don't encode color in the codec)
- Buffer (video / audio in seconds)
- Dropped frames
- Time + speed (`0:42 / 14:48 @ 1.50×`)
- Live latency (`0.42s behind edge`, only for live sources)
- Available qualities (`9 video / 3 audio`) + codec families (`avc1, hevc, av01`)
- DRM systems, manifest URL (truncated), stream + player versions

Refreshes every 1 s while playing and on quality / seek / source events; timer pauses when the overlay is hidden. Survives the controls auto-hide (reparented to the player container, not the UI container). Text is **selectable** and selections survive refreshes (no innerHTML rewrite while a selection is active). The overlay can be **dragged** anywhere on the page including outside the player bounds, and follows native fullscreen (prefers `document.fullscreenElement` over `<body>` on reparent so it stays visible).

### `PlayerContextMenu`

Right-click context menu shown over the player. Reparented into `document.fullscreenElement` when present, otherwise `<body>`.

- Bitmovin Player heading + subtitle
- Player and UI versions
- "About Bitmovin →" link
- "Show / Hide video stats" toggle (label flips with the overlay state)
- "Copy link at current time" — produces `<page-url>?t=<seconds>s`, preserves existing query params + fragment, replaces an existing `t=`
- "Copy source" — `JSON.stringify(player.getSource())`
- "Copy player config" — `JSON.stringify(player.getConfig())`
- "Copy debug info" — structured snapshot for support: timestamp, page URL, user agent, player + UI version, source URLs + DRM systems, full playback state, current quality + parsed color, audio, codec arrays, track / quality counts

Right-clicking on the `<video>` element itself is **skipped** so the browser's native video context menu (Save video as, Picture-in-Picture, …) keeps working. `role="menu"` + `role="menuitem"` for AT; click outside / `Esc` dismiss; hidden state sets `pointer-events: none` so it doesn't intercept clicks. The `navigator.clipboard.writeText` path catches rejections and falls back to a textarea + `execCommand('copy')` for insecure-context / unfocused-document cases.

### Touch / TV / WebView entry point

The right-click menu isn't reachable without a mouse, so the small-screen layout's settings panel grows a new full-row tappable entry **"Player Info & Video Stats"** that opens the same `PlayerContextMenu` centered over the player. Whole row is the click target (not just the label), has `cursor: pointer`, `role="menuitem"`, `tabindex=0`, hover/focus styling, and keyboard `Enter` / `Space` activation. The desktop `main` layout omits the row — right-click already works there.

### `UIConfig.disableDebugUi`

One opt-out flag that removes everything in this PR — overlay, context menu, and the settings row — from the default and small-screen layouts. Defaults to `false` (debug UI shown). Use this in production where the debug surface should not be exposed to end users.

```ts
UIFactory.buildUI(player, { disableDebugUi: true });
```

## Public API additions

- `DebugInfoOverlay`, `DebugInfoOverlayConfig`
- `PlayerContextMenu`, `PlayerContextMenuConfig`
- `parseCodecColor(codec)`, `collectCodecFamilies(codecs[])`, `formatBitrate`, `formatSeconds`, `truncateMiddle`, `copyToClipboard`, `jsonReplacer`, `buildTimestampLink` (helpers, exported for reuse)
- `UIConfig.disableDebugUi`
- 10 new i18n keys (`videoStats.*`, `contextMenu.*`, `settings.playerInfo`) — added to `en.json` only; non-English vocabularies fall back to English via the existing vocab merge in `I18n#initializeVocabulary`

## Accessibility

- `PlayerContextMenu` carries `role="menu"`; each action button carries `role="menuitem"`.
- The settings-panel row carries `role="menuitem"`, `tabindex=0`, full keyboard activation.
- Localized strings are refreshed when the UI language changes (`onLanguageChanged` override on both components).
- Overlay text is selectable for copy-paste.

## Targets / constraint handling

- **ES5 target preserved.** No `URL` constructor, no modern Web APIs in hot paths. Plain regex string parsing for timestamp / color extraction.
- **Native fullscreen.** Both overlay (on drag) and menu (on show) prefer `document.fullscreenElement` (with the WebKit-prefixed fallback) over `<body>` so they stay visible in fullscreen.
- **Lifecycle.** Both components override `release()` to clear timers, unsubscribe from player / dispatcher events, remove document-level listeners, restore parent nodes, and drop reparented elements from `<body>`. Verified no double-release.
- **Storage.** Nothing in this PR writes to `localStorage` directly (the related preference / resume work lives in PR #869).

## Bundle size

Production build, `develop` vs. branch tip, gzipped: **+4.6 KB JS** + **+0.7 KB CSS** ≈ **+5.3 KB total on the wire** for the entire feature.

## Test plan

- [x] All 516 unit tests pass; ESLint and `tsc --noEmit` are clean (16 new tests across `DebugInfoOverlay`, `PlayerContextMenu`, `SettingsToggleButton`, `Localization`).
- [x] **Desktop (main layout):** right-click on the player → menu opens centered on click; all 4 copy actions copy correctly with "Copied!" flash; toggle shows / hides the overlay; right-click on the `<video>` element shows the browser's native menu, not ours.
- [x] **Small-screen (mobile):** tap settings cog → "Player Info & Video Stats" row visible at the bottom; the whole row is tappable; tapping opens the same menu centered over the player.
- [ ] **TV / spatial nav:** the settings row is keyboard-focusable; Enter / Space opens the menu.
- [ ] **Auto-hide:** with the controls auto-hidden, the overlay remains visible.
- [ ] **Drag:** drag the overlay header outside the player bounds — it stays attached to the cursor and persists where dropped.
- [ ] **Fullscreen:** enter native fullscreen with the overlay shown / menu open → both stay visible.
- [ ] **Selection:** select text inside the overlay, wait through several refresh ticks — the selection survives.
- [ ] **Language switch:** change the UI language at runtime — header / subtitle / button labels update.
- [ ] **`disableDebugUi: true`:** no `bmpui-ui-player-context-menu` / `bmpui-ui-debug-info-overlay` elements in the DOM, no "Player Info & Video Stats" row in the settings panel, right-click falls through to the browser default.

image


stonko1994:
Manual testing outcome: (code review still pending)

1. When selecting an HLS asset I get an exception in the console: (Might be also because of muxed content)
Screenshot 2026-05-11 at 15 00 39

2. On iOS the Player and Stream just shows unknown
Bildschirmfoto 2026-05-11 um 15 01 47

_there are also quite a few other infos missing on iOS:_
- Dropped frames: not exposed
- available video: not possible on iOS (will always show 0)

2.1. Copying source config / player config on mobile
Using the copy `SourceConfig`/`PlayerConfig` on mobile almost doesn't include any information. _They are just not present on mobile_

3. UI/UX feedback
3.1 General UI
The stats for nerds and the context menu don't really 'feel at home' in the UI. They look like an element which doesn't really belong to the rest of the UI. We use a quite transparent and rounded corner approach e.g. in the settings panel. The new overlays do not follow that and feel like a 'termina' to me. Personally I'd prefer to match the other UI design language.

3.2 Scroling
When scrolling while one of the overlays is visible they are static:

https://github.com/user-attachments/assets/3a7ceea1-8ffa-4d26-9d64-30fcda0d1729

3.3 About Bitmovin
Do we really want to include the `About Bitmovin` in such a view? Imagine this on an embedded customer page. IMO would feel weird.

3.4 Small screen menu
I find the `Player Info & Video Stats` too prominent in the settings panel on mobile (and the small screen UI on web). I'd prefer adding a `More` section and putting it there (similar to youtube)

3.5 Resolution -> Resolution
Screenshot 2026-05-11 at 15 17 02

Without the context of the PR description I had no idea what the `Resolution -> Resolution` line means. Maybe we should add a label for the second property as well.

src/ts/components/overlays/DebugInfoOverlay.ts
```diff
@@ -0,0 +1,518 @@
...
+    super.configure(player, uimanager);
+
+    // Reparent the overlay to the player container so it stays visible when the
+    // surrounding UI auto-hides (which sets `display: none` on the UI container).
+    const playerContainer = player.getContainer();
+    const overlayEl = this.getDomElement().get(0) as HTMLElement;
+    this.originalParent = overlayEl.parentElement;
+    if (overlayEl.parentElement !== playerContainer) {
+      playerContainer.appendChild(overlayEl);
+    }
```
stonko1994: I don't think that this is the correct approach to keep the overlay visible all the time. We already have other always visible elements such as the WaterMark.

I think if you don't include `@include hidden-animated;` the overlay should not hide in the first place.

src/ts/components/overlays/DebugInfoOverlay.ts
```diff
@@ -0,0 +1,518 @@
...
+
+    this.update = () => this.updateContent(player);
+    this.onUpdatedHandler = this.update;
+    this.uiManagerRef = uimanager;
+
+    const startTimer = () => {
+      this.stopTimer();
+      if (!this.isShown()) return;
+      this.refreshTimer = window.setInterval(this.update, this.config.refreshIntervalMs);
```
stonko1994: Please don't use `setInterval` directly. Use our `Timeout` util for something like this.

src/ts/components/overlays/DebugInfoOverlay.ts
```diff
@@ -0,0 +1,518 @@
...
+    player.on(player.exports.PlayerEvent.StallStarted, this.update);
+    player.on(player.exports.PlayerEvent.StallEnded, this.update);
+    player.on(player.exports.PlayerEvent.PlaybackFinished, () => this.stopTimer());
+    player.on(player.exports.PlayerEvent.Destroy, () => this.stopTimer());
+
+    this.onShow.subscribe(() => {
+      this.update();
+      if (!player.isPaused()) startTimer();
+    });
+    this.onHide.subscribe(() => this.stopTimer());
```
stonko1994, line 125: Why do we need this if it's always visible?

src/ts/components/overlays/DebugInfoOverlay.ts
```diff
@@ -0,0 +1,518 @@
...
+      rootElement.removeClass(draggingClass);
+      document.removeEventListener('pointermove', onPointerMove);
+      document.removeEventListener('pointerup', onPointerUp);
+      document.removeEventListener('pointercancel', onPointerUp);
+    };
+
+    this.onPointerMoveDocument = onPointerMove;
+    this.onPointerUpDocument = onPointerUp;
+
+    this.headerElement.on('pointerdown', (e: PointerEvent) => {
```
stonko1994, line 210: This doesn't seem to work nicely on iOS. It triggers the swipe-back system gesture.

https://github.com/user-attachments/assets/f44580c4-bcc5-4c3d-a4dc-6683e6440a82

src/ts/components/overlays/DebugInfoOverlay.ts
```diff
@@ -0,0 +1,518 @@
...
+    if (manifest) lines.push(`Manifest: ${truncateMiddle(manifest, 60)}`);
+    lines.push(`Stream: ${streamType} (${playerType})`);
+    lines.push(`Player: ${playerVersion}`);
+
+    // Don't clobber a user's text selection inside the overlay. Reassigning innerHTML
+    // tears down the text node and collapses any active selection range, so skip writes
+    // while the user is mid-copy.
+    if (this.hasSelectionInside()) return;
+
+    const next = escapeHtml(lines.join('\n'));
```
stonko1994: Can we find a better variable name than `next`? Just reading this feels like next of what? Suggestions:
- `playerStatsHtml`

src/ts/components/overlays/DebugInfoOverlay.ts
```diff
@@ -0,0 +1,518 @@
...
+ *
+ * Common values:
+ *  - 1: bt709
+ *  - 5: bt470bg
+ *  - 6: smpte170m
+ *  - 9: bt2020
+ *  - 16: smpte2084 (PQ, HDR10)
+ *  - 18: arib-std-b67 (HLG)
+ */
+export function parseCodecColor(codec: string): { primaries: string; transfer: string } | null {
```
stonko1994: I kinda see this as player responsibility to provide this information.

src/ts/components/overlays/DebugInfoOverlay.ts
```diff
@@ -0,0 +1,518 @@
...
+    if (audioQuality) {
+      lines.push(`Audio: ${formatBitrate(audioQuality.bitrate)} (${audioQuality.codec || '?'})`);
+    }
+    if (downloadedAudio && !sameQuality(audioQuality, downloadedAudio)) {
+      lines.push(`  ↓ ${formatBitrate(downloadedAudio.bitrate)}`);
+    }
+    if (videoElement && videoElement.videoWidth > 0) {
+      const decoded = `${videoElement.videoWidth}×${videoElement.videoHeight}`;
+      const rendered = `${videoElement.clientWidth}×${videoElement.clientHeight}`;
+      lines.push(`Resolution: ${decoded} → ${rendered}`);
```
stonko1994: Texts like 'Resolution' here are not localized. I'm fine with this as it's 'just' stats for nerds but jfyi as we localize the title etc.

src/ts/components/PlayerContextMenu.ts
```diff
@@ -0,0 +1,444 @@
+import { Container, ContainerConfig } from './Container';
+import { DOM } from '../DOM';
+import { UIInstanceManager } from '../UIManager';
+import { PlayerAPI } from 'bitmovin-player';
+import { DebugInfoOverlay, parseCodecColor } from './overlays/DebugInfoOverlay';
+import { i18n } from '../localization/i18n';
+
+// Webpack's `string-replace-loader` substitutes {{VERSION}} with a JSON-stringified
+// value (i.e. surrounded by quotes), so peel the outer quotes back off for display.
+const UI_VERSION: string = '{{VERSION}}'.replace(/^"|"$/g, '');
```
stonko1994: I'd prefer to not introduce a second `{{VERSION}}`. Even though it's replaced at compile time I think we should rather re-use the already existing version from `main.ts`. Can't you simply import it?

src/ts/components/PlayerContextMenu.ts
```diff
@@ -0,0 +1,444 @@
...
+    const separator = new DOM('div', {
+      class: this.prefixCss('ui-player-context-menu-separator'),
+    });
+
+    const actionButton = () =>
+      new DOM('button', {
+        type: 'button',
+        class: this.prefixCss('ui-player-context-menu-button'),
+        role: 'menuitem',
+      });
```
stonko1994: Is there a reason why we can't re-use our own Button components?

src/ts/components/PlayerContextMenu.ts
```diff
@@ -0,0 +1,444 @@
...
+  public showCentered(): void {
+    // Use the player container (always laid out) rather than the UI container (which can
+    // be hidden by the auto-hide controls timer, returning a 0×0 rect when measured).
+    const anchor = this.playerContainerElement ?? this.uiContainerElement;
+    if (!anchor) return;
+    const rect = anchor.getBoundingClientRect();
+    const rootEl = this.getDomElement().get(0) as HTMLElement;
+    // Pre-measure the menu so we can center it on the player rather than have its
+    // top-left at the player center.
+    this.reparentToOverlayHost(rootEl);
```
stonko1994: Same as before. I'm not sure reparenting is the correct approach here.

src/ts/components/PlayerContextMenu.ts
```diff
@@ -0,0 +1,444 @@
...
+}
+
+/**
+ * A floating context menu shown when the user right-clicks on the player.
+ * Displays Bitmovin info, the Player and UI versions, and an action button to
+ * toggle the {@link DebugInfoOverlay}.
+ *
+ * @category Components
+ */
+export class PlayerContextMenu extends Container<PlayerContextMenuConfig> {
```
stonko1994, line 37: The `PlayerContextMenu` does not have a dedicated way to close it. Any input onto the UI directly reacts to the input (e.g. clicking in empty space trigger playback (play/pause). Maybe we could re-use our `DismissClickOverlay` for this.

src/ts/components/PlayerContextMenu.ts
```diff
@@ -0,0 +1,444 @@
...
+}
+
+/**
+ * Builds a structured snapshot of the current playback state for the "Copy debug info"
+ * action. Captures everything a Bitmovin support engineer would normally have to ask the
+ * user to gather manually — player + UI version, source URL, current quality, buffer
+ * levels, dropped frames, parsed color space, current time, page URL, user agent.
+ */
+function buildDebugInfo(player: PlayerAPI): Record<string, unknown> {
+  const safe = <T>(fn: () => T): T | undefined => {
```
stonko1994: This feels wrong. Wrapping every player call in this `safe` feels like we don't know which of them are avaialble when. What was the reason for adding this?

src/ts/localization/languages/en.json
```diff
@@ -12,6 +12,7 @@
   "settings.subtitles.background.color": "Background color",
   "settings.subtitles.background.opacity": "Background opacity",
   "settings.subtitles.options": "Options",
+  "settings.playerInfo": "Player Info & Video Stats",
```
stonko1994: This is quite a long text. especially on mobile, it almost fills the full width. Also, opening the context menu on mobile on that button and then only opening the stats window feels like an unnecessary extra navigation step.

Image

src/ts/localization/languages/en.json
stonko1994, line 1: Any reason not to localize those new texts?

src/ts/UIFactory.ts
```diff
@@ -246,7 +250,10 @@ export namespace UIFactory {
 
     export function main(config: UIConfig = {}): UIContainer {
       const subtitleOverlay = new SubtitleOverlay();
+      const debugUiEnabled = !config.disableDebugUi;
+      const debugInfoOverlay = debugUiEnabled ? new DebugInfoOverlay({ hidden: true }) : null;
 
+      // Desktop layout — right-click already exposes the context menu, so no settings entry.
```
stonko1994: This can also be used on mobile. I think we should rather use the touch (`BrowserUtils.isMobile`) check.

src/ts/UIFactory.ts
```diff
@@ -359,10 +370,15 @@ export namespace UIFactory {
...
-    export function smallScreen(): UIContainer {
+    export function smallScreen(config: UIConfig = {}): UIContainer {
       const subtitleOverlay = new SubtitleOverlay();
+      const debugUiEnabled = !config.disableDebugUi;
+      const debugInfoOverlay = debugUiEnabled ? new DebugInfoOverlay({ hidden: true }) : null;
+      const playerContextMenu = debugUiEnabled && debugInfoOverlay ? new PlayerContextMenu({ debugInfoOverlay }) : null;
 
-      const settingsPanel = buildDefaultSettingsPanel(subtitleOverlay, -1);
+      // Touch / mobile / WebView — expose the context menu through a settings-panel row
+      // since there is no right-click. Omitted when the debug UI is disabled.
```
stonko1994: The smallScreen UI variant can also be used on web devices. I think we should rather use the touch (`BrowserUtils.isMobile`) check.

src/ts/UIFactory.ts
```diff
@@ -775,6 +796,40 @@ export namespace UIFactory {
...
 
+    // On platforms without a native context menu (touch / TV / WebView / set-top boxes /
+    // game consoles), expose the same actions through a settings-panel row that opens the
+    // `PlayerContextMenu` centered over the player. The desktop `main` layout omits this
+    // row because right-click already works there.
+    if (playerContextMenu) {
+      const playerInfoLabel = new Label<LabelConfig>({
+        text: i18n.getLocalizer('settings.playerInfo'),
+      });
+      const playerInfoItem = new SettingsPanelItem({
```
stonko1994: Can we add a separator view above this item? Similar to youtube to have a visible separation of video related settings and 'UI related' settings?

src/ts/UIFactory.ts
```diff
@@ -775,6 +796,40 @@ export namespace UIFactory {
...
+      };
+      const itemEl = playerInfoItem.getDomElement();
+      itemEl.css('cursor', 'pointer');
+      itemEl.on('click', openMenu);
+      itemEl.on('keydown', (e: KeyboardEvent) => {
+        if (e.key === 'Enter' || e.key === ' ') {
+          e.preventDefault();
+          openMenu();
+        }
+      });
```
stonko1994: Please don't include business logic (even if just clicking) in the `UIFactory` code. If a customer wants to build a custom layout they would need to be aware of that and also includ this in their layout code.

For this case I'm afraid the target component needs to know the settings panel to close it from there. See other settings realted components following the same pattern.

CHANGELOG.md
```diff
@@ -9,7 +9,35 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- New `UIConfig.cea608SmallPlayerHeightThreshold` option (default `360`) to configure the rendered player height threshold at or below which small-player CEA-608 caption adjustments are applied
```
stonko1994, line 12: We don't touch existing changelog entries and just add new entries. This helps with merge conflicts if other PRs are merged in the meantime.

CHANGELOG.md
```diff
@@ -9,7 +9,35 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+- `DebugInfoOverlay`: a small "stats for nerds"-style overlay that shows live playback diagnostics. Hidden by default; opens via the new context menu. Lines rendered (when the source exposes them):
```
stonko1994: I'd prefer to write the top level changelog entry in a more general way:

```suggestion
- A new `DebugInfoOverlay` component to display a small "stats for nerds"-style overlay that shows live playback diagnostics. Hidden by default; opens via the new context menu. Lines rendered (when the source exposes them):
```

stonko1994: We usually don't include this level of details in the changelog. The information that unavaiable values are not rendered is a component detail not really relevant here.

```suggestion
- `DebugInfoOverlay`: a small "stats for nerds"-style overlay that shows live playback diagnostics. Hidden by default; opens via the new context menu. Supported insights:
```

CHANGELOG.md
```diff
@@ -9,7 +9,35 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+  - Playback / downloaded video and audio bitrate + codec (e.g. `Video: 1920×818@24 6.21 Mbps (avc1.4d4028)` with a `↓` line for the downloaded variant when ABR has switched)
+  - Decoded resolution → rendered element size (catches up/down-scaling)
+  - Color primaries / transfer characteristics parsed from AV1 / VP9 codec strings per ITU-T H.273 (omitted on AVC / HEVC, which don't encode color)
```
stonko1994: I think `Color primaries` is super misleading here. If someone isn't familliar with the implementation one could think what is the relation to colors here.

CHANGELOG.md
```diff
@@ -9,7 +9,35 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+  - Video and audio buffer length in seconds
+  - Dropped frames
```
stonko1994: Do we need to indicate Web only features?

```suggestion
  - Dropped frames (Web player only)
```

CHANGELOG.md
```diff
@@ -9,7 +9,35 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+  - Current time, duration, playback speed
+  - Live latency (`X.XXs behind edge`) on live sources
+  - Available quality counts and codec families (e.g. `Video codecs: avc1, hevc, av01`)
+  - DRM systems, manifest URL (truncated), stream type + player type, player version
+  - Refreshes every 1 s while playing plus on quality / seek / source events; the timer pauses while the overlay is hidden.
+  - Survives the controls auto-hide (reparented to the player container, not the UI container, so it stays visible when the rest of the UI fades out).
+  - Text is selectable and selections survive refresh ticks (writes are skipped while a selection is active inside the overlay).
```
stonko1994: This is not true on iOS. We disabled text selection.

CHANGELOG.md
```diff
@@ -9,7 +9,35 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+  - Decoded resolution → rendered element size (catches up/down-scaling)
+  - Color primaries / transfer characteristics parsed from AV1 / VP9 codec strings per ITU-T H.273 (omitted on AVC / HEVC, which don't encode color)
+  - Video and audio buffer length in seconds
+  - Dropped frames
+  - Current time, duration, playback speed
+  - Live latency (`X.XXs behind edge`) on live sources
+  - Available quality counts and codec families (e.g. `Video codecs: avc1, hevc, av01`)
+  - DRM systems, manifest URL (truncated), stream type + player type, player version
+  - Refreshes every 1 s while playing plus on quality / seek / source events; the timer pauses while the overlay is hidden.
+  - Survives the controls auto-hide (reparented to the player container, not the UI container, so it stays visible when the rest of the UI fades out).
```
stonko1994: Hopefully 😅 I wouldn't include this information at all.

CHANGELOG.md
```diff
@@ -9,7 +9,35 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+  - Localized strings refresh when the UI language changes.
+- "Player Info & Video Stats" settings-panel row added to the small-screen layout (touch / mobile / WebView) — opens the same context menu centered over the player, since right-click isn't reachable without a mouse. Full-width tappable row with hover / focus styling, `role="menuitem"`, `tabindex=0`, and keyboard `Enter` / `Space` activation.
+- `UIConfig.disableDebugUi`: opt-out flag that removes the entire debug surface (right-click `PlayerContextMenu`, `DebugInfoOverlay`, and the "Player Info & Video Stats" settings-panel row) from the default and small-screen layouts. Use this in production deployments where the debug surface should not be exposed to end users. Default `false`.
+- New localization keys (added to `en.json`; non-English vocabularies fall back to English via the existing `I18n#initializeVocabulary` merge): `videoStats.title`, `videoStats.show`, `videoStats.hide`, `contextMenu.title`, `contextMenu.subtitle`, `contextMenu.about`, `contextMenu.copySource`, `contextMenu.copyConfig`, `contextMenu.copyDebugInfo`, `contextMenu.copied`, `settings.playerInfo`.
```
stonko1994: This is to detailed. Please remove this entry

CHANGELOG.md
```diff
@@ -9,7 +9,35 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+  - Bitmovin Player heading + subtitle, Player and UI versions, "About Bitmovin" link.
+  - "Show / Hide video stats" toggle (label flips with the overlay state).
+  - "Copy source" — current `player.getSource()` as JSON.
+  - "Copy player config" — current `player.getConfig()` as JSON.
+  - "Copy debug info" — a single JSON snapshot (timestamp, page URL, user agent, player + UI version, source URLs + DRM systems, full playback state, current quality with parsed color, audio, available codecs, track / quality counts) for sharing in support tickets.
+  - Clipboard writes use `navigator.clipboard.writeText` with a textarea + `execCommand('copy')` fallback for insecure-context / unfocused-document cases.
+  - `role="menu"` on the container, `role="menuitem"` on each action button.
+  - Reparented into `document.fullscreenElement` when present, otherwise `<body>`, so the menu stays visible in native fullscreen.
+  - Right-clicking on the `<video>` element itself is skipped so the browser's native video context menu (Save video as, Picture-in-Picture, …) keeps working.
+  - Localized strings refresh when the UI language changes.
```
stonko1994: This is too detailed. Please remove them.

lukaskroepfl:
@stonko1994 ready for another look — all 27 inline threads addressed, and the manual-testing observations handled where possible:

## Inline review threads

**DebugInfoOverlay**
- Dropped `@include hidden-animated` (overlay no longer animates; `display: none` when hidden) and the configure-time reparenting to the player container.
- `window.setInterval` → `Timeout` util.
- `touch-action: none` on the header so iOS Safari doesn't claim drag gestures as swipe-back.
- Renamed `next` → `playerStatsHtml`.
- Dropped the parsed Color primaries line + the codec parser ("player responsibility").
- Suppress lines whose underlying value isn't available — no more `Stream: unknown (unknown)` on iOS native HLS, no `Dropped frames: 0` line when unsupported.
- The playback/downloaded getters go through a small `tryGet()` helper since they actually throw on muxed HLS / before-source-loaded (the source of the HLS console exception from manual testing).

**PlayerContextMenu**
- Reuses the existing `Button` component for the four action buttons instead of raw `<button>` DOM.
- Imports `version` from `main.ts` (no duplicate `{{VERSION}}` template).
- Dropped the runtime reparenting to `<body>` / fullscreen element; menu stays inside the UIContainer which is inside the player container (the fullscreen target), so visibility in fullscreen still works.
- Outside-click dismissal via a sibling `DismissClickOverlay` (same pattern as `SettingsPanel`); Escape stays as a direct hookup.
- Dropped the broad `safe(...)` wrapper from `buildDebugInfo` — only the four playback/downloaded-data getters that actually throw are now wrapped via `tryGet()`.

**UIFactory**
- `BrowserUtils.isMobile` (not the layout choice) gates the Player Info settings-panel entry — both `main` and `smallScreen` use it.
- Separator above the Player Info entry.
- `PlayerContextMenu.createSettingsPanelOpenerItem(parentSettingsPanel)` encapsulates the click-handler so UIFactory doesn't carry business logic.

**Localization / CHANGELOG**
- `settings.playerInfo` shortened to "Player info".
- CHANGELOG trimmed to ~5 concise entries per the feedback.

## Manual-testing observations

- **HLS console exception** (#1): fixed via `tryGet()` around the muxed-HLS-prone getters.
- **iOS unknowns / missing fields** (#2): suppress the lines instead of showing `unknown` or `0`.
- **Copy source/config empty on mobile** (#2.1): same root cause; will dump whatever's available on iOS native HLS.
- **UI/UX matches settings panel** (#3.1): translucent backdrop, blur, soft shadow on both the overlay and context menu.
- **Scrolling** (#3.2): the menu now uses position relative to the UIContainer (no body reparenting) so it stays anchored.
- **About Bitmovin link** (#3.3): removed.
- **`Resolution → Resolution` unclear** (#3.5): split into separate `Decoded:` and `Rendered:` lines.

The only manual-testing item I left for a follow-up is the **"More" subpage** on mobile (#3.4) — that's a bigger structural change.

## Manual smoke-test in Chrome DevTools

- Right-click → context menu opens; click outside / Esc / DismissClickOverlay all dismiss; `Show video stats` toggle opens the overlay.
- Overlay has the close button (×), Esc closes it, drag from header works.
- On emulated iPhone: settings panel shows the "Player info" row with separator; tapping it opens the menu centered.
- HLS source loaded — no console exceptions, fields populate.

Last commit: `aba9726b`.

Issue #111, opened 2018-06-14
Proposed features and refactorings for v3+


protyposis:
A list of thoughts and ideas for improvements in upcoming major UI releases.

### Features
 - Multilanguage support
 - Keep hidden elements out of DOM tree instead of just hiding them via CSS
 - Support runtime configuration of components
 - Allow configuration of components through the UIManager, i.e. to allow the player to pass a config object to the "built-in" UI

### Refactorings
 - Don't fire a fake `ON_READY` event from the UIManager when the UI is initialized, rather add a UI `onReady` event that is triggered when the UI is initialized and when `ON_READY` events from the player are received (the goal should be to get rid of `uiSourced` events and the necessity to differentiate them from "real" player events)

protyposis:
Added ES6 exports fix

PR #868, opened 2026-05-08
a11y: fix SettingsToggleButton ARIA and honor prefers-reduced-motion

lukaskroepfl:
## Summary

Two small accessibility fixes:

### 1. Fix invalid ARIA role on `SettingsToggleButton`

`SettingsToggleButton.ts` was setting `role="pop-up button"` on its `<button>` element. That value is not in the ARIA role registry — screen readers fell back to the default (`button`) and the popup relationship wasn't announced reliably.

The fix is to:

- Drop the custom `role` so the element keeps its implicit native `button` role.
- Set `aria-haspopup="menu"` (was `"true"` — `"menu"` is more specific and matches what the panel actually is).
- Add `aria-controls` pointing at the active panel-page id, refreshed on `onActivePageChanged` so it stays correct as the user navigates sub-pages.
- Add `aria-expanded`, initialized from the panel's current visibility and flipped on `onShow` / `onHide`.
- **Drop `aria-owns`**: when it points at the same element as `aria-controls`, iOS VoiceOver follows both relationships and announces the menu twice. The WAI-ARIA menu-button pattern uses `aria-controls` alone.
- Suppress the inherited `aria-pressed` (via matching on/off aria labels) so the button doesn't expose both `aria-pressed` and `aria-expanded`.

### 2. Honor `prefers-reduced-motion: reduce`

The UI had no `prefers-reduced-motion` support — the control bar fade, settings panel slide, buffering loader, huge play button "breathe" animation, etc. all ignored the user's OS preference.

Added a single global rule in `src/scss/bitmovinplayer-ui.scss` that disables every animation and transition inside `.bmpui-ui-uicontainer` (`animation: none` / `transition: none`) when the user has set `prefers-reduced-motion: reduce`. Disabling outright (rather than collapsing durations) avoids a glitch where `transition: visibility 0s <duration>` kept the settings panel on a discrete visibility delay and popped on open. `SubtitleOverlay.awaitTransitionEnd` already short-circuits when `transition-property` is `none`, so no listeners are stranded.

## Test plan

- [x] All 508 unit tests pass; ESLint and `tsc --noEmit` clean (ARIA wiring covered by 5 tests).
- [x] Open the demo player → inspect the settings cog → `aria-haspopup="menu"`, `aria-controls=<page-id>`, `aria-expanded="false"`, no `aria-owns`, no `aria-pressed`. *(verified in-browser)*
- [x] Open the settings panel → `aria-expanded` flips to `"true"`; close → flips back to `"false"`. *(open verified in-browser; close covered by unit test)*
- [x] Under `prefers-reduced-motion: reduce`, the settings panel opens with no glitch — computed `transition: none`, panel `visibility: visible` on the first frame. *(verified in-browser with the reduced-motion rule applied)*
- [x] Screen-reader-facing ARIA semantics verified structurally (native `button` role + `aria-haspopup="menu"` + `aria-expanded` collapsed/expanded). *(full VoiceOver/TalkBack pass not run in this environment)*

## Notes

I did not change color contrast (`#999` on `#111`) — that's a tokenized design call that touches the whole UI palette, worth its own PR with a designer in the loop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Support for the `prefers-reduced-motion` accessibility setting: all UI animations and transitions are collapsed to ~1 ms when the user has set the OS-level `prefers-reduced-motion: reduce` preference (animation / transition end events still fire so dependent UI logic continues to work)
+- Support for the `prefers-reduced-motion` accessibility setting: UI animations and transitions are disabled when the user has set the OS-level `prefers-reduced-motion: reduce` preference
 - New `UIConfig.cea608SmallPlayerHeightThreshold` option (default `360`) to configure the rendered player height threshold at or below which small-player CEA-608 caption adjustments are applied
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,16 +9,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Support for the `prefers-reduced-motion` accessibility setting: all UI animations and transitions are collapsed to ~1 ms when the user has set the OS-level `prefers-reduced-motion: reduce` preference (animation / transition end events still fire so dependent UI logic continues to work)
 - New `UIConfig.cea608SmallPlayerHeightThreshold` option (default `360`) to configure the rendered player height threshold at or below which small-player CEA-608 caption adjustments are applied
 
 ### Changed
 
 - CEA-608 captions on players with a rendered height at or below 360 CSS pixels (configurable via `cea608SmallPlayerHeightThreshold`) now stay readable by using a centered 80% safe area and avoiding control-bar pushup that can crowd caption rows on small players
-- The settings toggle button now exposes valid ARIA semantics for assistive technologies: it relies on the implicit `<button>` role (the previous custom `role="pop-up button"` is not a valid ARIA role), advertises the popup via `aria-haspopup="menu"`, links the panel via `aria-controls`, and reflects open / closed state in `aria-expanded`
-- All UI animations and transitions are collapsed to ~1 ms when the user has set the OS-level `prefers-reduced-motion: reduce` preference (animation / transition end events still fire so dependent UI logic continues to work)
 
 ### Fixed
 
+- Wrong ARIA semantics on the `SettingsToggleButton`: it now relies on the native `<button>` role (the previous custom `role="pop-up button"` is not a valid ARIA role), advertises the popup via `aria-haspopup="menu"`, links the panel via `aria-controls`, and reflects open / closed state in `aria-expanded`
 - CEA-608 captions could keep stale sizing after CEA subtitle rendering was disabled and re-enabled, causing captions to appear incorrectly scaled or positioned
 
 ## [4.13.0] - 2026-05-07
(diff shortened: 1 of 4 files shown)


lukaskroepfl:
Addressed the Copilot review pass in 4e898889 — summary:

- **`role` comment** → updated wording: the native `<button>` keeps the explicit `role="button"` from the Button base, not an implicit one.
- **`aria-controls` / `aria-owns` go stale on sub-page navigation** → now refreshed in `configure()` by subscribing to `settingsPanel.onActivePageChanged`. Added a test for it.
- **`aria-pressed` + `aria-expanded` redundancy** → suppressed `aria-pressed` by giving the parent `ToggleButton` matching `onAriaLabel` / `offAriaLabel` ("Settings"). The expanded state is the load-bearing one for a popup button.
- **`aria-expanded` initial sync** → set in `configure()` from `settingsPanel.isShown()` so an already-open panel doesn't desync the attribute.
- **Reduced-motion missed `*-delay`** → added `animation-delay: 0ms !important` and `transition-delay: 0ms !important` to the rule. Composite `transition` shorthand and the buffering overlay's staggered `animation-delay` are now covered.
- **Test reached into private `componentEvents`** → refactored to use the public `panel.show()` / `panel.hide()` API.

507 / 507 tests pass; ESLint + tsc clean.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - CEA-608 captions on players with a rendered height at or below 360 CSS pixels (configurable via `cea608SmallPlayerHeightThreshold`) now stay readable by using a centered 80% safe area and avoiding control-bar pushup that can crowd caption rows on small players
+- The settings toggle button now exposes valid ARIA semantics for assistive technologies: it relies on the implicit `<button>` role (the previous custom `role="pop-up button"` is not a valid ARIA role), advertises the popup via `aria-haspopup="menu"`, links the panel via `aria-controls`, and reflects open / closed state in `aria-expanded`
```
stonko1994: I think this should be a `### Fixed` entry as this is fixing an issue.


```suggestion
### Fixed
- Wrong ARIA semantics of the `SettingsToggleButton`
```

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+- All UI animations and transitions are collapsed to ~1 ms when the user has set the OS-level `prefers-reduced-motion: reduce` preference (animation / transition end events still fire so dependent UI logic continues to work)
```
stonko1994: IMO This should be an `### Added` entry as we never had support for that.


```suggestion
### Added
- Support for `prefers-reduced-motion` accessibility setting.
```

stonko1994 (commented):
Looks like the reduced motion creates a glitch in the settings panel opening.

https://github.com/user-attachments/assets/629e4549-a105-412b-9559-854750362ee4

---

The VoiceOver reader on iOS read the new settings panel aria twice.

stonko1994 (changes requested):
The reported glitches from the last review are still present.

lukaskroepfl:
The glitches from the last review were fixed in `ba4b18db` (pushed after your May 13 review), so this should be ready for another look:

- **Panel-open glitch**: the reduced-motion rule now sets `transition: none` / `animation: none` outright instead of collapsing durations. The old glitch came from `transition: visibility 0s <duration>` keeping `visibility` on a discrete delay; with `transition: none` the panel flips visible immediately. Verified in-browser under simulated `prefers-reduced-motion: reduce` — panel is `visibility: visible; opacity: 1` on the first frame, no pop.
- **iOS VoiceOver double-read**: dropped `aria-owns` (kept `aria-controls` only), matching the WAI-ARIA menu-button pattern.
- Also addressed the Copilot points (stale `aria-controls` on page nav, initial `aria-expanded` sync, `aria-pressed`/`aria-expanded` clash, brittle test) and your CHANGELOG categorization.

All 5 `SettingsToggleButton` ARIA unit tests pass.

stonko1994 (approved):

PR #869, opened 2026-05-08
Add frame stepping, resume, persistent prefs, timestamp share link

lukaskroepfl:
## Summary

Four new player-UI features, all gated by `UIConfig` flags (default on) and respectful of constrained runtimes (old TVs, Android / iOS WebViews, RDK / set-top boxes, game consoles).

> **Base note:** this PR targets `feature/video-stats-debug-overlay` so reviewers see only the feature delta. It will retarget `develop` once #867 lands.

### `FrameStepHandler`

- Keyboard-only frame-by-frame stepping. `,` steps one frame back, `.` steps one frame forward; pauses playback first if needed.
- Frame duration is read from `getPlaybackVideoData().frameRate`, with a 30 fps fallback for sources that don't expose it.
- Excludes form fields and modifier-key combinations so it never intercepts user input.
- TV remotes don't generally surface `,`/`.`, so this is a desktop / attached-keyboard feature; harmless on TV / WebView / console targets.

### Shareable timestamp link

- New "Copy link at current time" button in `PlayerContextMenu`.
- Builds `<href>?t=<seconds>s` using plain string operations (no `URL` constructor) for compatibility with the oldest TV / STB / console browsers.
- Preserves existing query params and the URL fragment, replaces an existing `t=` rather than duplicating.

### `ResumeOverlay`

- Banner that offers to resume playback at the last saved position when a known source is reloaded.
- Source identity hashed from `metadata.title` (preferred) or the manifest URL via a small djb2 hash. Sources without identifying metadata are not tracked.
- Skips entries within 5 s of start or 30 s of end; expires entries after 90 days; clears on `PlaybackFinished`.
- Saves on a 5 s interval while playing, plus on `Paused`, `SourceUnloaded`, and `beforeunload` (best-effort).

### `UIPreferences`

- Persists volume, mute state, and playback speed across sessions; reapplies them after `Ready`.
- All writes route through `StorageUtils` which already swallows storage errors, so on devices where `localStorage` is unavailable or restricted (private browsing, RDK without persistent storage, restricted WebViews) all four features degrade to no-ops without throwing.

### Config flags (all default `true`)

| Flag                                 | Effect                                                     |
| ------------------------------------ | ---------------------------------------------------------- |
| `enableFrameStepping`                | Adds the `,` / `.` keyboard handler.                       |
| `enableResumeFromLastPosition`       | Adds the `ResumeOverlay` and tracks playback position.    |
| `enablePersistentPreferences`        | Restores volume / muted / speed on `Ready`.                |

The existing `disableStorageApi` continues to globally gate all storage-backed features.

## Test plan

- [x] All 525 unit tests pass; ESLint and `tsc --noEmit` are clean (13 new tests).
- [ ] Press `,` / `.` while a non-live source plays → playback pauses, position steps by ~1/30 s. Hold `Shift+,` → ignored. Type `,` in a `<input>` overlay (if any) → not intercepted.
- [ ] Right-click → "Copy link at current time" → label flips to "Copied!"; clipboard contains `<page-url>?t=<seconds>s`.
- [ ] Reload the page with `?t=120s` → integrator-side seek logic still works (the UI no longer touches that param except to write it).
- [ ] Watch ~30 s of a source with a `metadata.title`, refresh → resume banner appears with the saved time. Click "Resume" → seeks to it. Click "Start over" → seeks to 0 and clears the entry.
- [ ] Set volume to 50 %, change speed to 1.5×, mute. Reload. Volume / speed / mute all restored after `Ready`.
- [ ] Disable `localStorage` (private mode or `disableStorageApi: true`) → all four features quietly do nothing; no console errors.


lukaskroepfl:
Superseded — split into four individual PRs for easier review:

- #870 — Frame-by-frame stepping (`,` / `.` keyboard) → base `develop`
- #871 — Resume from last position overlay → base `develop`
- #872 — Persist volume / mute / playback speed → base `develop`
- #873 — Shareable timestamp links (copy + deep-link consumer) → base `feature/video-stats-debug-overlay` (#867)

Closing this PR.

PR #870, opened 2026-05-11
Add frame-by-frame stepping (keyboard)

lukaskroepfl:
## Summary

Adds frame-by-frame stepping to the seekbar:

- **`,`** steps one frame back
- **`.`** steps one frame forward

Folded into `SeekBarController` so it follows the same focus-gated pattern as the existing arrow / Home / End keys — the keys only fire while the seekbar element is focused. Playback is paused first if it isn't already. Frame duration is taken from `getPlaybackVideoData().frameRate` when exposed by the player, falling back to **30 fps**.

**VOD only.** Frame stepping uses `seek()` on VOD. Live streams ignore the keys: `timeShift` is measured against the constantly advancing live edge and `getTimeShift()` is not a stable reference while paused, so a sub-frame step gets swamped by edge movement and the position drifts forward instead of stepping. Frame-accurate live stepping needs absolute-time seeking on the player side (follow-up). The Volume seekbar is also excluded, where frame stepping has no meaning.

No new config flag — focus gating makes the behavior inert unless the user has tabbed/clicked into the seekbar, so it doesn't intercept user typing elsewhere.

Split out from #869.

## Test plan

- [x] 503 unit tests pass; ESLint and `tsc --noEmit` clean.
- [x] In the demo, focus the seekbar → press `.` repeatedly. Playback pauses, frame advances by ~1/fps each press. *(verified in-browser, Sintel VOD @ 24fps)*
- [x] Press `,` → steps backward by one frame. *(verified in-browser; forward-then-back returns to start)*
- [x] On a live stream, `,` / `.` are ignored (frame stepping is VOD-only). *(verified in-browser against a DVR stream — no-op)*
- [x] Focus elsewhere (text input, button) → `,` / `.` are not intercepted. *(keydown listener is bound only to the focused seekbar element)*

### DIFF
--- a/src/ts/utils/SeekBarController.ts
+++ b/src/ts/utils/SeekBarController.ts
@@ -81,7 +81,8 @@ export class SeekBarController {
 
   public setSeekBarControls(domElement: DOM, type: () => SeekBarType) {
     domElement.on('keydown', (e: KeyboardEvent) => {
-      const controls = this.seekBarControls(type());
+      const seekBarType = type();
+      const controls = this.seekBarControls(seekBarType);
       switch (e.keyCode) {
         case UIUtils.KeyCode.LeftArrow: {
           controls.left();
@@ -118,7 +119,38 @@ export class SeekBarController {
           e.preventDefault();
           break;
         }
+        case UIUtils.KeyCode.Comma: {
+          if (seekBarType === SeekBarType.Vod) {
+            this.stepFrame(-1);
+            e.preventDefault();
+          }
+          break;
+        }
+        case UIUtils.KeyCode.Period: {
+          if (seekBarType === SeekBarType.Vod) {
+            this.stepFrame(1);
+            e.preventDefault();
+          }
+          break;
+        }
       }
     });
+  }
+
+  /**
+   * Steps the playback position by one frame in the given direction (-1 = back, 1 = forward).
+   * Pauses the player first if necessary so the step lands on a stable frame. The frame
+   * duration is derived from the active video quality's `frameRate` and falls back to
+   * 30 fps when the player doesn't expose one (e.g. progressive sources).
+   */
+  protected stepFrame(direction: number): void {
+    if (!this.player.isPaused()) {
+      this.player.pause('ui');
+    }
+    const videoData = this.player.getPlaybackVideoData() as { frameRate?: number } | null | undefined;
+    const fps =
+      videoData && typeof videoData.frameRate === 'number' && videoData.frameRate > 0 ? videoData.frameRate : 30;
+    const target = Math.max(0, this.player.getCurrentTime() + direction * (1 / fps));
+    this.player.seek(target);
   }
 }
--- a/src/ts/utils/SeekBarController.ts
+++ b/src/ts/utils/SeekBarController.ts
@@ -120,14 +120,14 @@ export class SeekBarController {
           break;
         }
         case UIUtils.KeyCode.Comma: {
-          if (seekBarType === SeekBarType.Vod) {
+          if (seekBarType !== SeekBarType.Volume) {
             this.stepFrame(-1);
             e.preventDefault();
           }
           break;
         }
         case UIUtils.KeyCode.Period: {
-          if (seekBarType === SeekBarType.Vod) {
+          if (seekBarType !== SeekBarType.Volume) {
             this.stepFrame(1);
             e.preventDefault();
           }
@@ -150,7 +150,14 @@ export class SeekBarController {
     const videoData = this.player.getPlaybackVideoData() as { frameRate?: number } | null | undefined;
     const fps =
       videoData && typeof videoData.frameRate === 'number' && videoData.frameRate > 0 ? videoData.frameRate : 30;
-    const target = Math.max(0, this.player.getCurrentTime() + direction * (1 / fps));
-    this.player.seek(target);
+    const delta = direction * (1 / fps);
+    if (this.player.isLive()) {
+      const target = this.player.getTimeShift() + delta;
+      const clamped = Math.max(this.player.getMaxTimeShift(), Math.min(0, target));
+      this.player.timeShift(clamped);
+    } else {
+      const target = Math.max(0, this.player.getCurrentTime() + delta);
+      this.player.seek(target);
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


src/ts/UIConfig.ts
```diff
@@ -159,6 +159,15 @@ export interface UIConfig {
...
   disableStorageApi?: boolean;
+  /**
+   * If set to true, keyboard frame-by-frame stepping is enabled. Pressing `,` steps one
+   * frame back, `.` steps one frame forward. Pauses playback first if needed.
+   * Frame duration is derived from the active video quality's frame rate, falling back
+   * to 30 fps when not exposed by the player.
+   *
+   * Default: true
+   */
+  enableFrameStepping?: boolean;
```
stonko1994: We already have `keyStepIncrements` on the SeekBar. IMO we should try to be consistent and also move `enableFrameStepping` there and only enable it when the seekbar is focused. This is how the `keyStepIncrements` works as well.

src/ts/components/FrameStepHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+ * frame.
+ *
+ * The component is keyboard-only by design and is meaningful primarily on devices with a
+ * physical keyboard (desktop, attached BT keyboards). It does not register any TV remote
+ * mappings — most TV remotes don't expose `,` / `.` codes — but it is harmless on those
+ * devices. Form fields are excluded so typing in inputs / textareas isn't intercepted.
+ *
+ * @category Components
+ */
+export class FrameStepHandler extends Component<FrameStepHandlerConfig> {
```
stonko1994: Please do not use the `Component` for elements without any UI. This should be a helper or live on a component.

We already have similar features on the `Seekbar`. IMO we should put it there.

stonko1994 (changes requested):

lukaskroepfl:
Restructured per @stonko1994's feedback (c86701e0):

- Dropped `FrameStepHandler` — no more `Component` for an invisible service.
- Logic moved into `SeekBarController.stepFrame()` alongside the arrow / Home / End handlers, so the keys are naturally gated on seekbar focus (same pattern as `keyStepIncrements`).
- Added `Comma` / `Period` to `UIUtils.KeyCode`. They only fire for VOD; live seekbars ignore them.
- Dropped the `UIConfig.enableFrameStepping` flag and the UIFactory wiring — the feature now ships with the seekbar itself.

503 / 503 tests still pass.

src/ts/utils/SeekBarController.ts
```diff
@@ -118,7 +119,38 @@ export class SeekBarController {
           e.preventDefault();
           break;
         }
+        case UIUtils.KeyCode.Comma: {
+          if (seekBarType === SeekBarType.Vod) {
```
stonko1994: In the UI we follow the pattern that the player is the source of truth. Please don't use the seekBarType for detecting a live stream.

If changing source types (live to VOD) is a concern please use the `LiveStreamDetector`. Otherwise a `player.isLive'` check should be enough.

That said, why don't we support live? if we are paused we should also be able to do frame accurate seeking in live no?

stonko1994 (commented):
Your latest changes removed the config options? was that intentional?

src/ts/utils/SeekBarController.ts
lukaskroepfl: Good call on both — fixed in 086fc1aa.

- Dropped the `SeekBarType.Vod` check; `stepFrame` now branches on `player.isLive()` directly. Only the Volume seekbar stays excluded (frame stepping has no meaning there).
- Added live support: when live, the step is applied via `timeShift()` clamped to `[getMaxTimeShift(), 0]`. VOD path keeps using `seek()`.

lukaskroepfl:
Re: removing the config options — yes, intentional. Once the keys are folded into `SeekBarController` they inherit the same focus-gated pattern as the existing arrow / Home / End keys: inert unless the seekbar has focus. That makes the explicit opt-out flag redundant — there's no scenario where `,` / `.` would intercept user typing or other intended input. Updated the PR description to reflect this.

src/ts/utils/SeekBarController.ts
stonko1994, line 118: Looking at the new code it made me wonder whether this is intended to toggle playback on the volume bar when hitting space 😄

Fine for me to ignore this for now as this was nothing new.

src/ts/utils/SeekBarController.ts
```diff
@@ -118,7 +119,45 @@ export class SeekBarController {
...
+  protected stepFrame(direction: number): void {
+    if (!this.player.isPaused()) {
+      this.player.pause('ui');
+    }
+    const videoData = this.player.getPlaybackVideoData() as { frameRate?: number } | null | undefined;
+    const fps =
+      videoData && typeof videoData.frameRate === 'number' && videoData.frameRate > 0 ? videoData.frameRate : 30;
+    const delta = direction * (1 / fps);
+    if (this.player.isLive()) {
+      const target = this.player.getTimeShift() + delta;
```
stonko1994: Frame stepping in live doesn't work.

I think we can't use `getTimeShift` here as timeShift is not a consistent value and progresses while the player is paused. I think we need to build the calculation based on `player.currentTime`. This and the 'seekable range' could be used to calculate a proper `timeShift` value. Ideally the player would accept absolute time seeking which would make this a lot easier from the UI perspective.

stonko1994 (changes requested):
The live use case doesn't work. It basically always steps forward. See comments for more details.

lukaskroepfl:
Pushed `5e91636` — **restricting frame stepping to VOD** and dropping the live branch.

I dug into the live behaviour against a real DVR stream and the "always steps forward" you saw is a genuine limitation of stepping via `timeShift`, not a fixable formula bug:

- While paused on live, `getCurrentTime()` is the only stable, frame-accurate anchor. `getTimeShift()` drifts / goes stale, and `getSeekableRange()` returns `{ start: -1, end: -1 }`.
- `timeShift(v)` parks `v` seconds behind the **live edge**, which keeps advancing — so re-applying an offset derived from `getTimeShift()` re-anchors the frame to the moved edge, and the sub-frame (±1/fps) delta gets swamped → the position creeps forward in both directions.
- Because `edge = getCurrentTime() − getTimeShift()`, every "anchor on `currentTime` + seekable range" rewrite reduces algebraically to exactly the current `getTimeShift() + delta`, so it can't fix it from the UI. This matches your own note — it really needs absolute-time seeking on the player side.

So live now ignores `,`/`.` and VOD keeps frame-accurate `seek()`. Verified in-browser: VOD steps exactly one frame per press (paused, both directions); live is a no-op. Happy to open a follow-up ticket for live stepping once/if the player exposes absolute-time seeking.

stonko1994 (approved):

PR #871, opened 2026-05-11
Add resume from last position

lukaskroepfl:
## Description
Adds opt-in resume-from-last-position support via `UIConfig.enableResumeFromLastPosition`.

When enabled, the UI derives a per-source `localStorage` key from the loaded source title or URL, tracks the latest playback position from `TimeChanged`, stores it through the existing `StorageUtils` on pause/source unload/beforeunload, and automatically seeks to that saved position when the same source is loaded again.

Current behavior:
- Disabled by default.
- Live sources and sources without a stable identifier are not tracked.
- Saved positions below 5 seconds are cleared or ignored.
- `PlaybackFinished` clears the saved position.
- Valid shareable timestamp links (`?t=` or `#t=`) take precedence over resume.

- [x] `CHANGELOG` entry


stonko1994 (commented):
This doesn't seem to work for me. I don't get any banner after reloading the same source.

lukaskroepfl:
~@stonko1994 that was because the test page cleared the storage on every load. added a config to top~

lukaskroepfl:
Thanks for testing! This turned out to be the demo page tripping you up, not the feature. `src/html/index.html` calls `localStorage.clear()` on every page load (in the `buildUI` branch), which wipes the saved resume position before the banner can read it — so you'd never see a banner after a reload. The checkbox I added defaults to ON (clearing), so it still bites by default; uncheck "clear localStorage on load" (takes effect next reload) and the banner shows.

Verified end-to-end in a browser with the Sintel source (clear disabled):
- Play past ~5s → reload → **"Resume at 00:02:15"** banner appears.
- **Resume** seeks to the saved time and plays.
- **Start over** resets to 0 and clears the stored entry.

tsc / eslint / jest all green (510 tests, incl. `ResumeStorage.spec.ts`). The feature code is unchanged.

If we want to spare future reviewers this confusion, I'm happy to flip that demo checkbox to default OFF (or have it clear only the player's own keys instead of all of `localStorage`) — let me know and I'll do it in a follow-up.

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,138 @@
...
+      this.lastPosition = null;
+      return;
+    }
+
+    if (!this.activeSourceKey) return;
+
+    if (!isFinite(time) || time < MIN_RESUME_POSITION) {
+      StorageUtils.removeItem(this.activeSourceKey);
+      return;
+    }
```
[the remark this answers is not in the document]
stonko1994: Addressed in [d195dfca](https://github.com/bitmovin/bitmovin-player-ui/commit/d195dfca9b87c4913673bc087a251211e2ece915).

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,138 @@
...
+    this.refreshKey();
+    this.lastPosition = null;
+    if (!this.activeSourceKey) return;
+
+    const time = Number(StorageUtils.getItem(this.activeSourceKey));
+    if (!isFinite(time) || time < MIN_RESUME_POSITION) return;
+
+    this.lastPosition = time;
+    this.player.seek(time, 'ui');
+  };
```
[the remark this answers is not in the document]
stonko1994: Addressed by updating the PR description to match the current no-overlay auto-resume behavior. Related implementation is in [e8091535](https://github.com/bitmovin/bitmovin-player-ui/commit/e80915358796fd231865c140d762c336a8417612).

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,138 @@
...
+  private activeSourceKey: string | null = null;
+  private lastPosition: number | null = null;
+
+  constructor(private readonly player: PlayerAPI) {
+    player.on(player.exports.PlayerEvent.SourceLoaded, this.startPositionTracking);
+    player.on(player.exports.PlayerEvent.SourceUnloaded, this.unloadSource);
+    player.on(player.exports.PlayerEvent.TimeChanged, this.updatePosition);
+    player.on(player.exports.PlayerEvent.Paused, this.pausePositionTracking);
+    player.on(player.exports.PlayerEvent.PlaybackFinished, this.finishPositionTracking);
+    window.addEventListener('beforeunload', this.storeCurrentPosition);
```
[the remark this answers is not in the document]
stonko1994: Addressed by updating the PR description to remove the stale 5s interval, TTL, and near-end policy claims. Related implementation is in [e8091535](https://github.com/bitmovin/bitmovin-player-ui/commit/e80915358796fd231865c140d762c336a8417612).

src/ts/UIManager.ts
```diff
@@ -602,6 +605,10 @@ export class UIManager {
 
     // Initialize the UI
     resolveUiVariant(null);
+
+    if (this.config.enableResumeFromLastPosition === true) {
+      this.resumePositionTracker = new ResumePositionTracker(wrappedPlayer);
+    }
```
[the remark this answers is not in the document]
stonko1994: Addressed in [e8091535](https://github.com/bitmovin/bitmovin-player-ui/commit/e80915358796fd231865c140d762c336a8417612).

src/ts/UIConfig.ts
```diff
@@ -187,6 +187,15 @@ export interface UIConfig {
...
   disableStorageApi?: boolean;
+  /**
+   * If set to true, the UI stores playback progress and resumes known sources from the
+   * last saved position. Source identity is derived from the loaded source title or URL;
+   * sources with no stable identifier are not tracked. Has no effect when
+   * `disableStorageApi` is true or `localStorage` is unavailable.
+   *
+   * Default: false
+   */
+  enableResumeFromLastPosition?: boolean;
```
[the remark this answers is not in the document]
stonko1994: Addressed by updating the PR description to state that resume is disabled by default. The default was already correct in code; the pushed review-fix series ends at [e8091535](https://github.com/bitmovin/bitmovin-player-ui/commit/e80915358796fd231865c140d762c336a8417612).

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,139 @@
...
+    this.player.off(this.player.exports.PlayerEvent.Paused, this.pausePositionTracking);
+    this.player.off(this.player.exports.PlayerEvent.PlaybackFinished, this.finishPositionTracking);
+  }
+
+  getStoredPosition(): number | null {
+    if (!this.activeSourceKey) return null;
+
+    const time = Number(StorageUtils.getItem(this.activeSourceKey));
+    return isFinite(time) && time >= MIN_RESUME_POSITION ? time : null;
+  }
```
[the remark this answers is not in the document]
stonko1994: Addressed in [80e81d4b](https://github.com/bitmovin/bitmovin-player-ui/commit/80e81d4b85b3a2b60aa0b7da7757232a06405560).

src/ts/utils/StorageUtils.ts
```diff
@@ -38,6 +38,20 @@ export namespace StorageUtils {
...
+   */
+  export function removeItem(key: string): void {
+    if (shouldUseLocalStorage()) {
+      try {
+        window.localStorage.removeItem(key);
+      } catch (e) {
+        console.debug(`Failed to remove storage item ${key}`, e);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
stonko1994: Addressed in [d7d9a791](https://github.com/bitmovin/bitmovin-player-ui/commit/d7d9a791153589da3904979ce69fc0e3dae912cc).

src/ts/UIManager.ts
```diff
@@ -399,21 +402,35 @@ export class UIManager {
...
+    // Determine the initial start position.
+    // Either apply the time-deep-link or the resume-position on the player
+    const timestampDeepLinkTargetTime =
+      this.config.enableTimestampDeepLink === true ? TimestampLinkUtils.parseTimestampFromUrl() : null;
+
+    if (this.config.enableResumeFromLastPosition === true) {
+      this.resumePositionTracker = new ResumePositionTracker(wrappedPlayer);
+    }
+
+    if (timestampDeepLinkTargetTime != null || this.resumePositionTracker != null) {
```
dweinber: Always prefer the strict (in-)equality operator (`!==`) over the loose (in-)equality operator (`!=`) if possible.

`timestampDeepLinkTargetTime` is explicitly set to `null` if disabled, and also `TimestampLinkUtils.parseTimestampFromUrl()` returns `null` if not available.

`this.resumePositionTracker` can actually not be `null` but `undefined`.

Options to improve this:
1. Use the same approach that is used to initialize `timestampDeepLinkTargetTime` also for the `resumePositionTracker` (i.e. set it to `null` if not used) and use the strcit operators: `if (timestampDeepLinkTargetTime !== null || this.resumePositionTracker !== null) {`
2. Check for `undefined` if that's the value it might have: `if (timestampDeepLinkTargetTime !== null || this.resumePositionTracker !== undefined) {`
3. Use simple truthy checks in the first place: `if (timestampDeepLinkTargetTime || this.resumePositionTracker) {`


All of these are fine for me, but my suggestion is to use option 3: 

```suggestion
    if (timestampDeepLinkTargetTime || this.resumePositionTracker) {
```

src/ts/UIManager.ts
```diff
@@ -399,21 +402,35 @@ export class UIManager {
...
-        if (isTimestampDeepLinkHandled) return;
-        isTimestampDeepLinkHandled = true;
-        wrappedPlayer.off(this.player.exports.PlayerEvent.SourceLoaded, seekToTimestampDeepLink);
+      const seekToInitialPosition = () => {
         if (wrappedPlayer.isLive()) return;
-        const targetTime = TimestampLinkUtils.parseTimestampFromUrl();
-        if (targetTime != null && targetTime > 0) {
-          wrappedPlayer.seek(targetTime);
+
+        if (timestampDeepLinkTargetTime != null && !isTimestampDeepLinkHandled) {
```
dweinber: Same as above, please avoid loss (in-)equality operators:


```suggestion
        if (timestampDeepLinkTargetTime && !isTimestampDeepLinkHandled) {
```
or
```suggestion
        if (timestampDeepLinkTargetTime !== null && !isTimestampDeepLinkHandled) {
```

src/ts/UIManager.ts
```diff
@@ -399,21 +402,35 @@ export class UIManager {
...
+          isTimestampDeepLinkHandled = true;
+          if (timestampDeepLinkTargetTime > 0) {
+            wrappedPlayer.seek(timestampDeepLinkTargetTime, 'ui');
+          }
+        } else {
+          const storedPosition = this.resumePositionTracker?.getStoredPosition();
+          if (storedPosition != null) {
```
dweinber: Loose inequality again:

```suggestion
          const storedPosition = this.resumePositionTracker?.getStoredPosition() ?? null;
          if (storedPosition !== null) {
```

src/ts/UIManager.ts
```diff
@@ -399,21 +402,35 @@ export class UIManager {
...
+            wrappedPlayer.seek(storedPosition, 'ui');
+          }
         }
       };
-      wrappedPlayer.on(this.player.exports.PlayerEvent.SourceLoaded, seekToTimestampDeepLink);
-      // Source may already be loaded by the time the UI is built (e.g. variant switch).
-      if (wrappedPlayer.getSource() != null) seekToTimestampDeepLink();
+      wrappedPlayer.on(this.player.exports.PlayerEvent.SourceLoaded, seekToInitialPosition);
+      // Source may already be loaded by the time the UI is built.
+      if (wrappedPlayer.getSource() != null) seekToInitialPosition();
```
dweinber: Avoid loose inequality. But this line may have a bigger side effect:

Re-creating a UIManager on an already-playing source fires a spurious backward seek. Release-and-recreate mid-playback (the demo page's own UI switcher does exactly this) persists lastPosition on release, then the new constructor sees getSource() != null and immediately seeks back to the just-saved, TimeChanged-trailing position, interrupting playback.

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,140 @@
+import { PlayerAPI, SourceConfig, TimeChangedEvent } from 'bitmovin-player';
+import { StorageUtils } from './StorageUtils';
+
+const STORAGE_KEY_PREFIX = 'bitmovin.player.ui.resume.';
+const MIN_RESUME_POSITION = 5;
```
dweinber: I think those two should be configurable for these reasons:

1. Some customers want to avoid that `bitmovin` shows up anywhere, that's also why the UI CSS prefix can be changed at UI build time, for example. But here we're now hardcoding it.
2. Customers may decided they want to always store the resume position (i.e. `MIN_RESUME_POSITION = 0` or `MIN_RESUME_POSITION = 1`), or only once a user has watched at least a minute or two (`MIN_RESUME_POSITION = 60`). This value might even change for a single customer depending on content duration or something like that.

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  constructor(private readonly player: PlayerAPI) {
+    player.on(player.exports.PlayerEvent.SourceLoaded, this.startPositionTracking);
+    player.on(player.exports.PlayerEvent.SourceUnloaded, this.unloadSource);
+    player.on(player.exports.PlayerEvent.TimeChanged, this.updatePosition);
+    player.on(player.exports.PlayerEvent.Paused, this.pausePositionTracking);
+    player.on(player.exports.PlayerEvent.PlaybackFinished, this.finishPositionTracking);
+    window.addEventListener('beforeunload', this.storeCurrentPosition);
+
+    if (player.getSource() != null) this.startPositionTracking();
```
dweinber:
```suggestion
    if (player.getSource() !== null) this.startPositionTracking();
```

src/ts/utils/ResumePositionTracker.ts
dweinber: We also may want to consider a `MAX_RESUME_POSITION`, i.e. if the user stops watching during the movie's credits, we may want to consider not offering the resume point there. Default could still be Infinity / asset duration though, but customers should be able to configure it.

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,140 @@
...
+  const id = sourceIdentifier(source);
+  return id ? STORAGE_KEY_PREFIX + hashSourceIdentifier(id) : null;
+}
+
+function sourceIdentifier(source: SourceConfig | null): string | null {
+  if (!source) return null;
+  if (source.title) return 'title:' + source.title;
+  if (source.dash) return 'dash:' + source.dash;
+  if (source.hls) return 'hls:' + source.hls;
+  if (source.smooth) return 'smooth:' + source.smooth;
```
dweinber, line 139: Nice in theory, but I wonder how well that will work. Many times URLs are token-protected, so they might not be the same within two loads. Not sure what the best solution is here 😕 

We may want to consider checking if the analytics video_title / video_id is set and use those if available.

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  private readonly updatePosition = (event: TimeChangedEvent) => {
+    if (this.player.isLive()) return;
+
+    if (isFinite(event.time)) {
+      this.lastPosition = event.time;
+    }
+  };
+
+  private readonly savePosition = (time: number | null) => {
```
dweinber, line 50: I think there's one big case we're missing here:

If the user watches an ad and decides to stop watching _during the ad_, the current time will be the time within the ad 😬 
This may lead to dropping a stored resume position or storing the position of the ad (if time in ad >= MIN_RESUME_POSITION)

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  release(): void {
+    this.storeCurrentPosition();
+    window.removeEventListener('beforeunload', this.storeCurrentPosition);
+    this.player.off(this.player.exports.PlayerEvent.SourceLoaded, this.startPositionTracking);
+    this.player.off(this.player.exports.PlayerEvent.SourceUnloaded, this.unloadSource);
+    this.player.off(this.player.exports.PlayerEvent.TimeChanged, this.updatePosition);
+    this.player.off(this.player.exports.PlayerEvent.Paused, this.pausePositionTracking);
+    this.player.off(this.player.exports.PlayerEvent.PlaybackFinished, this.finishPositionTracking);
+  }
```
dweinber, line 30: If the player is destroyed before calling `uiManager.release()`, this method will throw several `PlayerAPINotAvailableError` because `.off` and `.isLive` are called on the destroyed player.

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,140 @@
...
+  };
+
+  private readonly savePosition = (time: number | null) => {
+    if (this.player.isLive()) {
+      this.activeSourceKey = null;
+      this.lastPosition = null;
+      return;
+    }
+
+    if (!this.activeSourceKey) return;
```
dweinber, line 51: If a user watches to the end (`PlaybackFinished` event is triggered), the tracking is stopped and `this.activeSourceKey` is set to `null`. That's fine.
But if a user then clicks play again to re-watch it, the tracking is never started again. Is this expected?

spec/UIManager.spec.ts
```diff
@@ -535,3 +636,12 @@ describe('UIManager', () => {
...
   });
 });
+
+function resumeStorageKeyForSourceIdentifier(sourceIdentifier: string): string {
+  let hash = 0;
+  for (let i = 0; i < sourceIdentifier.length; i++) {
+    hash = (hash << 5) - hash + sourceIdentifier.charCodeAt(i);
+    hash |= 0;
+  }
+  return `bitmovin.player.ui.resume.${(hash >>> 0).toString(36)}`;
```
dweinber: NIT: The hardcoded string isn't nice. If the `STORAGE_KEY_PREFIX` in `ResumePositionTracker` changes, this needs to be updated manually.

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,140 @@
...
+      this.activeSourceKey = null;
+      this.lastPosition = null;
+      return;
+    }
+
+    this.activeSourceKey = storageKeyFor(this.player.getSource());
+  };
+
+  private readonly updatePosition = (event: TimeChangedEvent) => {
+    if (this.player.isLive()) return;
```
dweinber: So we really need to call `isLive` on every `TimeChanged` event? Maybe we stop tracking when we know a source is live (ignoring the live-to-vod transition case, or re-establish tracking in that case)?

dweinber (changes requested):
I'm still not sure if this really belongs in the UI, or if this should be part of the UI. But looks like that decision was already made 🤷

src/ts/utils/ResumePositionTracker.ts
stonko1994: I thought about this as well and thought about something like a VideoID. But the UI is only developed against the Player API, and there is nothing like that to identify a source across loads.

I didn't want to introduce such a new API so all I could do was relying on what the current API gives us. Happy to discuss further, but for the initial version of this feature (MVP), I'd stick to this.

src/ts/utils/ResumePositionTracker.ts
stonko1994: Re 2.:
For the initial version of this feature (MVP) I think hard-coding this (or potentially even removing it) might be sufficient or? Otherwise, we add configs for features that will not be used much. Personally, I'd see the whole feature more like a start-offset-like feature (which the player already has), so I didn't want to overcomplicate it.

spec/UIManager.spec.ts
stonko1994: IMO, that's the point of a test. Catching unintentional changes to such a string. Testing against the actual value will always silently succeed and there is no point in such a test.

src/ts/UIManager.ts
stonko1994: 1d7d6d7b9b80b4e8c7eafc260e79cc8460566b95

src/ts/UIManager.ts
stonko1994: 1d7d6d7b9b80b4e8c7eafc260e79cc8460566b95

src/ts/UIManager.ts
stonko1994: 1d7d6d7b9b80b4e8c7eafc260e79cc8460566b95

src/ts/UIManager.ts
stonko1994: f92d37e565de6b02168df6c94dec5b99cb4d9cba

src/ts/utils/ResumePositionTracker.ts
stonko1994: f1ef1b8cebf68d1165b82fa41cac70498b23b634 I also fixed the newly introduced storage key from #908

src/ts/utils/ResumePositionTracker.ts
stonko1994: 1d7d6d7b9b80b4e8c7eafc260e79cc8460566b95

src/ts/utils/ResumePositionTracker.ts
stonko1994: 472cf96299591a1c72e0ec3837ef43f1479207fd

src/ts/utils/ResumePositionTracker.ts
stonko1994: 8bbd0663915073405345aa1b50f975c4553f4433

src/ts/utils/ResumePositionTracker.ts
stonko1994: 8bbd0663915073405345aa1b50f975c4553f4433

spec/UIManager.spec.ts
stonko1994: f1ef1b8cebf68d1165b82fa41cac70498b23b634

src/ts/utils/ResumePositionTracker.ts
stonko1994: 8bbd0663915073405345aa1b50f975c4553f4433

src/ts/utils/ResumePositionTracker.ts
felix-hoc: Agree with @dweinber, and especially having the `title` as the first option could be dangerous. Titles may be reused or have default values.

We could hash the full `source` object instead (or at least all of the relevant fields).

src/ts/UIManager.ts
```diff
@@ -410,21 +414,35 @@ export class UIManager {
...
-        isTimestampDeepLinkHandled = true;
-        wrappedPlayer.off(this.player.exports.PlayerEvent.SourceLoaded, seekToTimestampDeepLink);
+      const seekToInitialPosition = () => {
         if (wrappedPlayer.isLive()) return;
-        const targetTime = TimestampLinkUtils.parseTimestampFromUrl();
-        if (targetTime != null && targetTime > 0) {
-          wrappedPlayer.seek(targetTime);
+
+        if (timestampDeepLinkTargetTime !== null && !isTimestampDeepLinkHandled) {
+          isTimestampDeepLinkHandled = true;
```
felix-hoc, line 438: should we not also set this to `true` when we load a live source? Otherwise it will apply to any VoD source loaded afterwards, which goes against the doc: https://github.com/bitmovin/bitmovin-player-ui/blob/a8908bf0cbde9928cab20d5baa31ae3402232172/src/ts/UIConfig.ts#L215-L223

src/ts/utils/ResumePositionTracker.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    const activeSourceKey = storageKeyFor(this.player.getSource());
+    if (!activeSourceKey) return;
+
+    this.activeSourceKey = activeSourceKey;
+    this.player.on(this.player.exports.PlayerEvent.TimeChanged, this.updatePosition);
+    this.player.on(this.player.exports.PlayerEvent.Paused, this.pausePositionTracking);
+    this.player.on(this.player.exports.PlayerEvent.PlaybackFinished, this.finishPositionTracking);
+    window.addEventListener('beforeunload', this.storeCurrentPosition);
```
felix-hoc: I'm wondering if we even need to update the stored position on each `TimeChanged`, as it will be spamming LocalStorage quite a bit (not sure if that is an issue, though). I'm wondering if we could just listen for `Paused` / `PlaybackFinished` / `beforeunload` / maybe `AdBreakStarted` to get the position. Or at least only save it to LocalStorage when unloading / moving away from the page.

felix-hoc (commented):
seems to work well, had a few comments though

src/ts/utils/ResumePositionTracker.ts
stonko1994: I agree as well, but I also don't want to invent new **Player** APIs for this feature, which we don't know how/if it will be used. _There is no customer request behind this feature._

This approach is best effort with the options we have available. If someone turns on this feature, I'd assume they know how to identify their videos. Hashing the source object takes away all options to use this feeature in case tokenized urls are used (as even unique titles wouldn't be identified as unique anymore).

One other option I could see is to extend the `UIConfig.metadata` with an 'identifier' attribute. Then the requirement would be to pass that through to the `player.load` (which requires `any` or another cast ofc.). I'm not into JS/TS deep enough to know if this is something we want our customers to do.

stonko1994: @dweinber @felix-hoc any input? The only viable option would be to use a `videoId` on `SourceConfig.metadata` or something like this. We could document it on the new `uiconfig` property.

felix-hoc: Fair, I also would not suggest to introduce a new Player API just for this. I was rather thinking about solving it using the info we have, but fair point about tokenized urls.

Fine for me to go with the current solution until we get more customer feedback 👍🏼  I like the idea about `SourceConfig.metadata.videoId?` though, maybe we can revisit that later.

dweinber: I think the primary attempt should be to use `SourceConfig.analytics?.videoId`, if available: https://cdn.bitmovin.com/player/web/8/docs/interfaces/Analytics.AnalyticsConfig.html#videoId

We could also add a `SourceConfig.metadata.videoId` or something like that if you want.

But yeah, after that let's keep it as is for now.

src/ts/utils/ResumePositionTracker.ts
dweinber: Fair enough, fine for me.

dweinber (approved):
Thanks for addressing my concerns. LGTM 🙂

src/ts/utils/ResumePositionTracker.ts
stonko1994: > SourceConfig.analytics?.videoId

I'm thinking what if someone wants this feature who has no idea about analytics (and is not using it). Adding a requirement in the UI to the analytics 'domain' is not a good idea IMO.

src/ts/utils/ResumePositionTracker.ts
stonko1994: I did some research and the local storage access might indeed be a problem on low end devices. I switched to an event driven approach to store the state on the events you mentioned.

Also MDN recommends to use `visibilitychange` instead of `beforeunload` for storing state ([ref](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#:~:text=Note:%20It%20is%20recommended%20to%20use%20the%20visibilitychange%20event%20as%20a%20more%20reliable%20signal%20for%20automatic%20app%20state%20saving%20that%20gets%20around%20problems%20like%20the%20above.%20See%20Don't%20lose%20user%20and%20app%20state%2C%20use%20Page%20Visibility%20for%20more%20details.))

[f338088](https://github.com/bitmovin/bitmovin-player-ui/pull/871/commits/f338088cf59c6c239c847929990c7ad4045a049a)

src/ts/UIManager.ts
stonko1994: [ac5e42e](https://github.com/bitmovin/bitmovin-player-ui/pull/871/commits/ac5e42e0f45b720424c94be9783ce2c6adb5df12)

PR #872, opened 2026-05-11
Persist volume, mute, and playback speed across sessions

lukaskroepfl:
## Summary

Adds `UIPreferences`, a tiny utility that:

1. Subscribes to `VolumeChanged`, `Muted` / `Unmuted`, and `PlaybackSpeedChanged` events to write the current value to `localStorage`.
2. Re-applies the stored values on the `Ready` event so the user picks up where they left off across sessions and across in-page player rebuilds (e.g. UI variant switch).

Bounded reads / writes (volume 0-100, speed 0-4) so a corrupted localStorage entry can't crash the player.

All writes route through the existing `StorageUtils`. On devices where `localStorage` is unavailable or the existing `disableStorageApi` flag is set (private browsing, RDK / set-top boxes, restricted WebViews), the feature degrades to a no-op without throwing.

Wired into `UIManager` init. Enabled by default; opt-out via new `UIConfig.enablePersistentPreferences` flag.

Split out from #869.

## Test plan

- [x] 503 unit tests pass; ESLint and `tsc --noEmit` clean.
- [ ] Set volume to 50 %, mute, change speed to 1.5×, reload the page → volume / mute / speed restored after `Ready`.
- [ ] Set `disableStorageApi: true` → no preferences saved or restored; no errors.
- [ ] Set `enablePersistentPreferences: false` → no preferences saved or restored.
- [ ] In private browsing → no errors, no save.


stonko1994 (commented):
Shouldn't this be rather a Player feature to also enable the same when our UI is not used?

lukaskroepfl:
Fair point, but I'd argue persistence is a UX concern rather than a playback one. The Player already exposes the state (volume, mute, speed) via its API — what to do with it across sessions is a product decision that belongs in whatever layer owns the UX.

Anyone building a custom UI is already making opinionated calls about controls, hover behavior, keyboard shortcuts, mobile gestures, etc. "Should we remember the user's last volume?" sits in the same bucket. They'll either implement it themselves, or deliberately choose not to (kiosk mode, muted-autoplay landing pages, A/B fresh state).

If the Player did it by default, those customers would have to actively *disable* it, and some wouldn't notice until a user reported that their "always muted" page wasn't behaving that way.

So I'd keep it here. Happy to revisit if there's appetite to make it a Player-level opt-in down the road.

src/ts/UIConfig.ts
```diff
@@ -159,6 +159,14 @@ export interface UIConfig {
...
    */
   disableStorageApi?: boolean;
+  /**
+   * If set to true, the UI persists volume, mute state, and playback speed across sessions
+   * and reapplies them when a player using this UI is initialized. Has no effect when
+   * `disableStorageApi` is true or `localStorage` is unavailable.
+   *
+   * Default: true
+   */
+  enablePersistentPreferences?: boolean;
```
stonko1994: IMO, enabling this by default now is super dangerous. It works well but also affects systems such as our own internal player tests or every customer using the latest UI version.

Use-cases:
- I'm pretty certain the playback speed will break a lot of our automated tests (at least on iOS).
- If a customer build a custom persistant approach already this might interfere with it.
- It could be confusing for end-user to automatically apply stuff like playback speed to all videos you watch with the Bitmovin Player (all platforms)

What do you think aobut the idea to add another config which shows a new toggle to give the end-user the option if they want persistent data? I could think that a lot of users prefer explicit opt-in over 'magically' remembering a selection.

src/ts/utils/UIPreferences.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Applies the persisted preferences to the player. Should be called after the player is
+   * ready to accept volume / mute / speed calls (typically on `Ready` or after the source
+   * has been loaded). Each preference is applied independently — missing entries are skipped.
+   */
+  export function apply(player: PlayerAPI): void {
+    const volume = readNumber(KEY_VOLUME);
+    if (volume !== null && volume >= 0 && volume <= 100) {
+      try {
+        player.setVolume(volume, ISSUER);
```
stonko1994: Why should those API calls fail? Why do we need so conservative with calling our own Player APIs? **All occurences**

src/ts/UIConfig.ts
lukaskroepfl: Good call — flipped the default to opt-in in 0ae166b0 so existing integrations (automated tests, customer custom persistence layers) aren't affected. The end-user toggle idea is worth a follow-up but I'd keep this PR focused on the persistence plumbing.

src/ts/utils/UIPreferences.ts
lukaskroepfl: Fair — dropped the try/catch wrappers around all the player API calls in 0ae166b0.

src/ts/UIConfig.ts
lukaskroepfl: Followed up in 0acc9bfd — added a `PersistentPreferencesToggleButton` that appears in the settings panel when the integrator opts in via `enablePersistentPreferences`. The toggle's own state is persisted, so the end user gets an explicit opt-in that survives reloads. Flipping it on captures the current volume / mute / speed immediately; flipping it off clears the stored values.

lukaskroepfl:
@123mpozzi heads-up — while adding the new `PersistentPreferencesToggleButton` I copied the SCSS from `_eco-mode-toggle-button.scss` and ran into a horizontal jump on toggle. Root cause: the `.bmpui-on` state has `margin-left: 2%` on the icon while `.bmpui-off` doesn't, so the icon (and the surrounding row) shifts a few pixels when toggling.

Fixed it for the new toggle and also dropped the offset from EcoMode in 936edbf2 since I assume the shift was unintentional. Let me know if it was actually load-bearing for some EcoMode-specific layout and I'll revert that part.

123mpozzi:
> @123mpozzi heads-up — while adding the new `PersistentPreferencesToggleButton` I copied the SCSS from `_eco-mode-toggle-button.scss` and ran into a horizontal jump on toggle. Root cause: the `.bmpui-on` state has `margin-left: 2%` on the icon while `.bmpui-off` doesn't, so the icon (and the surrounding row) shifts a few pixels when toggling.
> 
> Fixed it for the new toggle and also dropped the offset from EcoMode in [936edbf](https://github.com/bitmovin/bitmovin-player-ui/commit/936edbf25d09b7bdd66ebfafa68e22fb3402494d) since I assume the shift was unintentional. Let me know if it was actually load-bearing for some EcoMode-specific layout and I'll revert that part.

Thanks for the heads-up. I am no expert on EcoMode styling/elements, but the change sounds fine to me if the layout still behaves correctly.

src/scss/components/buttons/_eco-mode-toggle-button.scss
```diff
@@ -15,7 +15,6 @@
       background-image: svg-load('../../assets/images/toggleOn.svg', fill=$color-icon);
       background-position: 20px center;
       background-size: 45% auto;
-      margin-left: 2%;
```
stonko1994, line 18: Why this change? Usually it's preferred to keep unrelated changes to a minimum.

src/ts/UIConfig.ts
```diff
@@ -159,6 +159,19 @@ export interface UIConfig {
...
+   * state, and playback speed across sessions and reapplies them when a player using this
+   * UI is initialized. Has no effect when `disableStorageApi` is true or `localStorage`
+   * is unavailable.
+   *
+   * Disabled by default so existing integrations (including automated tests and custom
+   * persistence layers) are not affected. Opt in explicitly per UI instance.
+   *
+   * Default: false
+   */
+  enablePersistentPreferences?: boolean;
```
stonko1994, line 202: @lukaskroepfl What do you think from a DX perspective about this property? I'm thinking that for the best customization we should offer both properties.
1. a property to enable/disable the storage so whoever integrate our Player can decide if they want to have that automatically or
2. a property to enable/disable the settings panel toggle to delegate the decision to the end user.

src/scss/components/buttons/_persistent-preferences-toggle-button.scss
```diff
@@ -0,0 +1,29 @@
...
+  &:hover {
+    @include svg-icon-shadow;
+  }
+
+  &.#{$prefix}-on {
+    .#{$prefix}-ui-icon {
+      background-image: svg-load('../../assets/images/toggleOn.svg', fill=$color-icon);
+      background-position: 20px center;
+      background-size: 45% auto;
+      margin-left: 2%;
```
stonko1994: You just removed this from the eco mode toggle but still have it here? Is that intentional?

src/ts/utils/UIPreferences.ts
```diff
@@ -0,0 +1,131 @@
...
+ * {@link StorageUtils} which already swallows storage errors.
+ *
+ * Persistence is gated by an end-user opt-in flag stored alongside the preferences
+ * themselves (see {@link isEnabled} / {@link setEnabled}). When the integrator enables
+ * the feature via `UIConfig.enablePersistentPreferences`, the bundled default UI exposes
+ * a toggle in the settings panel that flips this flag.
+ *
+ * @category Utils
+ */
+export namespace UIPreferences {
```
stonko1994: Let's switch this implementation to the same pattern as the `SubtitleSettingsManager` for consistency. This should improve the usage of this in the UIManager as the `UIPreferencesManager` can handle the player lifecycle then instead of the `UIManager`.

src/ts/utils/UIPreferences.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Subscribes to the player events that change a tracked preference and writes the
+   * latest value back to storage. Writes are skipped while {@link isEnabled} returns
+   * `false`, so the listeners can be attached unconditionally.
+   */
+  export function attach(player: PlayerAPI): void {
+    player.on(player.exports.PlayerEvent.VolumeChanged, () => {
+      if (!isEnabled()) return;
```
stonko1994: All those isEnabled check should go away after chaging to the `UIPreferencesManager` pattern.

src/ts/utils/UIPreferences.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Applies the persisted preferences to the player. Should be called after the player is
+   * ready to accept volume / mute / speed calls (typically on `Ready` or after the source
+   * has been loaded). Each preference is applied independently — missing entries are skipped.
+   * No-op when {@link isEnabled} returns `false`.
+   */
+  export function apply(player: PlayerAPI): void {
+    if (!isEnabled()) return;
```
stonko1994: All those isEnabled check should go away after chaging to the `UIPreferencesManager` pattern.

src/ts/utils/UIPreferences.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    const muted = readBoolean(KEY_MUTED);
+    if (muted === true) {
+      player.mute(ISSUER);
+    } else if (muted === false) {
+      player.unmute(ISSUER);
+    }
+
+    const speed = readNumber(KEY_PLAYBACK_SPEED);
+    if (speed !== null && speed > 0 && speed <= 4) {
```
stonko1994: Why do we restrict the speed to 0 and 4 here? What if the player supports more than that and customers want more than that?

src/ts/utils/UIPreferences.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (volume !== null && volume >= 0 && volume <= 100) {
+      player.setVolume(volume, ISSUER);
+    }
+
+    const muted = readBoolean(KEY_MUTED);
+    if (muted === true) {
+      player.mute(ISSUER);
+    } else if (muted === false) {
+      player.unmute(ISSUER);
+    }
```
stonko1994: This reads weird. Wouldn't `Boolean(muted)` read much better?

src/ts/UIFactory.ts
```diff
@@ -746,6 +755,20 @@ export namespace UIFactory {
       components.unshift(ecoModeContainer);
     }
 
+    if (enablePersistentPreferences) {
```
stonko1994: If the `disableStorageApi` is disabled, this is still added but has no effect. This is against the documentation of the property.

src/ts/UIManager.ts
```diff
@@ -248,6 +249,13 @@ export class UIManager {
     }
     this.subtitleSettingsManager.initialize();
 
+    if (uiconfig.enablePersistentPreferences === true) {
+      UIPreferences.attach(this.player);
+      const applyPrefs = () => UIPreferences.apply(this.player);
+      this.player.on(this.player.exports.PlayerEvent.Ready, applyPrefs);
+      applyPrefs();
```
stonko1994: Let's improve this after switching to the `subtitleSettingsManager` pattern.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- `UIPreferences`: persists volume, mute state, and playback speed across sessions and reapplies them on player init (after `Ready`). Configurable via `UIConfig.enablePersistentPreferences` (default `true`). All persistent storage routes through the existing `disableStorageApi` flag and degrades to a no-op when `localStorage` is unavailable (private browsing, restricted WebViews, RDK / set-top boxes).
```
stonko1994: We are trying to write the changelog entries as sentences including the heading term:


```suggestion
- `UIConfig.enablePersistentPreferences` to configure whether the UI persists volume, mute state, and playback speed across sessions and reapplies them on player init.
```

stonko1994 (commented):
Is it intentional to have the new setting in the middle of the existing settings? Dosn't look good to me. Ideally we can move that into a 'more' section in the settings if we really want to have this.

Image

+1 let's move the toggle to the right instead of in the middle of the row?

lukaskroepfl:
Thanks for the thorough pass — pushed `e05818d` addressing the round (and merged latest `develop`). Summary of how each point was handled:

**Architecture**
- **`UIPreferences` → `UIPreferencesManager`** (SubtitleSettingsManager pattern): done. It's now a class that owns the player lifecycle via `configure(player, …)` / `release()`. `UIManager` just creates, configures and releases it.
- **`isEnabled()` guards everywhere**: gone. The manager only subscribes to the player events while enabled and tears them down on disable/release, so there's nothing to guard.
- **`UIManager.ts` inline wiring**: replaced by the manager; `UIManager` no longer hand-wires the `Ready`/persist listeners.

**DX — config shape (your `UIConfig` question)**: went with offering **both**, as you suggested:
- `enablePersistentPreferences` — auto-persist + reapply, no end-user interaction.
- `showPersistentPreferencesToggle` — shows the settings-panel opt-in toggle (the user's choice is itself persisted; `enablePersistentPreferences` seeds its initial state).
Both default to `false` and are inert when `disableStorageApi` is set.

**Fixes**
- **Toggle added when storage disabled** (`UIFactory`): now gated on `disableStorageApi !== true`; the manager is also only configured when storage is usable.
- **Speed clamped to ≤ 4**: removed the upper clamp — restores whatever the player supports (still guards `> 0`).
- **`Boolean(muted)`**: reworked into a clean `if (muted !== null) { muted ? mute() : unmute() }`.
- **EcoMode margin change**: reverted — left the EcoMode SCSS untouched. The new toggle simply omits the asymmetric `margin-left` so its on/off states don't shift.
- **CHANGELOG**: rewritten as a sentence including the config terms.

**Placement**: moved the toggle to the **bottom** of the settings list (after Subtitles) and right-aligned it into the value column. Verified in-browser; also confirmed the full persist → reload → restore and opt-out → clear cycle works end-to-end.

Left the "should this be a Player feature" question as-is per the earlier discussion — happy to revisit separately if there's appetite.

stonko1994 (commented):
Manual testing outcome:
- When enabling `enablePersistentPreferences` and `showPersistentPreferencesToggle` the default value of the toggle is still off. This is kinda undexpected to me and had to figure out why it's not working. IMO if `enablePersistentPreferences` is enabled the toggle should be on.
- The toggle is still misaligned. The controls should be aligned trailing in the menu

Image

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- `UIConfig.enablePersistentPreferences` and `UIConfig.showPersistentPreferencesToggle` to configure whether the UI persists volume, mute state, and playback speed across sessions and reapplies them on player init — automatically and/or via an end-user opt-in toggle in the settings panel. Both default to `false`; they have no effect when `disableStorageApi` is set or `localStorage` is unavailable (private browsing, restricted WebViews, RDK / set-top boxes).
```
stonko1994: Please merge development and move the entry to the unreleased section.

stonko1994: Nit: this entry is very detailed. feels like parts of it should be in the documentation and not in the changelog itself.

```suggestion
- `UIConfig.enablePersistentPreferences` and `UIConfig.showPersistentPreferencesToggle` to configure whether the UI persists volume, mute state, and playback speed across sessions.
```

lukaskroepfl:
Thanks — all four addressed, and I merged latest `develop`:

1. **Toggle defaulted off with `enablePersistentPreferences` on** — that was a real bug, good catch. `configure()` was routing through `setEnabled()`, which always persisted the enabled flag — so an earlier load with the feature defaulting *off* wrote `enabled=0`, and that stored `0` then shadowed `enablePersistentPreferences=true` on later loads (toggle stuck off). Fixed: `configure()` now applies the resolved state (stored choice ?? integrator default) **without** persisting it, and only an explicit end-user toggle writes the flag. So `enablePersistentPreferences=true` now reliably shows the toggle **on**, while a user's explicit opt-in/out still persists. Verified in-browser: fresh load with the flag on → toggle on and nothing written to storage until the user acts; explicit off → persisted `0`; explicit on → persisted `1` + captures current state.

2. **Toggle alignment** — fixed; the toggle graphic now sits at the trailing edge, lined up with the value column of the other rows (moved the icon's `background-position` to `right`). Verified visually.

3. **CHANGELOG → develop + Unreleased** — merged `develop` (it had advanced quite a bit — `PlayerContextMenu`, `PlayerInsightsPanel`, stats-for-nerds) and moved the entry into the `## [Unreleased]` section.

4. **CHANGELOG too detailed** — trimmed to your suggested one-liner.

tsc / eslint clean, 548 tests pass. Ready for another look 🙏

stonko1994:
Closing in favor of #907 and #908

PR #878, opened 2026-05-20
Add support to add/remove recommendations at runtime

stonko1994:
## Description

Currently, it's only possible to specify the recommendations statically before loading a source. Recommendation data can now be updated after source load. This lets applications fetch recommendations during playback and inject them into the UI before or after playback finishes, without reloading the source.

## Changes
`UIManager.recommendations` exposes a small API for dynamically managing recommendation items:

- `add(recommendation)`
- `remove(recommendation)`
- `list()`

## Manual Testing

Open the demo page and use the browser console:

```js
const recommendation = {
  title: 'Recommendation 1: The best video ever',
  resource: {
    dash: 'https://cdn.bitmovin.com/content/assets/sintel/sintel.mpd',
    poster: 'https://cdn.bitmovin.com/content/assets/sintel/poster.png',
  },
  duration: 10.4,
};

uiManager.recommendations.add(recommendation);
uiManager.recommendations.remove(recommendation);
```

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/UIManager.ts
+++ b/src/ts/UIManager.ts
@@ -262,7 +262,7 @@ export class UIManager {
      */
     const updateConfig = () => {
       const playerSourceConfig = player.getSource() || {};
-      this.config.metadata = JSON.parse(JSON.stringify(uiconfig.metadata || {}));
+      this.config.metadata = { ...uiconfig.metadata };
 
       // Extract the UI-related config properties from the source config
       const playerSourceUiConfig: UIConfig = {
@@ -280,9 +280,10 @@ export class UIManager {
       // lifetime of the player instance.
       this.config.metadata.title = playerSourceUiConfig.metadata.title || uiconfig.metadata.title;
       this.config.metadata.description = playerSourceUiConfig.metadata.description || uiconfig.metadata.description;
-      this.config.metadata.markers = playerSourceUiConfig.metadata.markers || uiconfig.metadata.markers || [];
-      this.config.metadata.recommendations =
-        playerSourceUiConfig.metadata.recommendations || uiconfig.metadata.recommendations || [];
+      this.config.metadata.markers = [...(playerSourceUiConfig.metadata.markers || uiconfig.metadata.markers || [])];
+      this.config.metadata.recommendations = [
+        ...(playerSourceUiConfig.metadata.recommendations || uiconfig.metadata.recommendations || []),
+      ];
 
       StorageUtils.setStorageApiDisabled(uiconfig);
     };
--- a/src/ts/UIManager.ts
+++ b/src/ts/UIManager.ts
@@ -7,7 +7,7 @@ import { NoArgs, EventDispatcher, CancelEventArgs } from './EventDispatcher';
 import { UIUtils } from './utils/UIUtils';
 import { ArrayUtils } from './utils/ArrayUtils';
 import { BrowserUtils } from './utils/BrowserUtils';
-import { TimelineMarker, UIConfig } from './UIConfig';
+import { RecommendationConfig, TimelineMarker, UIConfig } from './UIConfig';
 import { PlayerAPI, PlayerEventCallback, PlayerEventBase, PlayerEvent, AdEvent, LinearAd } from 'bitmovin-player';
 import { VolumeController } from './utils/VolumeController';
 import { i18n, CustomVocabulary, Vocabularies, I18n, LanguageChangedArgument } from './localization/i18n';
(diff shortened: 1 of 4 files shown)


src/ts/UIManager.ts
```diff
@@ -211,6 +234,24 @@ export class UIManager {
...
+      remove: (recommendation: RecommendationConfig): boolean => {
+        if (ArrayUtils.remove(this.config.metadata.recommendations, recommendation) === recommendation) {
+          this.config.events.onUpdated.dispatch(this);
+          return true;
+        }
+
+        return false;
+      },
+      list: (): RecommendationConfig[] => {
+        return this.config.metadata.recommendations;
```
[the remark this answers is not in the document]
stonko1994: [90892ca](https://github.com/bitmovin/bitmovin-player-ui/pull/878/commits/90892ca4184b2df79b98ebbb915e2afd2cfd814d)

src/ts/UIManager.ts
```diff
@@ -65,6 +65,28 @@ export interface InternalUIConfig extends UIConfig {
...
 }
 
+/**
+ * API for managing recommendations displayed by the {@link RecommendationOverlay}.
+ */
+export interface RecommendationsApi {
+  /**
+   * Adds a recommendation which will be displayed in the {@link RecommendationOverlay}.
+   */
+  add(recommendation: RecommendationConfig): void;
```
dweinber, line 79: Might be worth pointing out that no de-duplication happens (or add de-duplication)

src/ts/UIManager.ts
```diff
@@ -65,6 +65,28 @@ export interface InternalUIConfig extends UIConfig {
   adBreakTracker: AdBreakTracker;
 }
 
+/**
+ * API for managing recommendations displayed by the {@link RecommendationOverlay}.
+ */
+export interface RecommendationsApi {
```
dweinber, line 71: Should we add a note that recommendations are cleared when a new source is loaded into the player? Not sure if that's clear for developers.

dweinber (commented):
Nice improvement! I like the API namespace ❤️ 
A few nits on the JSDoc, the rest looks good and works nicely!

src/ts/UIManager.ts
stonko1994: [3b2cb8d](https://github.com/bitmovin/bitmovin-player-ui/pull/878/commits/3b2cb8de882d2151a04e3e2cc717b32312e59711)

src/ts/UIManager.ts
stonko1994: Actually, just dynamically scheduled recommendations are cleared. The one from the UIConfig are re-applied. [3b2cb8d](https://github.com/bitmovin/bitmovin-player-ui/pull/878/commits/3b2cb8de882d2151a04e3e2cc717b32312e59711)

dweinber: If I have recommendations in the SourceConfig and schedule additional ones dynamically, all are cleared when changing the source.

If I have no static recommendations configured anywhere and only schedule some dynamically, all are cleared when changing the source.

If I have recommendations in the UIConfig and schedule additional ones dynamically, all stay across sources (also the dynamic ones).

stonko1994: That is true yes. Should we outline all 3 use-cases explicitly?

dweinber: I think we should have dynamically scheduled recommendations either always stick or always be removed (I vote for removed)

stonko1994: [98c2211](https://github.com/bitmovin/bitmovin-player-ui/pull/878/commits/98c2211c5fc7d2c78e59892edb0baf56158595a2)

dweinber: Most scenarios are covered now as it seems:
- ✅ recommendations in source 
- ✅ recommendations in source + dynamic
- ✅ recommendations in config
- ✅ recommendations in config + dynamic
- ✅ recommendations only dynamic
- ❌ recommendations in config + source + dynamic
  - recommendations from config are ignored

Is this on purpose?

Additionally, if I schedule a dynamic recommendation in the player's `load` promise resolve path (`player.load(...).then(() => uiManager.recommendations.add(...))`), it is ignored. as soon as I add a timeout of even `0` it works (`player.load(...).then(() => setTimeout(() => uiManager.recommendations.add(...), 0))`).
Not necessarily a blocker but we should document/clarify the requirements when the API may be used.

src/ts/UIManager.ts
```diff
@@ -235,9 +280,10 @@ export class UIManager {
       // lifetime of the player instance.
       this.config.metadata.title = playerSourceUiConfig.metadata.title || uiconfig.metadata.title;
       this.config.metadata.description = playerSourceUiConfig.metadata.description || uiconfig.metadata.description;
-      this.config.metadata.markers = playerSourceUiConfig.metadata.markers || uiconfig.metadata.markers || [];
-      this.config.metadata.recommendations =
-        playerSourceUiConfig.metadata.recommendations || uiconfig.metadata.recommendations || [];
+      this.config.metadata.markers = [...(playerSourceUiConfig.metadata.markers || uiconfig.metadata.markers || [])];
+      this.config.metadata.recommendations = [
+        ...(playerSourceUiConfig.metadata.recommendations || uiconfig.metadata.recommendations || []),
+      ];
```
dweinber, line 286: This means we're ignoring recommendations/markers from uiConfig if the source config contains those. Why?

src/ts/UIManager.ts
```diff
@@ -65,6 +65,32 @@ export interface InternalUIConfig extends UIConfig {
...
+
+  /**
+   * Removes a recommendation by reference and returns `true` if the recommendation has
+   * been part of the recommendations and successfully removed, or `false` if the recommendation
+   * could not be found and thus not removed.
+   */
+  remove(recommendation: RecommendationConfig): boolean;
+
+  /**
+   * Returns the list of all added recommendations in display order.
```
dweinber: For timeline markers we're stating `in insertion order`, which makes sense. I think it's not necessarily clear  here what the display order is. I'd suggest changing to `in insertion order` here as well.

Further, I think we should enhance this by defining the order in relation to recommendations scheduled via config/source/dynamic API (with config/source ones first and dynamic ones after those)

src/ts/UIManager.ts
dweinber: Clarified offline, let keep it as it is now to not change existing behavior and keep behavior consistent across recommendations and timeline markers

stonko1994: [9877a52](https://github.com/bitmovin/bitmovin-player-ui/pull/878/commits/9877a52ca42292ebc0d73a8b3acbe7610eafaff2)

src/ts/UIManager.ts
stonko1994: [9877a52](https://github.com/bitmovin/bitmovin-player-ui/pull/878/commits/9877a52ca42292ebc0d73a8b3acbe7610eafaff2)

dweinber (approved):

dweinber:
Minor documentation point, good to merge after that

PR #882, opened 2026-05-22
Add custom context menu component

stonko1994:
## Description

Adds a desktop-only player context menu to the default UI layouts. The menu currently shows Bitmovin Player and UI version information and is built on a reusable `ContextMenu` component so future context-menu content can share the same behavior.

### Expected user behavior

- Right-clicking inside the active player UI opens the custom context menu.
- A second right click while the custom context menu is already open is left to the browser, so the native context menu opens again. If the second right click happens outside the custom menu, the custom menu closes first. If it happens inside the custom menu, the custom menu stays open so browser actions like Inspect Element can target the custom menu itself.
- Clicking anywhere outside the custom context menu closes it. Clicking inside the custom menu keeps it open. This prevents stale menus from staying visible after normal page/player interaction while still allowing interaction with the menu content.

### Changes

- Added `UIInstanceManager.onActive` and `onInactive` events so components can attach document-level behavior only while their UI variant is active and cleanly detach it when the active UI changes or the UI is released.
- Added a reusable `ContextMenu` component that attaches document `contextmenu`, `click`, and `keydown` handlers only while active.
- `ContextMenu` subclasses `SettingsPanel` to reuse its explicit page structure, state reset behavior, hide delay handling, and existing settings-panel styling.
  - `ContextMenu` extends `SettingsPanel` so it can reuse the existing panel paging behavior, hide/reset behavior, and styling instead of duplicating those primitives.
- The context menu DOM element is attached lazily to the active root UI element right before opening and removed again after the hide transition. This keeps the menu inside the same styled root as the UI, including Shadow DOM support.
- Positioning uses pointer coordinates relative to the active root UI element with `position: absolute`, so the menu scrolls with the UI/page instead of sticking to the viewport. The coordinates are clamped to the UI bounds to keep the menu reachable near the right and bottom edges.
- Added a player-specific `PlayerContextMenu` subclass and wired it into the default main and small-screen desktop layouts.

- [x] `CHANGELOG` entry

### DIFF
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -246,6 +246,9 @@
         playback: {
           muted: true,
         },
+        logs: {
+          level: 'warn',
+        },
       };
 
       var uiConfig = {
--- a/src/ts/DOM.ts
+++ b/src/ts/DOM.ts
@@ -529,6 +529,10 @@ export class DOM {
       const transitionHandler = (event: Event) => {
         const transitionEvent = event as TransitionEvent;
 
+        if (transitionEvent.target !== element) {
+          return;
+        }
+
         if (transitionEvent.propertyName !== propertyName) {
           return;
         }
--- a/src/ts/DOM.ts
+++ b/src/ts/DOM.ts
@@ -512,6 +512,38 @@ export class DOM {
   }
 
   /**
+   * Resolves when the next CSS transition on the first element finishes or is canceled.
+   * A transition property is required so unrelated or bubbling transitions do not resolve the promise too early.
+   * @param propertyName CSS transition property to wait for
+   * @returns {Promise<void>}
+   */
+  async waitForTransitionEnd(propertyName: string): Promise<void> {
+    const element = this.get(0);
+    const hasTransition = element && getComputedStyle(element).transitionProperty !== 'none';
+
+    if (!hasTransition) {
+      return;
+    }
+
+    return new Promise<void>(resolve => {
+      const transitionHandler = (event: Event) => {
+        const transitionEvent = event as TransitionEvent;
+
+        if (transitionEvent.propertyName !== propertyName) {
+          return;
+        }
+
+        this.off('transitionend', transitionHandler);
+        this.off('transitioncancel', transitionHandler);
+        resolve();
+      };
+
+      this.on('transitionend', transitionHandler);
+      this.on('transitioncancel', transitionHandler);
+    });
+  }
+
+  /**
    * Adds the specified class(es) to all elements.
    * @param className the class(es) to add, multiple classes separated by space
    * @returns {DOM}
--- a/src/ts/UIManager.ts
+++ b/src/ts/UIManager.ts
@@ -491,6 +491,7 @@ export class UIManager {
 
     // Hide the currently active UI variant
     if (this.currentUi) {
+      this.currentUi.onInactive.dispatch(this.currentUi.getUI());
       this.currentUi.getUI().hide();
     }
 
@@ -514,6 +515,7 @@ export class UIManager {
       onShow();
     }
     this.currentUi.getUI().show();
+    this.currentUi.onActive.dispatch(this.currentUi.getUI());
     this.events.onActiveUiChanged.dispatch(this, { previousUi, currentUi: nextUi });
   }
 
@@ -619,6 +621,11 @@ export class UIManager {
   release(): void {
     this.config.adBreakTracker.release();
 
+    if (this.currentUi) {
+      this.currentUi.onInactive.dispatch(this.currentUi.getUI());
+      this.currentUi = null;
+    }
+
     for (const uiInstanceManager of this.uiInstanceManagers) {
       this.releaseUi(uiInstanceManager);
     }
@@ -706,6 +713,8 @@ export class UIInstanceManager {
 
   private events = {
     onConfigured: new EventDispatcher<UIContainer, NoArgs>(),
+    onActive: new EventDispatcher<UIContainer, NoArgs>(),
+    onInactive: new EventDispatcher<UIContainer, NoArgs>(),
     onSeek: new EventDispatcher<SeekBar, NoArgs>(),
     onSeekPreview: new EventDispatcher<SeekBar, SeekPreviewArgs>(),
     onSeeked: new EventDispatcher<SeekBar, NoArgs>(),
@@ -758,6 +767,22 @@ export class UIInstanceManager {
    */
   get onConfigured(): EventDispatcher<UIContainer, NoArgs> {
     return this.events.onConfigured;
+  }
+
+  /**
+   * Fires when this UI instance becomes the active UI variant.
+   * @returns {EventDispatcher}
+   */
+  get onActive(): EventDispatcher<UIContainer, NoArgs> {
+    return this.events.onActive;
+  }
+
+  /**
+   * Fires when this UI instance stops being the active UI variant.
+   * @returns {EventDispatcher}
+   */
+  get onInactive(): EventDispatcher<UIContainer, NoArgs> {
+    return this.events.onInactive;
   }
 
   /**
--- a/src/ts/components/contextmenu/ContextMenu.ts
+++ b/src/ts/components/contextmenu/ContextMenu.ts
@@ -1,4 +1,4 @@
-import { Container, ContainerConfig } from '../Container';
+import { SettingsPanel, SettingsPanelConfig } from '../settings/SettingsPanel';
 import { UIInstanceManager } from '../../UIManager';
 import { PlayerAPI } from 'bitmovin-player';
 
@@ -7,7 +7,7 @@ import { PlayerAPI } from 'bitmovin-player';
  *
  * @category Configs
  */
-export interface ContextMenuConfig extends ContainerConfig {}
+export interface ContextMenuConfig extends SettingsPanelConfig {}
 
 /**
  * A floating context menu shown at the pointer position when the user opens the
@@ -15,7 +15,7 @@ export interface ContextMenuConfig extends ContainerConfig {}
  *
  * @category Components
  */
-export class ContextMenu<Config extends ContextMenuConfig = ContextMenuConfig> extends Container<Config> {
+export class ContextMenu<Config extends ContextMenuConfig = ContextMenuConfig> extends SettingsPanel<Config> {
   constructor(config: Config = {} as Config) {
     super(config);
 
@@ -23,6 +23,7 @@ export class ContextMenu<Config extends ContextMenuConfig = ContextMenuConfig> e
       config,
       {
         cssClass: 'ui-context-menu',
+        hideDelay: -1,
         hidden: true,
         role: 'menu',
       } as Config,
@@ -52,7 +53,7 @@ export class ContextMenu<Config extends ContextMenuConfig = ContextMenuConfig> e
 
       if (this.isShown()) {
         if (!this.isEventTargetInsideContextMenu(event)) {
-          this.hide();
+          this.hideAndReset();
         }
         // Let the browser handle a second contextmenu event while the custom menu is open,
         // so users can still access native actions like Inspect Element.
@@ -69,13 +70,13 @@ export class ContextMenu<Config extends ContextMenuConfig = ContextMenuConfig> e
 
     const documentKeyDownHandler = (event: KeyboardEvent) => {
       if (event.key === 'Escape' && this.isShown()) {
-        this.hide();
+        this.hideAndReset();
       }
     };
 
     const documentClickHandler = (event: MouseEvent) => {
       if (this.isShown() && !this.isEventTargetInsideContextMenu(event)) {
-        this.hide();
+        this.hideAndReset();
       }
     };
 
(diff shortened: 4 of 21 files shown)


src/ts/DOM.ts
```diff
@@ -511,6 +511,38 @@ export class DOM {
...
+        }
+
+        this.off('transitionend', transitionHandler);
+        this.off('transitioncancel', transitionHandler);
+        resolve();
+      };
+
+      this.on('transitionend', transitionHandler);
+      this.on('transitioncancel', transitionHandler);
+    });
```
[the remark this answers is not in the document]
stonko1994: [4ce1e07](https://github.com/bitmovin/bitmovin-player-ui/pull/882/commits/4ce1e0746d44a9653e23fefe7a03c7906f3286f9)

src/ts/components/contextmenu/ContextMenu.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  private attachContextMenuElement(): void {
+    const rootElement = this.getDomElement().get(0) as HTMLElement;
+
+    if (rootElement.parentElement === document.body) {
+      return;
+    }
+
+    document.body.appendChild(rootElement);
```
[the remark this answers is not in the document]
stonko1994: [09bd0a0](https://github.com/bitmovin/bitmovin-player-ui/pull/882/commits/09bd0a089af90ac2945d32fb36e8aff787b22b72)

src/ts/components/contextmenu/PlayerContextMenu.ts
```diff
@@ -0,0 +1,62 @@
+import { PlayerAPI } from 'bitmovin-player';
+import { i18n } from '../../localization/i18n';
+import { version as UI_VERSION_RAW } from '../../main';
+import { Label, LabelConfig } from '../labels/Label';
+import { UIInstanceManager } from '../../UIManager';
+import { ContextMenu, ContextMenuConfig } from './ContextMenu';
+
+// `version` in `main.ts` carries the JSON-stringified package version (i.e. surrounded
+// by quotes from the build-time replacement). Peel them off for display.
+const UI_VERSION: string = UI_VERSION_RAW.replace(/^"|"$/g, '');
```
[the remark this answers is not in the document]
stonko1994: [6397e77](https://github.com/bitmovin/bitmovin-player-ui/pull/882/commits/6397e772417e18abee4b6303e40c9731739c519d)

src/html/index.html
```diff
@@ -246,6 +246,9 @@ <h4 class="card-title">Stats</h4>
         playback: {
           muted: true,
         },
+        logs: {
+          level: 'warn',
+        },
```
stonko1994, line 251: Sneaked this as otherwise the player is swallowing all exceptions from an event listener.

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -906,21 +906,3 @@ export class SubtitleRegionContainer extends Container<ContainerConfig> {
 function isCea608SubtitleCue(cue: SubtitleCueEvent): boolean {
   return cue.position != null;
 }
-
-function awaitTransitionEnd(domElement: DOM) {
```
stonko1994, line 910: Extracted into `DOM.ts`.

src/ts/components/settings/SettingsPanel.ts
```diff
@@ -36,6 +36,12 @@ export interface SettingsPanelConfig extends ContainerConfig {
    * Default: 5 seconds (5000)
    */
   stateResetDelay?: number;
+
+  /**
+   * Specifies if the settings panel should hide when the UI controls hide.
+   * Default: true
+   */
+  hideOnControlsHide?: boolean;
```
stonko1994, line 44: Needed to prevent the `ContextMenu` from hiding when the mouse leaves the UI

src/ts/UIManager.ts
```diff
@@ -760,6 +769,22 @@ export class UIInstanceManager {
...
+  }
+
+  /**
+   * Fires when this UI instance stops being the active UI variant.
+   * @returns {EventDispatcher}
+   */
+  get onInactive(): EventDispatcher<UIContainer, NoArgs> {
+    return this.events.onInactive;
+  }
+
```
stonko1994, line 892: Two new events to improve live-cycle events within UI variants such as adding/removing listeners accordingly when a UI variant get's active/inactive. Used by the `ContextMenu` for the `contextmenu` listener.

src/ts/components/contextmenu/ContextMenu.ts
```diff
@@ -0,0 +1,155 @@
...
+  private contextMenuHost: HTMLElement;
+
+  constructor(config: Config = {} as Config) {
+    super(config);
+
+    this.config = this.mergeConfig(
+      config,
+      {
+        cssClass: 'ui-context-menu',
+        hideDelay: -1,
```
123mpozzi, line 28: This skips the `hideTImeout` setup from `SettingsPanel.configure()`.

That can be dangerous since there are some APIs assuming it exists: `suspendHideTimeout()` - in general it looks like the class assumes the timeout to exist - what's the reason of putting it `-1`?

It also can crash on UI updates since on `configure()` it still subscribes to UI updates, and calls `trackComponentViewMode` which uses timeout-related APIs

src/ts/components/contextmenu/ContextMenu.ts
```diff
@@ -0,0 +1,155 @@
...
+ * A floating context menu shown at the pointer position when the user opens the
+ * browser context menu inside the player UI.
+ *
+ * @category Components
+ */
+export class ContextMenu<Config extends ContextMenuConfig = ContextMenuConfig> extends SettingsPanel<Config> {
+  private contextMenuHost: HTMLElement;
+
+  constructor(config: Config = {} as Config) {
+    super(config);
```
123mpozzi, line 22: Here we accept an empty/default config, but `SettingsPanel` requires at least one `SettingsPanelPage`. Without one, `getRootPage()` returns `undefined`.

This is dangerous because it sets `activePage` to `undefined` on initialize, and later on the `SettingsPanel.configure()` method assumes a page exists  - I think it could crash in this case

src/ts/components/contextmenu/ContextMenu.ts
```diff
@@ -0,0 +1,155 @@
+import { SettingsPanel, SettingsPanelConfig } from '../settings/SettingsPanel';
+import { UIInstanceManager } from '../../UIManager';
+import { PlayerAPI } from 'bitmovin-player';
+
+/**
+ * Configuration interface for a generic {@link ContextMenu}.
+ *
+ * @category Configs
+ */
+export interface ContextMenuConfig extends SettingsPanelConfig {}
```
123mpozzi, line 10: nit: I see no specs, can we add some? E.g.  at least:
- `new ContextMenu({ components: [...] })` does not crash/throw
- a `ContextMenu` with `hideDelay: -1` does not crash/throw when `onComponentViewModeChanged` is dispatched

src/ts/components/contextmenu/ContextMenu.ts
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * A floating context menu shown at the pointer position when the user opens the
+ * browser context menu inside the player UI.
+ *
+ * @category Components
+ */
+export class ContextMenu<Config extends ContextMenuConfig = ContextMenuConfig> extends SettingsPanel<Config> {
```
123mpozzi, line 18: `ContextMenuConfig` extends `SettingsPanelConfig`, which allows:

```ts
  new ContextMenu({
    components: [new SomeMenuItem()]
  })
```

But because `ContextMenu` extends `SettingsPanel`, those components are not treated as menu content unless they are `SettingsPanelPage` instances - `SettingsPanel` expects pages, not arbitrary children.

`PlayerContextMenu` avoids this because it explicitly wraps its content:

```ts
  components: [
    new SettingsPanelPage({
      components: [...]
    }),
  ]
```

but here we're uncovered.
How about wrapping the content automatically? E.g.:

```ts
    super({
      ...config,
      components: [
        new SettingsPanelPage({
          components,
        }),
      ],
    });
```

src/ts/DOM.ts
```diff
@@ -511,6 +511,42 @@ export class DOM {
     return this;
   }
 
+  /**
+   * Resolves when the next CSS transition on the first element finishes or is canceled.
+   * A transition property is required so unrelated or bubbling transitions do not resolve the promise too early.
+   * @param propertyName CSS transition property to wait for
+   * @returns {Promise<void>}
+   */
+  async waitForTransitionEnd(propertyName: string): Promise<void> {
```
123mpozzi, line 520: nit (since it's not a regression):

I see that currently this returns a `Promise` that only resolves when a `transitionend` or `transitioncancel` event arrives for the exact property.

Can we add a timeout here so that it cannot hang forever if the requested property never emits a transition event?

src/ts/components/contextmenu/ContextMenu.ts
```diff
@@ -0,0 +1,155 @@
...
+    super(config);
+
+    this.config = this.mergeConfig(
+      config,
+      {
+        cssClass: 'ui-context-menu',
+        hideDelay: -1,
+        hideOnControlsHide: false,
+        hidden: true,
+        role: 'menu',
```
123mpozzi, line 31: nit: we are declaring stronger accessibility semantics than the component actually implements.

The overall interaction is mouse-driven, which diverges from the [expected usage patterns](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menu_role). We do not:
- focus the menu on open,
- implement arrow-key navigation
- and the default `PlayerContextMenu` is informational, not an action menu (which e.g. would have actions instead, like 'copy video URL' , 'see stats', 'open settings', ...)

src/ts/UIManager.ts
123mpozzi: That's great! Should we add them to the changelog? E.g.:

```md
- Added `UIInstanceManager.onActive` and `onInactive` lifecycle events for UI variant activation changes.
```

123mpozzi (commented):
Nice addition! 💪 

Manual testing looks good, the menu is well-positioned inside of the player bounds and behaves as expected (tested right-click, native menu, left click, and some interactions with settings pages - safari and chrome).

I've just left some comments, most are inheritance concerns.

src/ts/components/contextmenu/ContextMenu.ts
stonko1994: > what's the reason of putting it -1?
The idea is that the ContextMenu never hides itself automatically.

I checked the current state and you are correct but this is nothing new. I'd treat that as out of scope of this PR but I will create a mini fix PR for this.

src/ts/components/contextmenu/ContextMenu.ts
stonko1994: Yes that is correct. Again nothing new. Was already like this in the `SettingsPanel`.

src/ts/components/contextmenu/ContextMenu.ts
stonko1994: I'm not a fan of 'does not crash' tests. But I'll see which tests we can add that make sense.

src/ts/components/contextmenu/ContextMenu.ts
stonko1994: Was already like this in the SettingsPanel API. I'll see if we can improve that.

src/ts/components/contextmenu/ContextMenu.ts
stonko1994: While this is correct, is it possible to open a context menu with the keyboard? I don't see that as a huge requirement to be honest.

src/ts/DOM.ts
stonko1994: I think this API should only be used surgically when needed. In addition, the method returns early if the property does not have a transition configured, so we should be fairly safe.

src/ts/components/contextmenu/ContextMenu.ts
123mpozzi: Yeah,  that's a good point, not sure which use-cases the keyboard would have

src/ts/DOM.ts
123mpozzi: I think the dangerous part is that the guard `transitionProperty !== 'none'` only proves some transition is configured. So a call can pass the guard and then wait for an event that never comes (in case we wait for a different property)

E.g. the element has transition: `opacity 0.3s` configured, but the caller does `waitForTransitionEnd('bottom')`. The guard passes (a transition exists), but no `transitionend` for bottom ever fires, so the promise hangs forever.

FIne to keep it as is for now though, since it was already the case

src/ts/UIManager.ts
123mpozzi: Any update on this?

src/ts/components/contextmenu/ContextMenu.ts
123mpozzi: Agree. For a trivial fix, we could apply a: `if (this.hideTimeout)` guard in suspend/resume in `SettingsPanel`

123mpozzi (approved):

src/ts/UIManager.ts
stonko1994: [60f16e7](https://github.com/bitmovin/bitmovin-player-ui/pull/882/commits/60f16e7be7aa72bbb064dc5c737d5c44da6f4dfe)

src/ts/components/contextmenu/ContextMenu.ts
123mpozzi: Following up in #899

PR #885, opened 2026-05-22
Add player insights overlay

stonko1994:
## Description
Adds a Player Insights panel to the UI which gives insights into the player.

Screenshot 2026-06-08 at 12 46 59

### Changes
The Player Insights panel can be opened via a new context menu item. Once open, it updates its information based on player events or the refresh timer. The panel will not automatically hide and will stay visible until the user actively closes it.

**Current included insights:**
- Video: playback and downloaded video quality, including resolution, frame rate, bitrate, and codec when available.
- Viewport / Frames: video element viewport size and dropped video frame count.
- Audio: playback and downloaded audio quality, including bitrate and codec when available.
- Buffer Video / Audio: forward video and audio buffer levels.
- Time: current playback time and duration for VOD, or live wall-clock time plus time shift for live streams.
- Stream: stream type and player type.
- Player version: active player version.

Rows without a current value are hidden until the player exposes the corresponding insight.

### Limitations
Support for the small screen UI layout (including mobile SDKs) will be added on a follow-up PR.

- [x] `CHANGELOG` entry


src/ts/components/contextmenu/ContextMenu.ts
```diff
@@ -152,4 +179,24 @@ export class ContextMenu<Config extends ContextMenuConfig = ContextMenuConfig> e
   private attachContextMenuElement(): void {
     this.contextMenuHost.appendChild(this.getDomElement().get(0));
   }
+
+  private setContextMenuForComponent(component: Component<ComponentConfig>): void {
+    if (component instanceof InteractiveContextMenuItem) {
+      component.setContextMenu(this);
+    }
```
stonko1994, line 186: This is convenience to 'automatically' register the parent `ContextMenu` on its items. Otherwise this is required when constructing the Items (similar to the `SettingsPanel` in the UIFactory).

src/html/index.html
```diff
@@ -173,7 +173,7 @@ <h4 class="card-title">Stats</h4>
           ],
         },
         basicHls: {
-          hls: '//cdn.bitmovin.com/content/assets/sintel/hls/playlist.m3u8',
+          hls: 'https://cdn.bitmovin.com/content/assets/sintel/hls/playlist.m3u8',
```
stonko1994, line 176: Sneak to enable using it on localhost.

src/ts/utils/PlayerInsightsProvider.ts
```diff
@@ -0,0 +1,192 @@
...
+    viewportFrames: new PlayerInsightProperty('Viewport / Frames', player =>
+      PlayerInsightsUtils.formatViewportFramesInsight(player),
+    ),
+    audio: new PlayerInsightProperty('Audio', player => PlayerInsightsUtils.formatAudioQualityInsight(player)),
+    bufferVideoAudio: new PlayerInsightProperty('Buffer Video / Audio', player =>
+      PlayerInsightsUtils.formatBufferInsight(player),
+    ),
+    time: new PlayerInsightProperty('Time', player => PlayerInsightsUtils.formatTimeInsight(player)),
+    stream: new PlayerInsightProperty('Stream', player => PlayerInsightsUtils.formatStreamInsight(player)),
+    player: new PlayerInsightProperty('Player version', player => player.version),
```
stonko1994, line 41: Texts here are intentionally not localized. The player insights are only displayed in English.

src/ts/localization/languages/en.json
```diff
@@ -12,6 +12,8 @@
   "settings.subtitles.background.color": "Background color",
   "settings.subtitles.background.opacity": "Background opacity",
   "settings.subtitles.options": "Options",
+  "playerInsights.show": "Show player insights",
+  "playerInsights.hide": "Hide player insights",
```
123mpozzi: Nit: in other locals, we have capital letters for 'Player' and 'Insights'

src/ts/localization/languages/de.json
```diff
@@ -81,6 +83,8 @@
   "seekBar.durationText": "aus",
   "quickseek.forward": "{seekSeconds} Sekunden Vor",
   "quickseek.rewind": "{seekSeconds} Sekunden Zurück",
+  "contextMenu.title": "Bitmovin Player",
+  "contextMenu.subtitle": "Adaptive Streaming für das Web",
```
123mpozzi, line 85: nit: this is capitalized differently on other locales, each word starts lowercase

src/ts/components/settings/SettingsPanel.ts
```diff
@@ -42,6 +42,12 @@ export interface SettingsPanelConfig extends ContainerConfig {
    * Default: true
    */
   hideOnControlsHide?: boolean;
+
+  /**
+   * Specifies if the settings panel should be hidden when another settings panel is opened.
+   * Default: true
+   */
+  hideWithOtherSettingsPanels?: boolean;
```
123mpozzi: Nice change! I like the configurability instead of an hardcoded behavior

nit: not if the naming is clear enough, it could be mistook as 'hide together with other settings panel'. How about `hideOnOtherSettingsPanelOpening` or so?

src/scss/components/panels/_player-insights-panel.scss
```diff
@@ -0,0 +1,122 @@
...
+.#{$prefix}-ui-uicontainer.#{$prefix}-layout-max-width-400 {
+  .#{$prefix}-ui-player-insights-panel-item {
+    > .#{$prefix}-container-wrapper {
+      column-gap: 0.4em;
+      grid-template-columns: minmax(0, 1fr) minmax(0, 2fr);
+    }
+  }
+
+  .#{$prefix}-ui-player-insights-panel-item-leading-label {
+    font-size: calc($font-size-small * 0.85);
```
123mpozzi: I am not a fan of introducing a new font-size outside of `variables` file, since those tends to stay out-of-date and create inconsistencies.
Can't we add a new constant value there? E.g.:
- `$font-size-xsmall`
- or behavioral: `$font-size-paragraph`

src/ts/utils/PlayerInsightsProvider.ts
```diff
@@ -0,0 +1,192 @@
...
+      snapshots.push(this.properties[propertyName].getSnapshot(player));
+    }
+
+    return snapshots;
+  }
+
+  private startTimer(): void {
+    this.stopTimer();
+
+    if (!this.playerApi || !this.active || this.refreshIntervalMs === -1 || this.playerApi.isPaused()) {
```
123mpozzi: nit: what happens if refresh interval is `0` ? 

```suggestion
    if (!this.playerApi || !this.active || this.refreshIntervalMs < =0 || this.playerApi.isPaused()) {
```

src/ts/utils/PlayerInsightsUtils.ts
```diff
@@ -0,0 +1,238 @@
...
+    if (!playbackValue) {
+      return downloadedValue || null;
+    }
+    if (!downloadedValue || downloadVideoQualityInsight.id === playbackVideoQualityInsight.id) {
+      return playbackValue;
+    }
+
+    // Only show playback and downloaded quality separately if they differ,
+    // otherwise just show one value to avoid redundancy
+    return `${playbackValue} / \u2193${downloadedValue}`;
```
123mpozzi: nit: can we extract `\u2193` semantically into a var? E.g. `transitionCharacter`

src/ts/utils/PlayerInsightsProvider.ts
```diff
@@ -0,0 +1,192 @@
+import type { PlayerAPI } from 'bitmovin-player';
```
123mpozzi, line 1: Nice usages of `null` to distinguish explicitly unset values from undefined 💪

src/ts/utils/PlayerInsightsProvider.ts
```diff
@@ -0,0 +1,192 @@
...
+    return this.events.onChanged.getEvent();
+  }
+
+  private readonly updateAndStartTimer = (): void => {
+    this.dispatchChanged();
+    this.startTimer();
+  };
+
+  private readonly dispatchChanged = (): void => {
+    if (!this.playerApi) {
```
123mpozzi: I don't think we need to keep data fresh while the panel is not active. No need to unsubscribe, we could just guard the updates:

```suggestion
    if (!this.active || !this.playerApi) {
```

We already fetch new data once back active:

```diff
  activate(): void {
    this.active = true;
+    this.dispatchChanged(); // opening the panel immediately refreshes the visible rows
    this.startTimer();
  }
```

src/ts/utils/PlayerInsightsUtils.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    if (!playbackValue) {
+      return downloadedValue || null;
+    }
+    if (!downloadedValue || downloadAudioQualityInsight.id === playbackAudioQualityInsight.id) {
+      return playbackValue;
+    }
+
+    // Only show playback and downloaded quality separately if they differ,
+    // otherwise just show one value to avoid redundancy
```
123mpozzi: nit: comment can be clearer, here they already have different values

```suggestion
    // Show both values when playback and downloaded renditions differ.
```

src/ts/utils/PlayerInsightsProvider.ts
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  initialize(player: PlayerAPI, uimanager: UIInstanceManager, refreshIntervalMs: number): void {
+    this.playerApi = player;
+    this.uimanager = uimanager;
+    this.refreshIntervalMs = refreshIntervalMs;
+
+    player.on(player.exports.PlayerEvent.Play, this.updateAndStartTimer);
+    player.on(player.exports.PlayerEvent.Playing, this.updateAndStartTimer);
+    player.on(player.exports.PlayerEvent.Paused, this.dispatchChanged);
```
123mpozzi, line 71: We could stop the repeating refresh timer when playback pauses:

```ts
  private readonly updateAndStopTimer = (): void => {
    this.dispatchChanged();
    this.stopTimer();
  };
```

And wire it to also `PlaybackFinished`, while keeping Destroy the only one using `this.stopTimer`:

```ts
  player.on(player.exports.PlayerEvent.Paused, this.updateAndStopTimer);
  player.on(player.exports.PlayerEvent.PlaybackFinished, this.updateAndStopTimer);
```

src/ts/UIFactory.ts
```diff
@@ -367,9 +369,10 @@ export namespace UIFactory {
 
     export function smallScreen(): UIContainer {
       const subtitleOverlay = new SubtitleOverlay();
-      const playerContextMenu = BrowserUtils.isMobile ? null : new PlayerContextMenu();
+      const playerInsightsPanel = BrowserUtils.isMobile ? null : new PlayerInsightsPanel({ hidden: true });
+      const playerContextMenu = playerInsightsPanel ? new PlayerContextMenu({ playerInsightsPanel }) : null;
 
-      const settingsPanel = buildDefaultSettingsPanel(subtitleOverlay, -1);
+      const settingsPanel = buildDefaultSettingsPanel(subtitleOverlay, -1, false);
```
123mpozzi: Is the explicit disabling of eco mode intentional?

src/ts/utils/PlayerInsightsUtils.ts
```diff
@@ -0,0 +1,238 @@
...
+    return `${playbackValue} / \u2193${downloadedValue}`;
+  }
+
+  /** Formats the currently playing and downloaded audio quality details for the panel row. */
+  export function formatAudioQualityInsight(player: PlayerAPI): string | null {
```
123mpozzi, line 31: nit: Behavior is the same as previous method and could be shared:

```ts
  export function formatAudioQualityInsight(player: PlayerAPI): string | null { // same for formatVideoQualityInsight
    return formatPlaybackAndDownloadedQuality(
      player.getPlaybackAudioData() as QualityInsight,
      enrichDownloadedAudioData(player.getDownloadedAudioData(), player.getAvailableAudioQualities()),
    );
  }
```

```ts
function formatPlaybackAndDownloadedQuality(
  playbackQuality: QualityInsight | undefined,
  downloadedQuality: QualityInsight | undefined,
): string | null {
  const playbackValue = formatQualityInsight(playbackQuality);
  const downloadedValue = formatQualityInsight(downloadedQuality);

  if (!playbackValue) {
    return downloadedValue || null;
  }

  if (!downloadedValue || haveSameQualityId(playbackQuality, downloadedQuality) || playbackValue === downloadedValue) {
    return playbackValue;
  }

  return `${playbackValue} / \u2193${downloadedValue}`;
}

function haveSameQualityId(
  playbackQuality: QualityInsight | undefined,
  downloadedQuality: QualityInsight | undefined,
): boolean {
  return (
    playbackQuality?.id !== undefined &&
    downloadedQuality?.id !== undefined &&
    playbackQuality.id === downloadedQuality.id
  );
}
```

And:

```diff
-function formatQualityInsights(quality: QualityInsight): string {
-  const parts = [];
+function formatQualityInsight(quality: QualityInsight | undefined): string {
+  if (!quality) {
+    return '';
+  }

+  const parts: string[] = [];

// ...
```

src/ts/utils/PlayerInsightsUtils.ts
```diff
@@ -0,0 +1,238 @@
...
+    const videoBufferLevel = player.buffer.getLevel(
+      player.exports.BufferType.ForwardDuration,
+      player.exports.MediaType.Video,
+    ).level;
+    const audioBufferLevel = player.buffer.getLevel(
+      player.exports.BufferType.ForwardDuration,
+      player.exports.MediaType.Audio,
+    ).level;
+
+    if (videoBufferLevel == null && videoBufferLevel == null) {
```
123mpozzi: typo:

```suggestion
    if (videoBufferLevel == null && audioBufferLevel == null) {
```

src/ts/utils/PlayerInsightsProvider.ts
```diff
@@ -0,0 +1,192 @@
...
+import { LocalizableText } from '../localization/i18n';
+import { Timeout } from './Timeout';
+import { PlayerInsightsUtils } from './PlayerInsightsUtils';
+
+export type PlayerInsightValueProvider = (player: PlayerAPI) => LocalizableText | null | undefined;
+
+interface PlayerInsights {
+  video: PlayerInsightProperty;
+  viewportFrames: PlayerInsightProperty;
+  audio: PlayerInsightProperty;
```
123mpozzi, line 13: Do we expose meaningful information on captions, like their format? because if so, we could add them to the list

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - New `ContextMenu` component to show custom context menus on right-click.
```
123mpozzi, line 12: nit: We could mention the missing public types:


```suggestion
- New `ContextMenu`, `PlayerContextMenu`, and `InteractiveContextMenuItem` components for custom and default right-click context menus.
```

src/ts/utils/PlayerInsightsUtils.ts
123mpozzi, line 31: nit: we could expose audio track label so we can identify the active language

src/ts/utils/PlayerInsightsProvider.ts
```diff
@@ -0,0 +1,192 @@
...
+  bufferVideoAudio: PlayerInsightProperty;
+  time: PlayerInsightProperty;
+  stream: PlayerInsightProperty;
+  player: PlayerInsightProperty;
+}
```
123mpozzi, line 18: Were other fields considered? These examples could be worthy:
- Live / seekability: `Live: timeshift -28.4s` / `window 300s` / `seekable 12:04:10-12:09:10`
- Playback state / health: 
```
  State: playing, stalled=false, speed=1x
  Health: 0 dropped / 3.42s stalled
```
- output screen/state: `View: fullscreen, cast=false, airplay=false`
- Ad schedule: `Ad schedule: 3 scheduled (pre, 10:00, post)`
- ABR selection decision: `video 1080p fixed / audio auto` ,we could via:
```ts
  function formatSelectedQuality(selection: VideoQuality | AudioQuality | 'auto'): string {
    if (selection === 'auto') {
      return 'auto';
    }

    return `${selection.id} fixed`;
  }
```
- segment duration
- discontinuity info?

We could also just ask the player teams which ones they deem more significant to debug issues.
Let me know your thoughts 🙌

src/ts/components/contextmenu/PlayerContextMenu.ts
```diff
@@ -4,29 +4,48 @@ import { version as UI_VERSION } from '../../version';
...
  * Configuration interface for the {@link PlayerContextMenu}.
  *
  * @category Configs
  */
-export interface PlayerContextMenuConfig extends ContextMenuConfig {}
+export interface PlayerContextMenuConfig extends ContextMenuConfig {
+  /**
+   * The player insights panel to expose as a default context menu action.
+   */
+  playerInsightsPanel: PlayerInsightsPanel;
```
123mpozzi, line 22: nit: note this is breaking towards the base PR since there `PlayerContextMenu` had no required fields. Should not be an issue since I guess they'll be released in same version?

src/ts/utils/PlayerInsightsProvider.ts
123mpozzi, line 1: Is the file placement under `utils/` intentional?

src/ts/utils/PlayerInsightsUtils.ts
```diff
@@ -0,0 +1,238 @@
...
+    return `${playbackValue} / ${downloadedValue}`;
+  }
+
+  /** Formats the current video element size together with dropped frame count. */
+  export function formatViewportFramesInsight(player: PlayerAPI): string {
+    const videoElement = player.getContainer().querySelector('video') as HTMLVideoElement;
+    const viewport =
+      videoElement.clientWidth > 0 && videoElement.clientHeight > 0
+        ? `${videoElement.clientWidth}x${videoElement.clientHeight}`
+        : null;
```
123mpozzi: This can crash If `<video>` is not defined. We could null-guard it:

```suggestion
    const videoElement = player.getContainer().querySelector('video');
    const viewport =
      videoElement && videoElement.clientWidth > 0 && videoElement.clientHeight > 0
        ? `${videoElement.clientWidth}x${videoElement.clientHeight}`
        : null;
```

And make `formatQualityInsights` tolerate a null/undefined quality so neither the playback nor downloaded path can throw

src/ts/utils/PlayerInsightsUtils.ts
```diff
@@ -0,0 +1,238 @@
+import { AudioQuality, DownloadedAudioData, DownloadedVideoData, PlayerAPI, VideoQuality } from 'bitmovin-player';
```
123mpozzi: nit: we could add some tests since there's conditionals in the functions, to future-proof the expected behavior

src/ts/utils/PlayerInsightsProvider.ts
123mpozzi, line 1: Can we add some behavioral tests?

src/ts/utils/PlayerInsightsUtils.ts
```diff
@@ -0,0 +1,238 @@
...
+}
+
+/** Adds codec data because downloaded audio data only exposes the downloaded rendition identity. */
+function enrichDownloadedAudioData(
+  downloadedAudio: DownloadedAudioData | undefined,
+  availableAudioQualities: AudioQuality[],
+): QualityInsight | undefined {
+  if (!downloadedAudio) {
+    return undefined;
+  }
```
123mpozzi: Is it intentional we don't do this check for downloaded video data?

src/ts/components/contextmenu/PlayerContextMenu.ts
```diff
@@ -4,29 +4,48 @@ import { version as UI_VERSION } from '../../version';
...
+
     super({
       ...config,
       cssClasses: ['ui-player-context-menu', ...(config.cssClasses ?? [])],
       components: [
         new SettingsPanelPage({
-          components: [new PlayerInfoContextMenuItem(), ...(config.components ?? [])],
+          components: [
+            new PlayerInfoContextMenuItem(),
+            ...(actionItems.length > 0 ? [new SettingsPanelSeparator()] : []),
```
123mpozzi: nit: the else path is unreachable, we always start `actionItems` with at least one `PlayerInsightsContextMenuItem`

src/ts/utils/PlayerInsightsUtils.ts
```diff
@@ -0,0 +1,238 @@
...
+    return bufferLevels.join(' / ');
+  }
+
+  /** Formats the current playback time, duration, live latency, and playback speed. */
+  export function formatTimeInsight(player: PlayerAPI): string | null {
+    const isLive = player.isLive();
+
+    if (isLive) {
+      const timeShift = player.getTimeShift();
+      const currentTime = player.getCurrentTime('absolutetime' as TimeMode);
```
123mpozzi: Nit: docs says default is absolute for `player.getCurrentTime()` so we could remove the type-cast

```suggestion
      const currentTime = player.getCurrentTime();
```

src/ts/components/panels/player-insights/PlayerInsightsPanel.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.insightsProvider.release();
+    super.release();
+  }
+
+  private readonly playerInsightsChangedHandler = (
+    _: PlayerInsightsProvider,
+    snapshots: PlayerInsightSnapshot[],
+  ): void => {
+    snapshots.forEach((snapshot, index) => {
+      const item = this.insightItems[index];
```
123mpozzi, line 88: nit: we do `this.insightItems[index]` against `snapshots[index]`, but the only thing guaranteeing row N's label lines up with row N's value is that both for...in loops happen to visit keys in the same order, which is fine today, frail implicit behavior.

We could key snapshot items to ensure the ordering.

src/ts/utils/PlayerInsightsUtils.ts
```diff
@@ -0,0 +1,238 @@
...
+/** Formats resolution and optional frame rate for video quality display. */
+function formatResolution(quality: QualityInsight): string | null {
+  return quality.width && quality.height
+    ? `${quality.width}x${quality.height}${quality.frameRate ? `@${quality.frameRate}` : ''}`
+    : null;
+}
+
+/** Formats bitrate in Mbps or kbps depending on its size. */
+function formatBitrate(bitrate: number | undefined): string {
+  if (!bitrate || !isFinite(bitrate)) {
```
123mpozzi, line 174: Nit: we can guard against negative bitrate (not sure if negative value can happen)

```suggestion
  if (!bitrate || !isFinite(bitrate) || bitrate <= 0) {
```

src/ts/utils/PlayerInsightsUtils.ts
```diff
@@ -0,0 +1,238 @@
...
+    return '';
+  }
+
+  if (bitrate >= 1_000_000) {
```
123mpozzi: Nit: magic numbers could be extracted:

```ts
  const BITS_PER_MBPS = 1_000_000;
  const BITS_PER_KBPS = 1_000;
```

123mpozzi:
Manual testing looks good btw! 🚀 Tested on Safari and Chrome

src/html/index.html
stonko1994: Addressed in `84274b5e24a6feb3fe0388f0b384860868d610b0`.

src/ts/localization/languages/en.json
stonko1994: Addressed in `82a0b1b948afe8086d2441e9d94f6870d44e6868`.

src/ts/localization/languages/de.json
stonko1994: Addressed in `70c5129f1e236d7a9a2693864f71b590ba2524e9` by removing the duplicated non-English label overrides.

src/ts/components/settings/SettingsPanel.ts
stonko1994: Addressed in `3a9e26b6a68a453734287a803d87473e571ec0b3`.

src/scss/components/panels/_player-insights-panel.scss
stonko1994: Addressed in `0edf9102786e8b87e692f0359ccb9f9a03b90923`.

src/ts/utils/PlayerInsightsProvider.ts
stonko1994: No change: the other delay-style UI configs use `-1` as the disabled value, so I kept `0` available instead of treating it as disabled here.

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: Addressed in `63820caa21bade20de9b3c4bd379d8639cc5aa75`.

src/ts/utils/PlayerInsightsProvider.ts
stonko1994: Addressed in `5a00d85e285b16c86a6c91be0819328e1e96bd95`.

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: Addressed in `abe725d18486441b6b53db57d48bcc5dd7bfcc5a`.

src/ts/utils/PlayerInsightsProvider.ts
stonko1994: No change: values can still change while playback is paused, for example live timeShift can drift and downloaded quality can change, so the panel should still refresh while visible.

src/ts/UIFactory.ts
stonko1994: Addressed in `e67acfd12ebe402d7e0dbdbc2176ebacdef96fe6`.

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: Addressed in `ade774e608f5d5ef1082251c2f6f7020d7c4e74d`; follow-up simplification in `2f38b46f72c5595ac9e82a392028f9a43729e74b`.

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: Addressed in `9c79dcf1ce080ffbd5e2e534b42ede1202b02609`.

src/ts/utils/PlayerInsightsProvider.ts
stonko1994: No change: deferred as a possible extension. The initial panel stays focused on the current default set instead of adding every possible sub-feature.

CHANGELOG.md
stonko1994: No change: these types are supporting pieces of the context menu feature, and we do not always list every new supporting type in the changelog.

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: No change: deferred as an extension option. The initial panel keeps the audio row basic for now.

src/ts/utils/PlayerInsightsProvider.ts
stonko1994: No change: out of scope for this PR. These can be added later if we decide to expand the panel diagnostics.

src/ts/components/contextmenu/PlayerContextMenu.ts
stonko1994: No change: intentional. `PlayerContextMenu` is our default context menu implementation and can require the paired player insights panel, similar to other default UI components that require their matching panel.

src/ts/utils/PlayerInsightsProvider.ts
stonko1994: No change: intentional placement under `utils` because this is not a UI component.

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: Addressed the viewport part in `2c137ac9c7be21c35a5e9a85988aec5b041fb164` by using the dedicated player video element API. Missing downloaded quality handling was addressed in `552a07fd61d61b57d93c5097e04ffa15933e3c84` and simplified in `2f38b46f72c5595ac9e82a392028f9a43729e74b`.

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: Addressed in `a73cc632b05a5f71ad24c9c05047fa752215c69d`.

src/ts/utils/PlayerInsightsProvider.ts
stonko1994: No change: skipped for now. The focused utility tests cover the risky conditional formatting paths without adding broader provider test setup.

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: Addressed in `552a07fd61d61b57d93c5097e04ffa15933e3c84`, generalized in `f8327461e25e8a68766e978c88e3f8567d59456d`, and simplified in `2f38b46f72c5595ac9e82a392028f9a43729e74b`.

src/ts/components/contextmenu/PlayerContextMenu.ts
stonko1994: Addressed in `1fb267b32ce2786bcde01138ea5c53cd8e1513ac`.

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: Addressed in `cdd251dc28a300ffaf2f1c5aeda9ad62a6e8bf08`.

src/ts/components/panels/player-insights/PlayerInsightsPanel.ts
stonko1994: No change: current snapshots keep the full row list and use the same provider property order as row creation. Since visibility only hides/shows existing rows, the index mapping is stable for the current design.

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: No change: a negative bitrate would be invalid Player API data, so I would treat that as an upstream player issue instead of silently normalizing it in the UI formatter.

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: Addressed in `6efda6197253bfe03184f02694bbdde2dfb4499c`.

src/ts/utils/PlayerInsightsProvider.ts
123mpozzi: I’m fine with keeping refreshes while paused, but the current behavior seems inconsistent: if the panel is opened while already paused, `startTimer()` bails on `isPaused()` and we only refresh once.

Can we either remove that guard or intentionally stop refreshes on pause?

123mpozzi (commented):
Thanks for addressing the comments! Just https://github.com/bitmovin/bitmovin-player-ui/pull/885#discussion_r3418486243 left

src/ts/utils/PlayerInsightsUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+    return `${(bitrate / BITS_PER_MEGABIT).toFixed(2)}Mbps`;
+  }
+
+  return `${Math.round(bitrate / BITS_PER_KILOBIT)}kbps`;
+}
+
+/** Formats a duration-like value as h:mm:ss or m:ss. */
+function formatSeconds(seconds: number): string {
+  if (!isFinite(seconds)) {
+    return 'Infinity';
```
123mpozzi, line 188: nit: the formatter translates `NaN` as 'Infinity', if we want to be more precise we could put something like '-' or 'unknown' instead

src/ts/utils/PlayerInsightsProvider.ts
stonko1994: Makes sense. Removed the `isPaused` guard [fa691fe](https://github.com/bitmovin/bitmovin-player-ui/pull/885/commits/fa691fea7706086edb886763c5704ed31f0469e1)

src/ts/utils/PlayerInsightsUtils.ts
stonko1994: I think the `NaN` is an edge case which shouldn't happen. The only seconds we format here are currentTime and duration. If they are `NaN` I'd treat that as unexpected value from the Player. Also showing `Infinity` in this case wouldn't be the biggest issue IMO

123mpozzi (approved):

PR #897, opened 2026-06-19
Add flag to treat closed captions like subtitles

LudoMichaud:
## Description
By default, the player UI applies a specific styling to Closed Captions.

I'm proposing a flag that disables this behaviour and treats 608/708 captions like regular subtitles from a styling perspective.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- New `SubtitleOverlayConfig.treatClosedCaptionsAsSubtitles` flag: by default, the player UI applies specific styling to Closed Captions. When enabled, this flag disables that behaviour and treats 608/708 captions like regular VTT subtitles from a styling perspective.
+- New `SubtitleOverlayConfig.enableCea608CaptionPositioning` flag (default `true`): by default, the player UI applies CEA-608 grid-based row/column positioning to Closed Captions. Setting this to `false` disables that positioning so that 608/708 captions are laid out like regular VTT subtitles, while CEA-608-specific text formatting remains independently controlled via `enableCea608CaptionFormatting`.
 
 ### Fixed
 
 - Seeking with the remote on the TV UI no longer gets stuck at the start or end of the seek bar: after scrubbing all the way to one edge, pressing the opposite direction now moves the playback position immediately instead of requiring multiple presses.
 - Settings panels with disabled automatic hiding no longer crash when component view mode changes are dispatched.
-
 
 ## [4.16.0] - 2026-06-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- New `UIConfig.treatClosedCaptionsAsSubtitles` flag: when `true`, CEA-608 closed captions are rendered with the same CSS classes, styling, and DOM structure as regular VTT subtitles, with no CEA-608-specific grid positioning or formatting applied.
+- New `UIConfig.treatClosedCaptionsAsSubtitles` flag: by default, the player UI applies specific styling to Closed Captions. When enabled, this flag disables that behaviour and treats 608/708 captions like regular VTT subtitles from a styling perspective.
 
 ## [4.16.0] - 2026-06-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Added
+
+- New `UIConfig.treatClosedCaptionsAsSubtitles` flag: when `true`, CEA-608 closed captions are rendered with the same CSS classes, styling, and DOM structure as regular VTT subtitles, with no CEA-608-specific grid positioning or formatting applied.
+
 ## [4.16.0] - 2026-06-18
 
 ### Added
--- a/src/ts/components/overlays/SubtitleOverlay.ts
+++ b/src/ts/components/overlays/SubtitleOverlay.ts
@@ -28,13 +28,13 @@ export interface SubtitleOverlayConfig extends ContainerConfig {
    */
   enableCea608CaptionFormatting?: boolean;
   /**
-   * When `true`, closed captions (CEA-608/708) are treated identically to regular subtitles: no CEA-608-specific
-   * CSS classes (`cea608`, `cea608-formatting`), no grid-based row positioning, and no monospaced/uppercase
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,11 +7,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+### Added
+
+- New `SubtitleOverlayConfig.treatClosedCaptionsAsSubtitles` flag: by default, the player UI applies specific styling to Closed Captions. When enabled, this flag disables that behaviour and treats 608/708 captions like regular VTT subtitles from a styling perspective.
+
 ### Fixed
 
 - Seeking with the remote on the TV UI no longer gets stuck at the start or end of the seek bar: after scrubbing all the way to one edge, pressing the opposite direction now moves the playback position immediately instead of requiring multiple presses.
 - Settings panels with disabled automatic hiding no longer crash when component view mode changes are dispatched.
 
+
```
dweinber, line 27: Please remove that additional empty line to keep the layout consistent, thanks!


```suggestion
```

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -56,6 +64,7 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
   private cea608FontSizeFactor = 1;
   private ensureCea608GridSizeUpdated: () => void;
   private cea608SmallPlayerHeightThreshold = SubtitleOverlay.DEFAULT_CEA608_SMALL_PLAYER_HEIGHT_THRESHOLD;
+  private treatClosedCaptionsAsSubtitles = false;
```
dweinber: Please remove this field. It just duplicates the state, and the value is already available via `this.config.treatClosedCaptionsAsSubtitles`. Let's keep the config as the single source of truth.


```suggestion
```

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -27,6 +27,14 @@ export interface SubtitleOverlayConfig extends ContainerConfig {
...
    */
   enableCea608CaptionFormatting?: boolean;
+  /**
+   * When `true`, closed captions (CEA-608/708) are treated identically to regular subtitles: no CEA-608-specific
+   * CSS classes (`cea608`, `cea608-formatting`), no grid-based row positioning, and no monospaced/uppercase
+   * formatting are applied. The captions render with the same styles and DOM structure as VTT subtitles.
+   *
+   * Default: `false`
+   */
+  treatClosedCaptionsAsSubtitles?: boolean;
```
dweinber: Please change this to disable _only_ the CEA-specific positioning. We already have a flag to disable _only_ the CEA-specific formatting, so it would make more sense to have this pair of settings and customers can
- have all enabled (default, same as currently)
- enable positioning, disable formatting
- disable positioning, enable formatting
- disable positioning, disable formatting

With that change, I'd suggest to rename the config option to `enableCea608CaptionPositioning?: boolean` with default being `true`, to have this consistent with the formatting config flag. Please align the documentation then accordingly.

Something like this: 

```suggestion
  /**
   * Controls whether the CEA-608 grid-based row/column positioning is used or not.
   * The formatting (monospaced font, uppercase transform, character letter-spacing) is preserved.
   *
   * Defaults to `true` (CEA-608 grid is used for positioning, matching historical behavior).
   */
  enableCea608CaptionPositioning?: boolean;
```

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -82,6 +91,10 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
       this.cea608SmallPlayerHeightThreshold = uiConfig.cea608SmallPlayerHeightThreshold;
     }
 
+    if (this.config.treatClosedCaptionsAsSubtitles) {
+      this.treatClosedCaptionsAsSubtitles = true;
+    }
+
```
dweinber: Not needed anymore once the private class variable is removed.

```suggestion
```

CHANGELOG.md
```diff
@@ -7,11 +7,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- New `SubtitleOverlayConfig.treatClosedCaptionsAsSubtitles` flag: by default, the player UI applies specific styling to Closed Captions. When enabled, this flag disables that behaviour and treats 608/708 captions like regular VTT subtitles from a styling perspective.
```
dweinber: Please update the changelog to match the changes I requested in this review

dweinber (changes requested):
Thank you for your contribution!

Sorry, I might have been unclear in the initial discussion around this feature.

Instead of disabling everything at once, I'd like to see this PR to disable _only_ the CEA-specific positioning. We already have a flag to disable only the CEA-specific formatting, so it would make more sense to have this pair of settings and customers can
- have all enabled (default, same as currently)
- enable positioning, disable formatting
- disable positioning, enable formatting
- disable positioning, disable formatting

dweinber:
Would also be great if you could add some unit tests for this new feature. You might want to look at the tests for `enableCea608CaptionFormatting` in `SubtitleOverlay.spec.ts`

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -27,6 +27,14 @@ export interface SubtitleOverlayConfig extends ContainerConfig {
    * Defaults to `true` (CEA-608 text formatting is applied, matching historical behavior).
    */
   enableCea608CaptionFormatting?: boolean;
+  /**
+   * Controls whether the CEA-608 grid-based row/column positioning is used or not.
+   * The formatting (monospaced font, uppercase transform, character letter-spacing) is preserved
+   * and still controlled independently via {@link enableCea608CaptionFormatting}.
+   *
```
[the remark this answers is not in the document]
dweinber: True, please address @LudoMichaud

src/ts/components/overlays/SubtitleOverlay.ts
```diff
@@ -257,7 +266,7 @@ export class SubtitleOverlay extends Container<SubtitleOverlayConfig> {
     // We need to keep track of the original row position in case of recalculation.
     const originalRowNumber = event.position?.row || 0;
 
-    if (isCea608SubtitleCue(event)) {
+    if (isCea608SubtitleCue(event) && this.isCea608PositioningEnabled()) {
```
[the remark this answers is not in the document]
dweinber: The specific failure that's described here (stale cues, broken CueExit matching, reset never running) does not occur, because `calculateId` is symmetric within each mode. Removing the normalization doesn't desynchronize enter vs. exit ids; both sides read the same (un)normalized object. The premise that the normalization created cross-event consistency is wrong.

Having said that, the suggested solution could still be implemented (but not strictly required IMO)

src/ts/components/overlays/SubtitleOverlay.ts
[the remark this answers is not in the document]
LudoMichaud: Addressed

dweinber (approved):
Looks good, thanks for the effort!

PR #898, opened 2026-06-19
fix(spatialnavigation): clamp TV seek bar cursor to its bounds

nebrey:
## Summary

Fixes the TV UI seek bar "hanging" at its start/end when scrubbing with a remote.

`SeekBarHandler.updateCursorPosition()` accumulated `cursorPosition.x` unconditionally on every navigation:

```ts
this.cursorPosition.x += this.getIncrement(direction, seekBarWrapper);
```

The `SeekBar` only *visually* clamps the dispatched position (`getHorizontalOffset` → `sanitizeOffset` → `[0, 1]`), so holding the remote in one direction past an edge keeps pushing the (invisible) cursor far beyond the seek bar with no upper bound. Reversing direction then does nothing visible until all that overshoot has been "unwound" — to the user it looks like the seek bar is stuck.

The fix clamps the cursor to the seek bar bounds so reversing direction at either edge takes effect immediately:

```ts
const rect = getBoundingRectFromElement(seekBarWrapper);
const minX = rect.x;
const maxX = rect.x + rect.width;
this.cursorPosition.x = Math.min(Math.max(this.cursorPosition.x + increment, minX), maxX);
```

Notes for review:
- Uses `getBoundingRectFromElement` rather than `getBoundingClientRect()` directly, matching `initializeCursorPosition`, so `x` is populated on older TV browsers that only return `left`/`top`.
- `getIncrement()` is still called exactly once (it has the scrub-speed acceleration side effect).
- Clamps to the seek bar **wrapper** rect — the same reference frame `getIncrement` and `initializeCursorPosition` already use.

Reported by a customer who pinpointed the exact method (via support).

## Test plan

- [x] Build the TV UI (`UIFactory.buildTvUI`), focus the seek bar, **hold Right** to the end, then press **Left** — playback position moves back immediately (previously needed many presses). Same at the start with hold-Left → Right.
- [x] `npm run lint` clean
- [x] `npx tsc --noEmit` clean
- [x] `npx jest` green (551/551). Added 3 cases to `SeekBarHandler.spec.ts` covering clamp-at-end, clamp-at-start, and immediate reverse-direction; verified each fails on the unpatched source.

## Known limitations / open questions for reviewers

- While the cursor is pinned at an edge, `getIncrement()` keeps accelerating `scrubSpeedPercentage`, so an *immediate* reverse can make the first step large. It auto-resets after `ScrubSpeedClearInterval` (100 ms) idle. Happy to reset scrub speed on clamp if preferred.
- Clamps to the wrapper, not the inner seek bar element. If there's horizontal padding between them, the cursor can still travel a few px past the inner bar before `sanitizeOffset` pins it — negligible on the current TV skin. Can switch to `getSeekBar(seekBarWrapper)` bounds if you'd rather be exact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- Seeking with the remote on the TV UI no longer gets stuck at the start or end of the seek bar: after scrubbing all the way to one edge, pressing the opposite direction now moves the playback position immediately instead of requiring multiple presses.
+
 ## [4.16.0] - 2026-06-18
 
 ### Added
--- a/spec/spatialnavigation/SeekBarHandler.spec.ts
+++ b/spec/spatialnavigation/SeekBarHandler.spec.ts
@@ -114,6 +114,54 @@ describe('SeekBarHandler', () => {
     });
   });
 
+  describe('cursor position clamping', () => {
+    let preventDefaultSpy: jest.Mock;
+    let targetComponentMock: AnyComponent;
+
+    beforeEach(() => {
+      preventDefaultSpy = jest.fn();
+      targetComponentMock = createComponentMock();
+      jest.spyOn(toHtmlElementModule, 'toHtmlElement').mockReturnValue(seekBarWrapperMock);
+    });
+
+    it('should not move the cursor position beyond the end of the seek bar', () => {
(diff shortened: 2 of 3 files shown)


stonko1994 (approved):
Works like a charm 💪 I leave it up to you whether to address the performance aspect from Codex.

nebrey:
Thanks @stonko1994! Addressed the performance note from the bot in f9583b5f: `updateCursorPosition` now reads the seek bar layout once and passes the width into `getIncrement`, so there's a single `getBoundingClientRect()` per navigation instead of two. `getIncrement`'s scrub-speed side effect is unchanged. Tests/lint/tsc still green (551/551).

stonko1994 (approved):

PR #905, opened 2026-06-24
Add UI component config API

stonko1994:
## Description
Adds the public `UIConfig.components` API shape, including component and variant-scoped config types. Documents precedence in the TypeDoc-visible API and adds guard coverage so public component exports and UI variant identifiers stay represented.

- [x] `CHANGELOG` entry

### DIFF
--- a/spec/utils/UIComponentConfigOverrides.spec.ts
+++ b/spec/utils/UIComponentConfigOverrides.spec.ts
@@ -5,9 +5,9 @@ import { Component } from '../../src/ts/components/Component';
 import * as PlayerUI from '../../src/ts/main';
 import { UIVariantIdentifier } from '../../src/ts/UIManager';
 
-const UI_COMPONENTS_CONFIG_PATH = path.resolve(__dirname, '../../src/ts/UIComponentsConfig.ts');
+const UI_COMPONENT_CONFIG_OVERRIDES_PATH = path.resolve(__dirname, '../../src/ts/UIComponentConfigOverrides.ts');
 
-describe('UIComponentsConfig', () => {
+describe('UIComponentConfigOverrides', () => {
   it('contains every public component export', () => {
     const componentConfigNames = readInterfaceKeys('UIComponentConfigMap');
     const publicComponentNames = getPublicComponentExports().map(({ exportName }) => exportName);
@@ -19,24 +19,25 @@ describe('UIComponentsConfig', () => {
     if (missingConfigNames.length > 0) {
       throw new Error(
         `Missing UIComponentConfigMap entries for public component exports: ${missingConfigNames.join(', ')}. ` +
-          'Add them to src/ts/UIComponentsConfig.ts.',
+          'Add them to src/ts/UIComponentConfigOverrides.ts.',
       );
     }
   });
 
   it('contains every UI variant identifier', () => {
-    const uiComponentsConfigNames = readInterfaceKeys('UIComponentsConfig');
+    const uiComponentConfigOverrideNames = readInterfaceKeys('UIComponentConfigOverrides');
     const variantIdentifiers = Object.keys(UIVariantIdentifier).map(
       variantName => UIVariantIdentifier[variantName as keyof typeof UIVariantIdentifier],
     );
     const missingVariantIdentifiers = variantIdentifiers.filter(
-      variantIdentifier => !uiComponentsConfigNames.includes(variantIdentifier),
+      variantIdentifier => !uiComponentConfigOverrideNames.includes(variantIdentifier),
     );
 
     if (missingVariantIdentifiers.length > 0) {
       throw new Error(
-        `Missing UIComponentsConfig entries for UI variant identifiers: ${missingVariantIdentifiers.join(', ')}. ` +
-          'Add them to src/ts/UIComponentsConfig.ts.',
+        `Missing UIComponentConfigOverrides entries for UI variant identifiers: ${missingVariantIdentifiers.join(
+          ', ',
+        )}. ` + 'Add them to src/ts/UIComponentConfigOverrides.ts.',
       );
     }
   });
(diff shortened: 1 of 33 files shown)


spec/utils/UIComponentConfigOverrides.spec.ts
stonko1994, line 1: This is a verification test to ensure every component is added to the `UIComponentConfigMap`

src/ts/UIComponentConfigOverrides.ts
```diff
@@ -0,0 +1,525 @@
...
+
+/**
+ * Component config overrides keyed by public component class name.
+ *
+ * Each property matches a public component class name. The value type is the component-specific constructor config.
+ *
+ * @category Configs
+ */
+// Keep this map in sync with public component exports from main.ts.
+export interface UIComponentConfigMap {
```
stonko1994, line 160: This class mainly covers the documentation purpose. The actual implementation in #906 doesn't rely on it but to have a proper JSDoc rendering we decided to add it.

src/ts/UIManager.ts
```diff
@@ -168,6 +168,21 @@ export interface UIConditionResolver {
   (context: UIConditionContext): boolean;
 }
 
+/**
+ * Identifier for the different {@link UIVariant}s.
+ */
+export enum UIVariantIdentifier {
```
stonko1994, line 175: This new enum is required to identify the variant and the variant specific config overrides in the `UIComponentsConfig`.

typedoc.json
```diff
@@ -8,5 +8,15 @@
     "includeGroups": true
   },
   "excludeProtected": true,
-  "categoryOrder": ["UI Framework", "Configs", "Components", "Buttons", "Labels", "Containers", "Localization", "Utils"]
+  "categoryOrder": [
+    "UI Framework",
+    "Configs",
+    "UI variants",
```
stonko1994, line 14: This was added to ensure the UI variant identifiers are on top in the documentation of the `UIComponentsConfig` documentation.

src/ts/UIComponentConfigOverrides.ts
```diff
@@ -0,0 +1,525 @@
...
+ * config.
+ *
+ * Config precedence, from lowest to highest:
+ * 1. Component superclass config
+ * 2. Component default config
+ * 3. Constructor config from the {@link UIFactory}
+ * 4. Top-level base component config, e.g. `ToggleButton`
+ * 5. Top-level specific component config, e.g. `FullscreenToggleButton`
+ * 6. Variant-specific base component config, e.g. `main.ToggleButton`
+ * 7. Variant-specific component config, e.g. `main.FullscreenToggleButton`
```
dweinber, line 64: I'm still undecided if `main.ToggleButton` should have higher precedence than `FullscreenToggleButton` 🤔

src/ts/UIConfig.ts
```diff
@@ -235,6 +237,30 @@ export interface UIConfig {
...
+   *   ToggleButton: { buttonStyle: ButtonStyle.Text },
+   *
+   *   // Applies only to the main UI variant.
+   *   main: {
+   *     FullscreenToggleButton: { text: 'Main fullscreen' },
+   *   },
+   * }
+   * ```
+   */
+  components?: UIComponentsConfig;
```
dweinber: As a developer, by just reading the `components` I would assume I could specify which components to show (or add/remove components or so).

But `UIConfig.componentConfigs` sounds not too nice, and not sure what other ideas could be 🤔

stonko1994: what do you think about `componentsConfigOverrides`? Would also make it clearer from the name that this here has precedence. 🤔

dweinber: Sounds like the best option to me 🙂

stonko1994: [d4e72f9](https://github.com/bitmovin/bitmovin-player-ui/pull/905/commits/d4e72f97bfa8ee6688b68b6f0882c432fe13979a)

src/ts/UIComponentConfigOverrides.ts
stonko1994: So you're thinking of switching 5 and 6 around? 🤔

dweinber: As discussed offline: let's keep it as is.

src/ts/UIComponentConfigOverrides.ts
```diff
@@ -0,0 +1,525 @@
...
+ *   },
+ *   SeekBar: {
+ *     // Component specific config
+ *   },
+ * };
+ * ```
+ *
+ * @category Configs
+ */
+export interface UIComponentConfigOverrides extends UIComponentConfigMap {
```
dweinber, line 95: I think we also need to add these new exports to `main.ts` (import there, and export it again).

dweinber (changes requested):

src/ts/UIComponentConfigOverrides.ts
```diff
@@ -0,0 +1,525 @@
...
+   */
+  ListSelector?: Partial<ListSelectorConfig>;
+  /**
+   * @category Components
+   */
+  MetadataLabel?: Partial<MetadataLabelConfig>;
+  /**
+   * @category Components
+   */
+  PersistentPreferencesToggleButton?: Partial<ToggleButtonConfig>;
```
dweinber: What's a `PersistentPreferencesToggleButton`? Doesn't seem to exist 🤔

dweinber: And should we add a "reverse" check in the tests to ensure every entry in this map has a component? 😅

src/ts/UIComponentConfigOverrides.ts
stonko1994: I was already there as the `UIConfig.ts` has the export already but I moved it to `main.ts` to have all of them at a central place. [819aed3](https://github.com/bitmovin/bitmovin-player-ui/pull/905/commits/819aed360907a594595a3ed9d46b8f09e1a45272)

src/ts/UIComponentConfigOverrides.ts
stonko1994: Leftover from another branch 😬 Added the reverse check: [8a6df08](https://github.com/bitmovin/bitmovin-player-ui/pull/905/commits/8a6df08dd8ff372fdcfc27d55c564889a95c0bc1)

dweinber (approved):
Really cool feature 🙌

PR #907, opened 2026-06-25
Introduce toggle settings panel item

stonko1994:
## Description
Introduce `ToggleSettingsPanelItem`, a settings row that toggles its nested `ToggleButton` when the row is selected. Use it for the existing eco-mode settings toggle and keep spatial navigation focused on interactive settings rows.

## Manual Testing
Enable `ecoMode: true,` in the `UIConfig` for easier manual testing.

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/settings/ToggleSettingsPanelItem.ts
+++ b/src/ts/components/settings/ToggleSettingsPanelItem.ts
@@ -1,0 +1,63 @@
+import { PlayerAPI } from 'bitmovin-player';
+
+import { UIInstanceManager } from '../../UIManager';
+import { ToggleButton, ToggleButtonConfig } from '../buttons/ToggleButton';
+import { InteractiveSettingsPanelItem } from './InteractiveSettingsPanelItem';
+import { SettingsPanelItemConfig } from './SettingsPanelItem';
+
+/**
+ * Configuration interface for a {@link ToggleSettingsPanelItem}.
+ *
+ * @category Configs
+ */
+export interface ToggleSettingsPanelItemConfig extends SettingsPanelItemConfig {
+  /**
+   * The toggle button that will be toggled when this item is clicked.
+   */
+  settingComponent: ToggleButton<ToggleButtonConfig>;
+}
+
+/**
+ * A settings panel item that toggles a {@link ToggleButton} when the row is clicked.
+ *
+ * @category Components
+ */
+export class ToggleSettingsPanelItem extends InteractiveSettingsPanelItem<ToggleSettingsPanelItemConfig> {
+  protected settingComponent: ToggleButton<ToggleButtonConfig>;
+
+  constructor(config: ToggleSettingsPanelItemConfig) {
+    super(config);
+
+    this.config = this.mergeConfig(
+      config,
+      {
+        cssClasses: ['ui-toggle-settings-panel-item'],
+        role: 'menuitemcheckbox',
+        tabIndex: 0,
+      } as ToggleSettingsPanelItemConfig,
+      this.config,
+    );
+  }
+
+  configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
+    super.configure(player, uimanager);
+
+    // Keyboard selection belongs to the row, otherwise Enter/Space can activate both the button and the row.
+    this.settingComponent.getDomElement().attr('tabindex', '-1');
+    // The button handles its own click; keep that click from bubbling into the row and toggling twice.
+    this.settingComponent.getDomElement().on('click', event => event.stopPropagation());
+
+    this.onClick.subscribe(() => {
+      if (!this.settingComponent.isDisabled()) {
+        this.settingComponent.toggle();
+      }
+    });
+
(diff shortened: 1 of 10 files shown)


src/ts/components/settings/ToggleSettingsPanelItem.ts
```diff
@@ -0,0 +1,63 @@
...
+    );
+  }
+
+  configure(player: PlayerAPI, uimanager: UIInstanceManager): void {
+    super.configure(player, uimanager);
+
+    // Keyboard selection belongs to the row, otherwise Enter/Space can activate both the button and the row.
+    this.settingComponent.getDomElement().attr('tabindex', '-1');
+    // The button handles its own click; keep that click from bubbling into the row and toggling twice.
+    this.settingComponent.getDomElement().on('click', event => event.stopPropagation());
```
hiren3897: Heads up! 
this `stopPropagation` means clicking the button itself does nothing for a plain `ToggleButton`. The click stops here and never reaches the row's `toggle()` on line 50 right?, and `ToggleButton's` own click doesn't change state by itself. 

It only works for eco-mode because `EcoModeToggleButton` happens to call `toggle()` on its own click. Since this component is exported, anyone passing a vanilla `ToggleButton` gets a dead button. Could we let the row handle the button's click too (and dedup), or document that the row owns toggling?

src/ts/components/settings/ToggleSettingsPanelItem.ts
```diff
@@ -0,0 +1,63 @@
...
+  protected settingComponent: ToggleButton<ToggleButtonConfig>;
+
+  constructor(config: ToggleSettingsPanelItemConfig) {
+    super(config);
+
+    this.config = this.mergeConfig(
+      config,
+      {
+        cssClasses: ['ui-toggle-settings-panel-item'],
+        role: 'menuitemcheckbox',
```
hiren3897, line 35: Small a11y thing: the row is `menuitemcheckbox` + `aria-checked`, but the inner toggle button still reports `aria-pressed` (ToggleButton falls back to it when there's no on/off aria label). So a `screen reader` hears a checkbox wrapping a pressable toggle -> two state semantics for one control. Since the button is now non-focusable, maybe drop its `role/aria-pressed` so only the row speaks 🤔

src/ts/components/settings/ToggleSettingsPanelItem.ts
```diff
@@ -0,0 +1,63 @@
...
+    // Keyboard selection belongs to the row, otherwise Enter/Space can activate both the button and the row.
+    this.settingComponent.getDomElement().attr('tabindex', '-1');
+    // The button handles its own click; keep that click from bubbling into the row and toggling twice.
+    this.settingComponent.getDomElement().on('click', event => event.stopPropagation());
+
+    this.onClick.subscribe(() => {
+      if (!this.settingComponent.isDisabled()) {
+        this.settingComponent.toggle();
+      }
+    });
```
hiren3897, line 66: We skip toggling when the button is disabled, nice! 💪  but the row stays focusable and still announces as a `menuitemcheckbox` with no `aria-disabled`!!

So a disabled eco toggle still reads as actionable and sits in the focus order while clicks do nothing. Worth mirroring the disabled state onto the row?

hiren3897 (commented):
Took a pass through this! heads up that I'm still new to the player-ui codebase 😃, so I mostly read through the diff and ran it locally to play with the eco-mode toggle. Behavior looked good: row click toggles, keyboard select works, button's out of the tab order as intended.

Left a few inline comments 😄  all minor, nothing blocking. 
- Main one worth a look is the button-click path (clicking the button itself vs. the row) for when this gets reused with a plain `ToggleButton`
- The rest are small `a11y`

Could easily be missing context on conventions here, so feel free to wave any of them off.

Nice generalizing the spatial-nav check to InteractiveSettingsPanelItem 👍

src/ts/components/settings/ToggleSettingsPanelItem.ts
stonko1994: [8ee4f62](https://github.com/bitmovin/bitmovin-player-ui/pull/907/commits/8ee4f629af4fe33fdbfa126dfcb144043a532d64)

src/ts/components/settings/ToggleSettingsPanelItem.ts
stonko1994: [9f1878f](https://github.com/bitmovin/bitmovin-player-ui/pull/907/commits/9f1878f39b06bed405bd10a32a0e0f373c95e0ef)

src/ts/components/settings/ToggleSettingsPanelItem.ts
stonko1994: [8385e08](https://github.com/bitmovin/bitmovin-player-ui/pull/907/commits/8385e089ae9183e663f2e6104dcfee19ac0d72fa) I kept it focusable but updated the accessibility accordingly.

hiren3897 (approved):
LGTM 🙌

PR #910, opened 2026-07-08
Add component layout overrides

stonko1994:
## Description
Adds a `UIConfig.componentLayoutOverrides` API for including or excluding default layout components without replacing the default layout.

Default layouts now build a superset of our default component tree. When `UIManager` resolves a UI variant, it applies layout overrides only if the variant has a `UIVariantIdentifier`. The `ComponentLayoutOverrideProcessor` then runs before `configureControls()`, walks the resolved tree, removes components configured as `Exclude` from their owning container, and immediately releases the removed subtree. Settings rows that own excluded setting components, such as `PlaybackSpeedSelectBox`, are removed as the owning row instead of only removing the nested component. Only the remaining component tree is initialized and configured.

## Limitations
This API only applies to UI variants with a `UIVariantIdentifier`. Variants without an identifier keep their resolved component tree unchanged, so custom or forked layouts are not stripped unless they opt into an identifier that matches the default-layout override handling.

`componentLayoutOverrides` controls membership only. Custom placement, ordering, duplicate instances, and per-instance configuration remain custom-layout or `componentConfigOverrides` use-cases.

- [x] `CHANGELOG` entry


src/ts/UIFactory.ts
```diff
@@ -533,7 +538,7 @@ export namespace UIFactory {
       });
     }
 
-    export function castReceiver(config: UIConfig = {}): UIContainer {
+    export function castReceiver(): UIContainer {
```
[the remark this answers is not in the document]
stonko1994: I'm fine with this breaking change as it's a minor change and a not frequently used layout.

src/ts/UIComponentLayoutOverrides.ts
stonko1994, line 1: This is just boiler plate code to have the same documentation coverage as for the `UIComponentConfigOverrides`

src/ts/components/buttons/QuickSeekButton.ts
```diff
@@ -135,10 +135,14 @@ export class QuickSeekButton extends Button<QuickSeekButtonConfig> {
   };
 
   release(): void {
-    this.player.off(this.player.exports.PlayerEvent.Seek, this.onSeek);
-    this.player.off(this.player.exports.PlayerEvent.Seeked, this.onSeekedOrTimeShifted);
-    this.player.off(this.player.exports.PlayerEvent.TimeShift, this.onTimeShift);
-    this.player.off(this.player.exports.PlayerEvent.TimeShifted, this.onSeekedOrTimeShifted);
+    super.release();
+
+    const player = this.player;
```
dweinber, line 140: Shouldn't this be done before `super.release()`?

src/ts/UIFactory.ts
```diff
@@ -255,6 +256,11 @@ export namespace UIFactory {
    *
    * Using these methods enables customization of certain aspects of the default UI without recreating all UI variants.
    * They can be used to recreate the default UI while changing the conditions based on which variant switching happens.
+   *
+   * TODO: `UIConfig.componentLayoutOverrides` are currently processed by UIManager for variants with identifiers.
+   * This skips older copied or forked custom layouts without identifiers, but cannot distinguish built-in default layouts
+   * from custom layouts that reuse default `UIVariantIdentifier` values. Decide whether default layouts need to
+   * explicitly opt in before stabilizing the API.
```
dweinber: Should we discuss this and finalize before merging?

This also means, that if a custom UI currently uses the `defaultLayouts` to customize parts of it, but the don't explicitly add an identifier, they would automatically get the QuickSeekButtons and others enabled by default.

One option I could think of is to add an identifier to the defaultLayouts.

src/ts/UIFactory.ts
```diff
@@ -307,7 +312,9 @@ export namespace UIFactory {
           }),
           new Container({
             components: [
+              new QuickSeekButton({ seekSeconds: -10 }),
               new PlaybackToggleButton(),
+              new QuickSeekButton({ seekSeconds: 10 }),
```
dweinber, line 312: Finally 🎉

src/ts/UIManager.ts
```diff
@@ -769,6 +790,12 @@ export class UIManager {
 
   private addUi(ui: InternalUIInstanceManager): void {
     const uiContainer = ui.resolveUI();
+
+    // Layout overrides target built-in variant identifiers. Variants without identifiers keep their tree unchanged.
+    if (ui.variantIdentifier != null) {
```
dweinber: Please try to avoid loose inequality comparisons. Use strict (`!==`) or just check if it's a truthy value (`if (ui.variantIdentifier) {`)

src/ts/utils/ComponentLayoutOverrideProcessor.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export class ComponentLayoutOverrideProcessor {
+  private config: UIConfig;
+
+  constructor(config: UIConfig) {
+    this.config = config;
+  }
+
+  process(uiContainer: Container<ContainerConfig>, variantIdentifier: UIVariantIdentifier): void {
+    if (this.config.componentLayoutOverrides == null) {
```
dweinber: strict equality please 🙏

src/ts/UIFactory.ts
dweinber: On a second thought, I think it's currently not possible for a customer to enable/disable only one of the two QuickSeekButtons, right?
Thinking of adding the "-10" button but no "+10". 

If that's the case, should we add a `QuickSeekForwardButton` and a `QuickSeekBackButton` or so?

src/ts/components/buttons/QuickSeekButton.ts
dweinber: Do we need to do this now for all components? I guess so. In that case, then please make sure this approach is applied to all components.

Looks like that if we add `VolumeControlButton` to the layout, but then `Exclude` it, there will be an unhandled exception in the `VolumeControlButton`'s `release` method as `this.volumeSliderHideTimeout` is `undefined` but `this.volumeSliderHideTimeout.clear()` is called. This leads to the whole UI not being available.

src/ts/UIComponentLayoutOverrides.ts
```diff
@@ -0,0 +1,489 @@
...
+   */
+  UIContainer?: UIComponentLayoutOverride;
+  /**
+   * @category Components
+   */
+  VideoQualitySelectBox?: UIComponentLayoutOverride;
+  /**
+   * @category Components
+   */
+  VolumeControlButton?: UIComponentLayoutOverride;
```
dweinber: FWIW, it seems we're not using that Button at all in the default layout.

src/ts/UIComponentLayoutOverrides.ts
```diff
@@ -0,0 +1,489 @@
...
+   */
+  ReplayButton?: UIComponentLayoutOverride;
+  /**
+   * @category Components
+   */
+  SeekBar?: UIComponentLayoutOverride;
+  /**
+   * @category Components
+   */
+  SeekBarLabel?: UIComponentLayoutOverride;
```
dweinber: Disabling it doesn't seem to have any effect.

(Note: I did not test all the components, just a few)

src/ts/components/buttons/QuickSeekButton.ts
dweinber: Maybe we should track a `configured` flag in `Component` and make pre-configure `release` run only constructor-level cleanup, or audit/guard all `release()` overrides the way `QuickSeekButton` was fixed. 

Additionally, we should think of automatically testing the exclusion of each key in the map (or at least the default-layout members) against a real default layout to catch issues like that.

dweinber (changes requested):
Really great feature, this will be very helpful going forward! Thanks 🙂

src/ts/UIConfig.ts
```diff
@@ -294,6 +301,27 @@ export interface UIConfig {
    * ```
    */
   componentConfigOverrides?: UIComponentConfigOverrides;
+  /**
+   * Allows including or excluding supported components from default UIFactory layouts without rebuilding the layout.
+   *
+   * Component layout overrides are only applied to UI variants with a {@link UIVariantIdentifier}. Variants without an
+   * identifier keep their resolved component tree unchanged.
+   *
+   * See {@link UIComponentLayoutOverrides} for more details.
```
123mpozzi:
```suggestion
   * @see {@link UIComponentLayoutOverrides} for more details.
```

nit: typescript/javascript have 'see' built-in. Generally document generators will handle this better with the tag

src/ts/UIComponentLayoutOverrides.ts
```diff
@@ -0,0 +1,489 @@
...
+ * 1. Built-in default layout membership
+ * 2. Legacy aliases (`includeWatermark`, `ecoMode`, and `playbackSpeedSelectionEnabled`)
+ * 3. Top-level `componentLayoutOverrides`
+ * 4. Variant-scoped `componentLayoutOverrides`
+ *
+ * Top-level entries apply to all identified UI variants. Variant-scoped entries override top-level entries for the
+ * matching variant.
+ *
+ * Base component keys also apply to subclasses, and more specific component keys override base component keys. For
+ * example, `ToggleButton` applies to `FullscreenToggleButton`, unless `FullscreenToggleButton` has its own override.
```
123mpozzi, line 34: Kudos on adding many examples! It makes it so much simpler to follow the docs

src/ts/UIComponentLayoutOverrides.ts
```diff
@@ -0,0 +1,489 @@
...
+/**
+ * Component layout overrides keyed by public component class name.
+ *
+ * Each property matches a public component class name. The value controls whether default layouts keep or remove
+ * matching component instances. Base component keys apply to subclasses, and more specific component keys override base
+ * component keys.
+ *
+ * @category Configs
+ */
+// Keep this map in sync with UIComponentConfigMap from UIComponentConfigOverrides.ts.
```
123mpozzi: Does the public doc still apply when split up with a code doc like this?

src/ts/UIFactory.ts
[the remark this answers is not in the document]
123mpozzi: Should we at least document in the CHANGELOG?

src/ts/UIManager.ts
123mpozzi: Yes, ☝️  - this should actually be enabled as a linter rule IMO

src/ts/UIComponentLayoutOverrides.ts
```diff
@@ -0,0 +1,489 @@
+import type { UIVariantIdentifier } from './UIManager';
+
+export enum UIComponentLayoutOverride {
+  Include = 'include',
```
123mpozzi, line 4: Include is slightly misleading: it does not really “add” a component. It only prevents a component from being removed if the default layout already built it.  I propose to reword to be more explicit:

_Applies to whole feature_

```suggestion
  keep = 'keep',
```

src/ts/UIComponentLayoutOverrides.ts
123mpozzi: Btw I think we can address the comment completely with dynamic typing:

```ts
     export type UIComponentLayoutOverrideMap = {
       [ComponentName in keyof UIComponentConfigMap]?: UIComponentLayoutOverride;
     };
```

Only downsides:
- TypeDoc output may be worse than the explicit interface
- It makes the layout API promise that every configurable component is also meaningfully layout-
  toggleable (not sure what's our current position here)

src/ts/UIConfig.ts
```diff
@@ -223,13 +226,17 @@ export interface UIConfig {
...
+   * @deprecated Use {@link UIConfig.componentLayoutOverrides} with `EcoModeContainer` instead.
    */
   ecoMode?: boolean;
   /**
    * Specifies if the Watermark element should be included in the UI.
    * Per default, the Watermark shows the Bitmovin Logo.
    *
    * Default: false
+   *
+   * @deprecated Use {@link UIConfig.componentLayoutOverrides} with `Watermark` instead.
```
123mpozzi: nit: I'd provide more explicit guidance for deprecations:

_Applies to all deprecations_

```suggestion
   * @deprecated Use {@link UIConfig.componentLayoutOverrides} to include or exclude the default `Watermark`.
```

src/ts/UIFactory.ts
123mpozzi: That's a good point!

I think our best options are either:
1. Add `QuickSeekBackwardButton` and `QuickSeekForwardButton`:
```ts
     componentLayoutOverrides: {
       QuickSeekBackwardButton: UIComponentLayoutOverride.Exclude,
       QuickSeekForwardButton: UIComponentLayoutOverride.Include,
     }
```
2. Introduce semantic slot keys: 
```ts
     componentLayoutOverrides: {
       quickSeekBackward: Exclude,
       quickSeekForward: Include,
     }
```

src/ts/UIComponentLayoutOverrides.ts
```diff
@@ -0,0 +1,489 @@
...
+  Exclude = 'exclude',
+}
+
+/**
+ * Component layout overrides keyed by public component class name, optionally scoped by UI variant.
+ *
+ * Overrides are applied after a UI variant has been resolved and before `configureControls()` runs. Excluded components
```
123mpozzi, line 11: I see this as the biggest concern: layout overrides are implemented as constructing the full default tree, then removing excluded components before `configureControls()`.

That means excluded components still run constructors and then receive `release()` before `configure()`. Several `release()` implementations assume `configure()` already initialized runtime fields; for example excluding `VolumeControlButton` can call `release()` while `volumeSliderHideTimeout` is still undefined, which breaks UI creation.

I think the safer API behavior is to decide default-layout membership while constructing the `UIFactory` layout, so excluded components are never instantiated.

src/ts/UIFactory.ts
123mpozzi: Agree. I think the best fix would be to apply the override while building the default layout, so excluded default components are never created. Otherwise direct `UIFactory.defaultLayouts.*` consumers without a variant identifier can bypass the pruning step and get newly unconditional components unexpectedly

123mpozzi (commented):
Adding my input: the API direction looks good to me overall. I think there are a few points we should clarify before merging

src/ts/UIManager.ts
dweinber: Aggreed @123mpozzi. I looked into that when I added several of these comments 😅 
While we should definitely do that, enabling that rule showed up a lot of places where this rule is violated 🙈 
I didn't have the bandwidth to fix all those yet, so I left it there (and definitely out of scope for this PR, should be its own).

123mpozzi: Yeah, definitely out of scope . Thanks for looking!

src/ts/UIFactory.ts
stonko1994: Adding `QuickSeekBackwardButton` and `QuickSeekForwardButton` would be an option but I'm not sure if we need this.

The whole feature is considered a convenience, which means the general 'fork and modify' approach is still very much possible and encouraged.

---

> Introduce semantic slot keys:

I would not go that far, as this would introduce a huge ton of maintenance right now.

src/ts/UIComponentLayoutOverrides.ts
stonko1994: No the docs won't look good then anymore. We explicitly agreed, with @dweinber, to manually create those lists to cover all of them in the docs. Yes the dynamic typing works but then we loose all the possible key value combinations.

src/ts/UIFactory.ts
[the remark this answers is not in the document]
stonko1994: I don't think it's required. Can add it though if you think it's worth it

src/ts/UIComponentLayoutOverrides.ts
stonko1994: I don't think include is misleading 🤔 A component is either `included` or `excluded`. What would be the counterpart to keep? `keep` and `drop`?

src/ts/UIComponentLayoutOverrides.ts
stonko1994: > I think the safer API behavior is to decide default-layout membership while constructing the UIFactory layout, so excluded components are never instantiated.

Yes that would be safer but it would require that every component initialization in the UIFactory needs to dynamically check whether it's included/excluded. I'm not saying this is bad design but it would completely change our `UIFactory` implementation which I consider as 'API breaking' or better said 'fork breaking' which I don't want to do now. I treat the `UIFactory` kinda as public API which should be easy to use and modify in a fork.

src/ts/UIFactory.ts
stonko1994: > Agree. I think the best fix would be to apply the override while building the default layout, so excluded default components are never created. Otherwise direct UIFactory.defaultLayouts.* consumers without a variant identifier can bypass the pruning step and get newly unconditional components unexpectedly

IMO this is not feasible right now. See this comment for more details: https://github.com/bitmovin/bitmovin-player-ui/pull/910#discussion_r3704140480

--- 


> This also means, that if a custom UI currently uses the defaultLayouts to customize parts of it, but the don't explicitly add an identifier, they would automatically get the QuickSeekButtons and others enabled by default.
>
> One option I could think of is to add an identifier to the defaultLayouts.

Yes this option is the fix and I'm thinking this auto fixes itself after updating to thie version. The `defaultLayouts` will get an id after updating and then will exclude the new components again. Also when a fork is updating to this version they will get the id and, worst case, also see the change that the new buttons are added.

---

_I forgot to remove the TODO but for me this was the desired approach._

src/ts/UIComponentLayoutOverrides.ts
stonko1994: Good catch. I wonder how far we go with this? We could just add the `VolumeControlButton` to the layouts and exclude it per default. However, this problem doesn't end here. As you outlined in your comment below (https://github.com/bitmovin/bitmovin-player-ui/pull/910#discussion_r3585176223) there are use-cases where the 'traversing' just can't see certain components.

src/ts/UIComponentLayoutOverrides.ts
stonko1994: This is true and relates to https://github.com/bitmovin/bitmovin-player-ui/pull/910#discussion_r3704356718

This mainly affects 'nested' components, such as the `SeekBarLabel` but also child components (e.g. a `SettingsPanelSelectOption` or a `RecommendationItem`).

I think it could be quite tedious and compex to include every possible component 🤔 Do you think this is a valid limitation for the initial version?

src/ts/UIManager.ts
stonko1994: [6362602](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/63626025895516b7823603c0ab75305cd6facdf0)

src/ts/utils/ComponentLayoutOverrideProcessor.ts
stonko1994: [6362602](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/63626025895516b7823603c0ab75305cd6facdf0)

src/ts/UIConfig.ts
stonko1994: [9b28e23](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/9b28e23db727c6160e069eb756f2e86af71b9ae9)

src/ts/UIConfig.ts
stonko1994: [1cd9481](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/1cd94813692688fde3d64276778c99381fc6a048)

src/ts/components/buttons/QuickSeekButton.ts
stonko1994: Decided against the configured flag and made release implementations safe. [d24b1c1](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/d24b1c148abb3f9cfe8c56c74ff587813b190e8f)

src/ts/UIComponentLayoutOverrides.ts
stonko1994: > That means excluded components still run constructors and then receive release() before configure(). Several release() implementations assume configure() already initialized runtime fields

[d24b1c1](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/d24b1c148abb3f9cfe8c56c74ff587813b190e8f)

src/ts/UIFactory.ts
stonko1994: [54df6e1](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/54df6e1626b26d70d5b2c70cb58126340377c719)

src/ts/UIComponentLayoutOverrides.ts
123mpozzi: My point is only that `Include` can read like it adds the component. In practice it only keeps a component if the selected default layout already created it. E.g.:

```ts
  componentLayoutOverrides: {
    Watermark: UIComponentLayoutOverride.Include,
  }
```

This only works for layouts that already create a Watermark and would otherwise remove it by default


It's just a wording concern tho, non blocking

src/ts/UIComponentLayoutOverrides.ts
123mpozzi: Thanks, I think the release-safety part is addressed by the new guards/tests.

src/ts/UIFactory.ts
123mpozzi: Ah I see, so users would need to build/fork a custom `UIFactory` layout and manually remove only `-10` or `+10` for this use case? I think that's fair for first version, since I don't see any simple way to handle that customization without many changes.

We could document it as initial limitation

src/ts/components/buttons/QuickSeekButton.ts
123mpozzi: Any reason to put `player.off(..)` after `super.release()` ? I see [CoPilot](https://github.com/bitmovin/bitmovin-player-ui/pull/910/changes#r3549463018) tagged it as well

src/ts/utils/ComponentLayoutOverrideProcessor.ts
```diff
@@ -0,0 +1,121 @@
...
+    if (settingComponent != null) {
+      this.releaseComponentTree(settingComponent, releasedComponents);
+    }
+
+    if (component instanceof Container) {
+      const childComponents = component.getComponents().slice();
+      for (const childComponent of childComponents) {
+        this.releaseComponentTree(childComponent, releasedComponents);
+      }
+    }
```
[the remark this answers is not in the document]
123mpozzi: _Unresolving for visibility_

nit: this could still be worth addressing and could be low hanging: capture capture `settingComponent` / `childComponents` before calling `release()`, then release the tree bottom-up.

It works today because `Container.release()` does not clear children but any future `release()` implementation that clears owned children would make this skip child cleanup

src/ts/UIFactory.ts
123mpozzi: I think this is acceptable for the first version as long as the limitation is explicit. Fixing the remaining custom `UIManager` case properly would either move override checks into every `defaultLayouts.*` builder or change the shape of what default layouts return, both of which add quite a bit of complexity / API impact.

I’d document the contract as: built-in `UIFactory.build*UI()` paths are covered, and custom `UIManager` variants need an `identifier` if they should participate in `componentLayoutOverrides`.

src/ts/UIConfig.ts
```diff
@@ -132,6 +133,9 @@ export interface UIConfig {
   /**
    * Specifies if the `PlaybackSpeedSelectBox` should be displayed within the `SettingsPanel`
    * Default: true
+   *
+   * @deprecated Use {@link UIConfig.componentLayoutOverrides} to include or exclude the default
+   * `PlaybackSpeedSelectBox`.
```
123mpozzi: nit: could be slightly clearer: deprecation says `PlaybackSpeedSelectBox`, but behaviorally this removes the default settings row that owns it.

```suggestion
   * @deprecated Use {@link UIConfig.componentLayoutOverrides} to include or exclude the default playback-speed settings
   * row via `PlaybackSpeedSelectBox`.
```

src/ts/UIConfig.ts
```diff
@@ -232,13 +236,17 @@ export interface UIConfig {
   enableTimestampDeepLink?: boolean;
   /**
    * Specifies if the `EcoModeToggleButton` should be displayed within the `SettingsPanel`
+   *
+   * @deprecated Use {@link UIConfig.componentLayoutOverrides} to include or exclude the default `EcoModeContainer`.
```
123mpozzi: nit: old config (line above changes) mentions `EcoModeToggleButton`, replacement says `EcoModeContainer`. That is probably correct implementation-wise, but confusing without explaining that `EcoModeContainer` is the default layout target.

```suggestion
   * @deprecated Use {@link UIConfig.componentLayoutOverrides} to include or exclude the default Eco Mode settings group
   * via `EcoModeContainer`.
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - `UIConfig.enableResumeFromLastPosition` option to enable storing playback progress and resume known sources from the last saved position.
+- `UIConfig.componentLayoutOverrides` and `UIComponentLayoutOverride` to include or exclude default UI components without rebuilding complete UI layouts through the `UIFactory`.
```
123mpozzi: nit: I think the current phrasing undersells the feature a bit, I'd emphasize more what it solves 

```suggestion
- `UIConfig.componentLayoutOverrides` and `UIComponentLayoutOverride` to keep or remove supported components from default UI layouts without rebuilding complete `UIFactory` layouts.
```

src/ts/UIComponentLayoutOverrides.ts
123mpozzi: I agree this does not end with `VolumeControlButton`. The bigger issue is that the public map currently documents keys that the traversal may never see, or can only affect indirectly/dynamically. That gives users typed options that look supported but silently do nothing.

For this first version, I’d prefer separating layout targets from config targets: expose only the component keys we can guarantee have a meaningful effect in default layouts, plus intentional special cases like `PlaybackSpeedSelectBox` removing its owning settings row. A guard test for the exposed keys would also keep the list from drifting.

123mpozzi (commented):
I think we're mostly there 🙂  - we just need to find an agreement with the remaining open threads, but I don't see any major blockers

src/ts/components/buttons/QuickSeekButton.ts
stonko1994: IMO, retaining the player in a component should be an anti pattern in our UI as stuff like this is usually handled in `configured`. Therefore there should be no player in `super`. In thise case it's totally fine to have it like this. I can change it if you want it but I don't see an issue with this now.

src/ts/UIFactory.ts
stonko1994: > I’d document the contract as: built-in UIFactory.build*UI() paths are covered, and custom UIManager variants need an identifier if they should participate in componentLayoutOverrides.

I did document that here: https://github.com/bitmovin/bitmovin-player-ui/blob/54df6e1626b26d70d5b2c70cb58126340377c719/src/ts/UIManager.ts#L818

and here: https://github.com/bitmovin/bitmovin-player-ui/blob/54df6e1626b26d70d5b2c70cb58126340377c719/src/ts/UIConfig.ts#L317-L318

src/ts/UIFactory.ts
stonko1994: Yes, that is the idea. [721a2ff](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/721a2ffab19b4b39321b16013b0262f0339c2a17)

CHANGELOG.md
stonko1994: [ee3212c](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/ee3212c36940ef5d82791bcb3310027cee047850)

src/ts/UIConfig.ts
stonko1994: [ee3212c](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/ee3212c36940ef5d82791bcb3310027cee047850)

src/ts/UIConfig.ts
stonko1994: [ee3212c](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/ee3212c36940ef5d82791bcb3310027cee047850)

src/ts/UIComponentLayoutOverrides.ts
stonko1994: Added some clarifying docs in [6f5bbd7](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/6f5bbd73e01a02e69fe12ea28f2b651073ab9713)

src/ts/UIComponentLayoutOverrides.ts
stonko1994: [6f5bbd7](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/6f5bbd73e01a02e69fe12ea28f2b651073ab9713)

src/ts/UIComponentLayoutOverrides.ts
stonko1994: I removed the unsupported types and added a test to cover this: [6f5bbd7](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/6f5bbd73e01a02e69fe12ea28f2b651073ab9713)

src/ts/UIFactory.ts
dweinber: From a customer perspective, showing a "-10" button and hiding "+10" button wouldn't be more complex than e.g. showing the fullscreen button but hiding the PIP button. I know and understand why it is more complex in our implementation, but it's still hard to argue IMO.

Anyway, I'm fine with leaving that detail out for the initial version.

src/ts/components/buttons/QuickSeekButton.ts
123mpozzi: I see, fair to keep it as is then 👍

123mpozzi (approved):
Nothing blocking from my side 💪 . Please just make sure the remaining open threads are resolved

src/ts/utils/ComponentLayoutOverrideProcessor.ts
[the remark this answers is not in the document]
dweinber: @stonko1994 I think that's a fair point. I guess it should be too much effort to add that handling, or do you see any issues with that?

dweinber (changes requested):
Looks like there is one more thing, and then we should be there from my perspective 🙌

src/ts/utils/ComponentLayoutOverrideProcessor.ts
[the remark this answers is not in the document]
stonko1994: [d212a76](https://github.com/bitmovin/bitmovin-player-ui/pull/910/commits/d212a76ff1744cdd7ee287a1f36f8ce14b734192)

dweinber (approved):

PR #911, opened 2026-07-13
fix(spatialnavigation): keep focus contained at navigation group edges on TVs

nebrey:
### The bug

On some smart TVs (reported on a Sony Bravia, Android 12), spatial navigation no longer keeps focus inside a `NavigationGroup`. Pressing a direction key at the edge of a group — e.g. `UP` when there is no focusable element above the current one *within the group* — moves focus to an element **outside** the group, because the TV browser applies its own native focus handling.

This was introduced in [`4d55c39`](https://github.com/bitmovin/bitmovin-player-ui/commit/4d55c394bcfead7a89e4ee1bce71d236f0af7079) ("Fix back button not working on TVs like Tizen"). That change made `SpatialNavigation.handleKeyEvent` call `preventDefault()` / `stopPropagation()` only when the active group *consumed* the event:

```ts
if (isDirection(event) && active.handleNavigation(event)) {
  e.preventDefault();
  e.stopPropagation();
}
```

At a group edge, `handleNavigation()` returns `false` (no focusable target in that direction), so `preventDefault()` is skipped, the keydown propagates, and the platform moves focus natively out of the group. Previously `preventDefault()` was always called for directional keys, so focus stayed contained.

### The fix

Directional keys are now always consumed while a navigation group is active, regardless of whether focus actually moved — restoring the pre-regression containment behavior:

```ts
if (isDirection(event)) {
  active.handleNavigation(event);
  e.preventDefault();
  e.stopPropagation();
} else if (isAction(event) && active.handleAction(event)) {
  e.preventDefault();
  e.stopPropagation();
}
```

Actions (e.g. `BACK`) intentionally remain *conditionally* consumed, so an unhandled action still propagates to the platform — this preserves the back-button behavior on TVs like Tizen that `4d55c39` was added to fix (the `RootNavigationGroup` returns `false` for `BACK` when the UI is already hidden, letting the TV handle it).

### Testing

- Added a spec asserting a directional key calls `preventDefault()` / `stopPropagation()` even when `handleNavigation()` returns `false`.
- All spatial-navigation specs pass (`npx jest spec/spatialnavigation`); lint clean.
- Not yet verified on physical TV hardware — appreciate a check on a reference smart TV.

- [x] `CHANGELOG` entry


stonko1994 (commented):
Code looks good however I'd have a question: what about the use-case where the player is **not** in fullscreen and there are additional elements outside of the player view?

Wouldn't the change _again_ prevent navigating to anything outside of the player? I'm also wondering which element the TV browser is exactly selecting if there is none in the described problematic case 🤔

nebrey:
@stonko1994 Claude says

---
Good question. Two things bound this:

**Scope:** `SpatialNavigation` is only instantiated by the `tv`/`tvAds` UI variants (`UIFactory`), i.e. via `buildTvUI()`. The default web UI never wires it up, so an embedded desktop player with surrounding page elements never hits this code path — arrow keys stay native there. The containment only applies to consumers who opted into the TV UI.

**Not new behavior:** this is an exact restoration of the pre-`4d55c39` directional handling, which always called `preventDefault()` for directional keys. `4d55c39` made it conditional and that's what let focus escape at group edges. Actions stay conditional here, so the Tizen back-button fix from `4d55c39` is preserved. So the "can't arrow out of the player" behavior for the TV UI is how it worked for years before the regression, not something this PR introduces.

Also worth noting the handler early-returns when the container `isHidden()`/`isDisabled()` (`SpatialNavigation.ts:125`), so once controls auto-hide, directional keys propagate natively again — containment only applies while the controls are actually shown.

On which element the TV grabs in the broken case: with `preventDefault()` skipped, the keydown propagates and the Android WebView's native `focusSearch` picks the next focusable node in that direction — outside the player container. It's platform-dependent/non-deterministic, which is the core reason the UI should own directional focus rather than defer to the native handler. If a TV-UI consumer genuinely needs to escape the group at an edge, that's what the `onNavigation`/`afterNavigation` hooks on `NavigationGroup` are for.

stonko1994 (approved):
Sounds reasonable 👍

PR #921, opened 2026-07-30
Make live indicator keyboard accessible

123mpozzi:
## Description

The LIVE button is not accessible via keyboard.

### Changes

- Make the LIVE indicator keyboard focusable and expose button semantics
  - Return to the live edge when Enter or Space is pressed
  - Add a visible, button-sized focus highlight
  - Restore non-interactive semantics when switching to VOD
- Add regression coverage for keyboard behavior, cleanup, and compiled styles

 ### Manual Testing Instructions

  1. In the Demo page, select a live video source
  1. Seek back behind the live edge
  1. Press Tab until the LIVE indicator is focused
  1. Confirm it has a clearly visible rounded focus highlight
  1. Press Enter and confirm playback returns to the live edge
  1. Move behind the live edge again, focus LIVE, and press Space
  1. Confirm playback returns to the live edge and the page does not scroll
  1. Switch to a VOD source and confirm the time label is no longer exposed as a focusable button

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- The LIVE indicator can now be focused and activated with Enter or Space to return to the live edge.
+
 ## [4.17.0] - 2026-07-02
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- `live.jumpToLiveEdge` localization key for the LIVE indicator's accessible name; interpolates the `live` label via `{liveLabel}`.
+
 ### Fixed
 
 - The LIVE indicator can now be focused and activated with Enter or Space to return to the live edge.
--- a/spec/components/labels/PlaybackTimeLabel.spec.ts
+++ b/spec/components/labels/PlaybackTimeLabel.spec.ts
@@ -24,6 +24,79 @@ describe('PlaybackTimeLabel', () => {
       playbackTimeLabel = new PlaybackTimeLabel();
     });
 
+    describe('keyboard accessibility', () => {
+      let mockDomElement: ReturnType<typeof MockHelper.generateDOMMock>;
+
+      beforeEach(() => {
+        mockDomElement = MockHelper.generateDOMMock();
+        jest.spyOn(playbackTimeLabel, 'getDomElement').mockReturnValue(mockDomElement);
+
+        playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
+      });
+
+      it('exposes the live indicator as an accessible button', () => {
+        expect(mockDomElement.attr).toHaveBeenCalledWith('tabindex', '0');
+        expect(mockDomElement.attr).toHaveBeenCalledWith('role', 'button');
+        expect(mockDomElement.attr).toHaveBeenCalledWith('aria-label', 'Jump to live edge');
+      });
+
+      it.each(['Enter', ' '])('jumps to the live edge when pressing %p', key => {
+        const keydownHandler = mockDomElement.on.mock.calls.find(([eventName]) => eventName === 'keydown')?.[1] as (
+          event: KeyboardEvent,
+        ) => void;
+        const keyboardEvent = {
+          key,
+          preventDefault: jest.fn(),
+        } as unknown as KeyboardEvent;
+
+        expect(keydownHandler).toEqual(expect.any(Function));
+        keydownHandler(keyboardEvent);
+
+        expect(keyboardEvent.preventDefault).toHaveBeenCalled();
+        expect(playerMock.timeShift).toHaveBeenCalledWith(0);
+      });
+
+      it('ignores unrelated keys', () => {
+        const keydownHandler = mockDomElement.on.mock.calls.find(([eventName]) => eventName === 'keydown')?.[1] as (
+          event: KeyboardEvent,
+        ) => void;
+        const keyboardEvent = {
+          key: 'ArrowRight',
+          preventDefault: jest.fn(),
+        } as unknown as KeyboardEvent;
+
+        keydownHandler(keyboardEvent);
+
+        expect(keyboardEvent.preventDefault).not.toHaveBeenCalled();
+        expect(playerMock.timeShift).not.toHaveBeenCalled();
+      });
+
+      it('restores non-interactive semantics when switching to VOD', () => {
+        jest.spyOn(playerMock, 'isLive').mockReturnValue(false);
+
+        playerMock.eventEmitter.fireDurationChangedEvent();
+
+        expect(mockDomElement.attr).toHaveBeenCalledWith('tabindex', '-1');
+        expect(mockDomElement.removeAttr).toHaveBeenCalledWith('role');
+        expect(mockDomElement.removeAttr).toHaveBeenCalledWith('aria-label');
+        expect(mockDomElement.off).toHaveBeenCalledWith('keydown', expect.any(Function));
+      });
+    });
+
+    it('restores the configured tab index when switching to VOD', () => {
+      const mockDomElement = MockHelper.generateDOMMock();
+      playbackTimeLabel = new PlaybackTimeLabel({
+        tabIndex: 2,
+      });
+      jest.spyOn(playbackTimeLabel, 'getDomElement').mockReturnValue(mockDomElement);
+      playbackTimeLabel.configure(playerMock, uiInstanceManagerMock);
+      jest.spyOn(playerMock, 'isLive').mockReturnValue(false);
+
+      playerMock.eventEmitter.fireDurationChangedEvent();
+
+      expect(mockDomElement.attr).toHaveBeenCalledWith('tabindex', '2');
+    });
+
     describe('switch to inactive', () => {
       let removeClassSpy: any;
       beforeEach(() => {
--- a/spec/components/labels/PlaybackTimeLabel.spec.ts
+++ b/spec/components/labels/PlaybackTimeLabel.spec.ts
@@ -72,6 +72,7 @@ describe('PlaybackTimeLabel', () => {
       });
 
       it('restores non-interactive semantics when switching to VOD', () => {
+        const keydownHandler = mockDomElement.on.mock.calls.find(([eventName]) => eventName === 'keydown')?.[1];
         jest.spyOn(playerMock, 'isLive').mockReturnValue(false);
 
         playerMock.eventEmitter.fireDurationChangedEvent();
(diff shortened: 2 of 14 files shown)


src/ts/components/labels/PlaybackTimeLabel.ts
```diff
@@ -77,6 +77,29 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
       player.timeShift(0);
     };
 
+    const liveKeyDownHandler = (event: KeyboardEvent) => {
+      if (event.key === 'Enter' || event.key === ' ') {
+        event.preventDefault();
+        liveClickHandler();
+      }
+    };
```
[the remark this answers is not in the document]
123mpozzi: Addressed in dc1d9988 by supporting `event.code === 'Space'`, suppressing repeated activation while still preventing the default action, and adding regression tests.

src/ts/components/labels/PlaybackTimeLabel.ts
```diff
@@ -77,6 +77,33 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
...
+        }
+      }
+    };
+
+    const enableLiveInteraction = () => {
+      const domElement = this.getDomElement();
+      domElement.attr('tabindex', '0');
+      domElement.attr('role', 'button');
+      // Describe the action rather than the displayed live state for assistive technologies.
+      this.setAriaLabel(i18n.getLocalizer('live.jumpToLiveEdge'));
```
jmsn: **Must-fix:** the accessible name drops the visible label text

**TL;DR** — `role="button"` + `aria-label` makes the name override the element's text, so "Live" leaves the accessibility tree and the control can't be addressed by what it displays. Composing the name from the `live` key fixes both.

Before this change the label is a plain `<div>` whose text *is* what a screen reader reads. After it, the name is "Jump to live edge" and the live state isn't exposed at all — and speech-input users ("click Live") can't activate the button this PR just made activatable. `PlayerInsightsContextMenuItem.ts:56-62` already handles a text-bearing control this way: one localizer feeds both `setText` and `setAriaLabel`.

The existing placeholder mechanism keeps `live` as the single source of truth, so custom vocabularies and partial translations follow automatically:

`"live.jumpToLiveEdge": "{liveLabel}, jump to live edge"`

```ts
const liveLabel = i18n.performLocalization(i18n.getLocalizer('live'));
this.setAriaLabel(i18n.getLocalizer('live.jumpToLiveEdge', { liveLabel }));
```

<details><summary>Verified in the demo</summary>

Default UI, `liveWithTimeshift`, focused live indicator:

`{"textContent":"Live","role":"button","aria-label":"Jump to live edge","tabindex":"0"}`

`aria-label` outranks name-from-content in the accessible name computation, so the rendered "Live" never reaches the name.

Note `replaceVariableWithPlaceholderIfExists` (`i18n.ts:247`) reads `.length` on `text.match(...)`, which is `null` when the string has no placeholder — so an override of this key must keep `{liveLabel}`.
</details>

— 🤖 Claude Code (Claude Opus 5)

src/ts/components/labels/PlaybackTimeLabel.ts
```diff
@@ -89,12 +116,14 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
           this.hide();
         }
         this.onClick.subscribe(liveClickHandler);
+        enableLiveInteraction();
```
jmsn: **Suggestion:** move the click wiring into the new helpers

`enableLiveInteraction`/`disableLiveInteraction` own the keyboard and ARIA half while the `onClick` subscribe/unsubscribe stays inline, so the names claim more than they do and one concern is wired in two places. Moving the `onClick` pair in fixes both.

— 🤖 Claude Code (Claude Opus 5)

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- The LIVE indicator can now be focused and activated with Enter or Space to return to the live edge.
```
jmsn, line 21: **Doc:** add the new localization key under `### Added`

**TL;DR** — `live.jumpToLiveEdge` lands in the exported `Vocabulary` interface, which AGENTS.md treats as public API — integrators maintaining their own vocabularies need to know it exists.

```markdown
### Added

- `live.jumpToLiveEdge` localization key for the LIVE indicator's accessible name; interpolates the `live` label via `{liveLabel}`
```

Precedent: `CHANGELOG.md:189`. The `{liveLabel}` clause assumes the placeholder fix above lands.

— 🤖 Claude Code (Claude Opus 5)

spec/components/labels/PlaybackTimeLabelStyles.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+      throw new Error(`Could not find ${property} for ${selector} in compiled CSS`);
+    }
+
+    return value;
+  };
+
+  it('renders a visible keyboard focus state for the live indicator', () => {
+    expect(css).toMatch(/\.bmpui-ui-playbacktimelabel\.bmpui-ui-playbacktimelabel-live\.bmpui-focus-visible\s*\{/);
+  });
+
```
jmsn, line 31: **Suggestion:** make the padding invariant structural instead of asserted

**TL;DR** — this test asserts that `calc($icon-size * 0.25)` was typed identically in `_button.scss:16` and `_playback-time-label.scss:13`. Sharing the value removes the duplication and the need for the test.

```scss
// _variables.scss
$control-padding: calc($icon-size * 0.25) !default;
```

Both stylesheets then use `$control-padding` and this test can go — and integrators get an overridable variable, consistent with the rest of `_variables.scss`.

— 🤖 Claude Code (Claude Opus 5)

src/ts/components/labels/PlaybackTimeLabel.ts
```diff
@@ -89,12 +116,14 @@ export class PlaybackTimeLabel extends Label<PlaybackTimeLabelConfig> {
           this.hide();
```
jmsn, line 119: NIT — with `hideInLivePlayback: true` (the default UI's CurrentTime label) this hides the label and then gives it `tabindex="0" role="button" aria-label`; moving `enableLiveInteraction()` into an `else` avoids the inert second "Jump to live edge" button in the DOM.

— 🤖 Claude Code (Claude Opus 5)

spec/components/labels/PlaybackTimeLabel.spec.ts
```diff
@@ -24,6 +24,113 @@ describe('PlaybackTimeLabel', () => {
...
+      });
+
+      it('exposes the live indicator as an accessible button', () => {
+        expect(mockDomElement.attr).toHaveBeenCalledWith('tabindex', '0');
+        expect(mockDomElement.attr).toHaveBeenCalledWith('role', 'button');
+        expect(mockDomElement.attr).toHaveBeenCalledWith('aria-label', 'Jump to live edge');
+      });
+
+      it.each(['Enter', ' '])('jumps to the live edge when pressing %p', key => {
+        const keydownHandler = mockDomElement.on.mock.calls.find(([eventName]) => eventName === 'keydown')?.[1] as (
```
jmsn: NIT — this three-line lookup repeats at 60, 76, 93 and 108; a single `getKeydownHandler()` helper in the enclosing `describe` covers all five.

— 🤖 Claude Code (Claude Opus 5)

jmsn (changes requested):
## Code review

**Not approved** — one must-fix (the accessible name), plus a few suggestions and nits inline.

The mechanism works. Verified in the demo (default UI, `liveWithTimeshift`): the live label gets `tabindex="0" role="button"`, `FocusVisibilityTracker` adds `bmpui-focus-visible` on focus, Enter and Space each drive `getTimeShift()` from `-30` back to `0`, and the VOD transition restores `tabindex="-1"` with no `role`/`aria-label`. `npx jest` (17 pass), `lint-ts`, `lint-sass` and `tsc --noEmit` are all clean.

🤖 Review generated with [Claude Code](https://claude.com/claude-code) using Claude Opus 5

src/ts/components/labels/PlaybackTimeLabel.ts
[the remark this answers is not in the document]
jmsn: Same predicate already exists at `InteractiveSettingsPanelItem.ts:33-41`, which also accepts `Action.SELECT` from `getKeyMapForPlatform()` — worth a look if you want them aligned (that one lacks the `repeat` handling you added here).

— 🤖 Claude Code (Claude Opus 5)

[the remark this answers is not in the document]
123mpozzi: Thanks. I’m keeping this handler local because PlaybackTimeLabel is not part of the settings/spatial-navigation action pipeline. Adding `getKeyMapForPlatform()` and `Action.SELECT` would expand the scope and platform behavior of this PR. 

The missing repeat handling in `InteractiveSettingsPanelItem` is pre-existing and is better handled separately.

src/ts/components/labels/PlaybackTimeLabel.ts
123mpozzi: Address in 34a3e228. The accessible name now includes the localized visible live label via `{liveLabel}`

src/ts/components/labels/PlaybackTimeLabel.ts
123mpozzi: Address in 34a3e228. Click subscription and cleanup now live inside `enableLiveInteraction()` and `disableLiveInteraction()` so all the interaction lifecycle is kept together

CHANGELOG.md
123mpozzi: Good point! Address in 34a3e228

spec/components/labels/PlaybackTimeLabelStyles.spec.ts
123mpozzi: Thanks. I’d keep the compiled-CSS assertion here. Equality with button padding is the behavior we want to preserve, while a new `!default` variable would expand the public theming API and permanently couple two components for a single use. I’d revisit the abstraction if this spacing becomes shared more broadly

src/ts/components/labels/PlaybackTimeLabel.ts
123mpozzi: Address in 34a3e228. `enableLiveInteraction()` now only runs for the visible live label

spec/components/labels/PlaybackTimeLabel.spec.ts
123mpozzi: Address in 34a3e228. I extracted a local helper and reused it across the keyboard interaction tests

jmsn (commented):
## Re-review

**Approved.** Everything from the prior review is addressed or settled, and CI is green.

The accessible name fix is verified across all six default vocabularies, with word order adapted per language rather than mechanically prefixed (`es: 'Directo'` → `'{liveLabel}, saltar al punto en directo'`). The cross-language spec in `Localization.spec.ts` is a better guard than I asked for.

Accepting both declines — keeping the key handler local, and keeping the compiled-CSS assertion over a new `!default` variable.

🤖 Review generated with [Claude Code](https://claude.com/claude-code) using Claude Opus 5

jmsn (approved):
Approved — see the re-review above.

🤖 Approved via [Claude Code](https://claude.com/claude-code) using Claude Opus 5 at Jameson's explicit instruction.

PR #922, opened 2026-08-03
Add CaptionToggleButton

dweinber:
## Description

Adds a `CaptionToggleButton` that turns captions and subtitles on and off with a single click, without having to open the settings panel.

Behavior:

- Hides itself while the source provides no caption tracks, since there is nothing to toggle.
- Derives its state from the player, so it stays in sync when captions are switched elsewhere in the UI (e.g. the settings panel or the subtitle list box).
- When captions are switched on, it restores the track that was enabled last. If no track has been enabled yet, it prefers a track matching the current audio language and falls back to the first available track.
- Registers its player event handlers on the wrapped player of the UI instance, which releases them together with the UI.

### Availability and how to enable it

The button is part of the **default**, **small-screen**, and **TV** layouts, placed directly after the `Spacer` so it is the first component of the right-hand control bar group. It is **excluded by default** through the default component layout overrides, so the default UI is unchanged until it is opted in:

```ts
const uiConfig = {
  componentLayoutOverrides: {
    CaptionToggleButton: UIComponentLayoutOverride.Include,
  },
};
```

This builds on the `componentLayoutOverrides` API from #910, which is why this PR targets that branch.

The ad layouts (`ads`, `smallScreenAds`, `tvAds`) are intentionally left out because they have no `SubtitleOverlay` and therefore never display captions. `emptyState`, `castReceiver`, and `subtitle` are left out because they have no control bar to place the button in.

Ships two new icons, `closed-captions.svg` (off) and `closed-captionsX.svg` (on, inverted), following the existing `fullscreen`/`fullscreenX` naming and icon-swap convention.

Main UI variant:
Screenshot 2026-08-12 at 11 35 08
Screenshot 2026-08-12 at 11 35 20

TV UI variant:
Screenshot 2026-08-12 at 11 35 30
Screenshot 2026-08-12 at 11 40 03

Small Screen UI variant:
Screenshot 2026-08-12 at 11 36 58
Screenshot 2026-08-12 at 11 37 06

## Verification

- `npx tsc --noEmit`, `npm run lint-ts`, `npm run lint-sass` clean
- 662 tests / 58 suites pass, including specs asserting the button is present in all three layouts, excluded by the default overrides, and kept when explicitly included
- `npm run build` and `npm run docs` exit 0
- Verified in the demo page with the override enabled: the button renders as the first element of the right control bar group, toggles captions on and off, and picks the track matching the audio language

### Not covered

No Android or iOS SDK verification. This touches the `player.subtitles` API and the unit tests mock the player, so mobile SDK behavior is unverified.

- [x] `CHANGELOG` entry


src/ts/components/buttons/CaptionToggleButton.ts
```diff
@@ -0,0 +1,153 @@
...
+    super.configure(player, uimanager);
+    this.player = player;
+    this.uimanager = uimanager;
+
+    player.on(player.exports.PlayerEvent.SourceLoaded, this.onPlayerStateChanged);
+    player.on(player.exports.PlayerEvent.SourceUnloaded, this.onSourceUnloaded);
+    player.on(player.exports.PlayerEvent.SubtitleAdded, this.onPlayerStateChanged);
+    player.on(player.exports.PlayerEvent.SubtitleRemoved, this.onPlayerStateChanged);
+    player.on(player.exports.PlayerEvent.SubtitleEnabled, this.onSubtitleEnabled);
+    player.on(player.exports.PlayerEvent.SubtitleDisabled, this.onPlayerStateChanged);
```
stonko1994: Usually we don't use this pattern to use 'this' for event callbacks. The drawback of this approach is that we need to capture the player to the component from within the configure method here. This adds the realease overhead etc.

It's common in the UI to simply but the whole 'lifecycle' handling into the configure method with local functions here.

I'd be in favor of this but not blocking.

stonko1994 (commented):
Should we already base this PR on #910 to make the enabling of this component easier?

src/ts/components/buttons/CaptionToggleButton.ts
dweinber: Done in d4e9222e. Moved the state and handlers into local functions in `configure()` and dropped `release()` along with the captured `player`/`uimanager`.

Worth noting for the record why this is safe: components are configured with the per-UI-instance wrapped player (`UIInstanceManager.configureControlsTree` -> `component.configure(this.getPlayer(), this)`), and that wrapper records every `on()` handler and unregisters them in `clearEventHandlers()` when the UI instance is released. So the closures are cleaned up by the framework and the manual `off()` calls were redundant.

Replaced the three `release()` specs with two that assert the handlers are registered on the player the component is configured with.

dweinber:
Re: basing this on #910 — I get your point and thought of that already. Happy to do it.

src/ts/components/buttons/CaptionToggleButton.ts
dweinber: Implemented in d4e9222e

dweinber:
Addressed the suppressed review comment on `CaptionToggleButtonLayout.spec.ts` in 5f78d5c0.

The regex stopped at the first conditional spread, so it only saw defaults declared above `...(uiconfig.ecoMode != null && { ... })`. It happened to capture `CaptionToggleButton` today purely because that key sorts alphabetically before the first spread — so the assertion was passing by luck rather than by construction. Now matching up to the closing `...uiconfig.componentLayoutOverrides` spread instead.

Verified both directions by mutating `UIManager.ts` locally:

- Removing the `CaptionToggleButton` default → the test **fails** (the guard still guards).
- Moving the default *below* the first conditional spread → the test **passes**, where it silently produced a false negative before.

The widened block also spans the conditional spreads, but those assign through ternaries (`cond ? Include : Exclude`), and the extraction pattern requires the enum immediately after the colon, so they contribute no entries and cannot overwrite the real defaults.

dweinber:
Addressed the second suppressed comment in 7b70283b — though the stated reasoning does not hold for the current code, so recording the check here.

**The claim as written is a false positive.** The conditional entries assign through ternaries, e.g. `EcoModeContainer: uiconfig.ecoMode ? UIComponentLayoutOverride.Include : ...`. The extraction pattern required `UIComponentLayoutOverride.X` immediately after the colon, and `uiconfig.ecoMode ?` sits in between, so those lines never matched. I verified this against the real block and against the prettier-wrapped multi-line form that `PlaybackSpeedSelectBox` already uses — zero matches in both. `EcoModeContainer` was never overwritten.

**The underlying risk is still real, so I took the fix.** If a conditional spread ever assigned a plain enum rather than a ternary, it *would* have overwritten a base default, and the guard would silently start asserting the wrong thing. Relying on ternary formatting for correctness is too subtle for a test whose whole job is to fail when a default disappears.

The parser is now anchored to the block's own indentation (`^ {8}Key: UIComponentLayoutOverride.Value,$`), so nested keys are excluded structurally regardless of formatting. Confirmed with a synthetic block where a spread assigns a plain `EcoModeContainer: UIComponentLayoutOverride.Include` — the default still resolves to `Exclude`.

Mutation checks re-run and unchanged: removing the `CaptionToggleButton` default fails the test; moving it below a conditional spread passes.

dweinber:
@stonko1994 re-requesting your review. Copilot came back clean on 7b70283b (no comments, none suppressed).

Since your last review on 014af9bb, the substantive changes are:

**Your feedback**
- Lifecycle moved into `configure()` with local functions and local state; `release()` and the captured `player`/`uimanager` are gone (d4e9222e). The handlers are registered on the UI instance's wrapped player, which releases them with the UI, so the manual `off()` calls were redundant.
- Type-only player imports (014af9bb, yours).

**Default layout wiring** — this is the part worth your attention, since it is new since you last looked:
- The button is now in the **default**, **small-screen**, and **TV** layouts, as the first component after the `Spacer` (9ce5b28e, d965076f, f531a8e9).
- It is **excluded by default** via the default `componentLayoutOverrides` in `UIManager`, so the default UI is unchanged until integrators opt in. This is why the PR is based on #910.
- Ad layouts are deliberately left out: `ads`, `smallScreenAds`, and `tvAds` have no `SubtitleOverlay`, so captions are never displayed there. `emptyState`, `castReceiver`, and `subtitle` have no control bar to place the button in.

stonko1994 (approved):

PR #927, opened 2026-08-10
Fix seek bar shrinking to zero width during live playback

123mpozzi:
## Description

Fixes a `4.18.0` regression: during live playback the seek bar loses width on every time update.

To stop the time label jumping around with non-monospaced fonts, `PlaybackTimeLabel` remembers the widest it has been and applies that as `min-width`. It measures with `offsetWidth`, which includes padding, but `min-width` under the default `content-box` excludes it.

With no padding the two matched. #921 added padding for the focus target, so each update now stores a value `12px` too large, the label grows, and the `SeekBar` next to it loses the same `12px`.

The issue is not reproducible on our demo page because it loads Bootstrap, which makes the label already `border-box`. Pages without one, including the iOS SDK WebView, are affected.

### Changes

This PR sets `box-sizing: border-box` on the label so both numbers describe the same box, instead of relying on the embedding page's reset.

### How to reproduce

As said, `index.html` cannot show this. It loads Bootstrap, whose `box-sizing: inherit` reset makes the label `border-box` and hides the bug. Use `simple.html`, which has no reset.

Point `src/html/simple.html` at a live source:

```js
var source = {
  hls: 'https://bitcdn-kronehit.bitmovin.com/v2/hls/playlist.m3u8',
};
```

Then `npm start` and open `simple.html`. Any live source works, a timeshift window is not needed. Watch the control bar for `~10` seconds.

- on `develop`: the seek bar shrinks on every time update
- on this branch: seek bar holds its size

- [x] `CHANGELOG` entry

### DIFF
--- a/src/scss/components/labels/_playback-time-label.scss
+++ b/src/scss/components/labels/_playback-time-label.scss
@@ -3,8 +3,7 @@
 .#{$prefix}-ui-playbacktimelabel {
   @extend %ui-label;
 
-  // The label sets its own min-width from its offsetWidth. offsetWidth counts padding and
-  // min-width does not, so without border-box the label gets wider on every time update.
+  // PlaybackTimeLabel writes offsetWidth back as min-width, so both must describe the same box.
   box-sizing: border-box;
   font-weight: $font-weight-medium;
   text-transform: uppercase;
--- a/src/scss/components/labels/_playback-time-label.scss
+++ b/src/scss/components/labels/_playback-time-label.scss
@@ -3,7 +3,8 @@
 .#{$prefix}-ui-playbacktimelabel {
   @extend %ui-label;
 
-  // PlaybackTimeLabel writes offsetWidth back as min-width, so both must describe the same box.
+  // PlaybackTimeLabel writes its border-box offsetWidth back as min-width. Keep min-width
+  // border-box too, or padding is added again and the measurement grows on every update.
   box-sizing: border-box;
   font-weight: $font-weight-medium;
   text-transform: uppercase;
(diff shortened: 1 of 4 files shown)


src/scss/components/labels/_playback-time-label.scss
```diff
@@ -3,6 +3,9 @@
 .#{$prefix}-ui-playbacktimelabel {
   @extend %ui-label;
 
+  // The label sets its own min-width from its offsetWidth. offsetWidth counts padding and
+  // min-width does not, so without border-box the label gets wider on every time update.
```
stonko1994: > gets wider on every time update

Is the issue description (symptom/impact) relevant enough to keep in the codebase? I'd phrase this comment more like the technical impact rather than the visible impact


```suggestion
  // The label sets its own min-width in the component. `offsetWidth` counts padding and
  // min-width does not, so without border-box the label width changes dynamically.
```

src/ts/components/labels/PlaybackTimeLabel.ts
stonko1994, line 169: I wonder if the CSS approach is the correct fix or if we should fix this here? (Just a question, no call to action)

src/scss/components/labels/_playback-time-label.scss
123mpozzi: Good point. I rephrased it around the box-model feedback loop while keeping border-box and the per-update growth explicit

src/ts/components/labels/PlaybackTimeLabel.ts
123mpozzi: I think CSS is the better fit. `offsetWidth` measures the border box and is written back as `min-width`, so declaring `border-box` aligns them and removes the dependency on the embedding page’s reset.

stonko1994 (approved):

PR #939, opened 2026-09-04
Fix duplicate ad break tracking across deserialized objects

jarhoax:
## Description

Deduplicate retained ad breaks by their stable ID so adapters that provide freshly deserialized objects do not inflate the displayed ad count. Includes regression coverage for repeated `AdStarted` events with distinct objects representing the same break.

- [x] `CHANGELOG` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- Ad counts no longer increase incorrectly when player adapters provide freshly deserialized instances of the same ad break.
+
 ## [4.20.2] - 2026-09-03
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-- Ad counts no longer increase incorrectly when player adapters provide freshly deserialized instances of the same ad break.
+- Ad count in `AdCounterLabel` no longer increases incorrectly when used on mobile SDKs.
 
 ## [4.20.2] - 2026-09-03
 
--- a/src/ts/utils/AdBreakTracker.ts
+++ b/src/ts/utils/AdBreakTracker.ts
@@ -129,8 +129,9 @@ export class AdBreakTracker {
 
       this.groupScheduleTime = activeBreak.scheduleTime;
 
-      // Add the active break if it's not already retained (new break in the group)
-      if (!this.groupBreaks.includes(activeBreak)) {
+      // Player adapters may deserialize a new object for the same break on every event,
+      // so use the stable break ID instead of object identity for deduplication.
+      if (!this.groupBreaks.some(adBreak => adBreak.id === activeBreak.id)) {
         this.groupBreaks.push(activeBreak);
       }
     } else {
--- a/src/ts/utils/AdBreakTracker.ts
+++ b/src/ts/utils/AdBreakTracker.ts
@@ -129,7 +129,7 @@ export class AdBreakTracker {
 
       this.groupScheduleTime = activeBreak.scheduleTime;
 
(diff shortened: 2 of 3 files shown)


src/ts/utils/AdBreakTracker.ts
```diff
@@ -129,8 +129,9 @@ export class AdBreakTracker {
 
       this.groupScheduleTime = activeBreak.scheduleTime;
 
-      // Add the active break if it's not already retained (new break in the group)
-      if (!this.groupBreaks.includes(activeBreak)) {
+      // Player adapters may deserialize a new object for the same break on every event,
+      // so use the stable break ID instead of object identity for deduplication.
+      if (!this.groupBreaks.some(adBreak => adBreak.id === activeBreak.id)) {
         this.groupBreaks.push(activeBreak);
       }
```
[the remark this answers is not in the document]
jarhoax: I can add this defensive approach, although per contract, `AdBreak.id` is required, and will be generated in case it is missing.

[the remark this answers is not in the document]
jarhoax: bc46877121ae19a42858084be57d2ef329028e63

src/ts/utils/AdBreakTracker.ts
```diff
@@ -129,8 +129,12 @@ export class AdBreakTracker {
 
       this.groupScheduleTime = activeBreak.scheduleTime;
 
-      // Add the active break if it's not already retained (new break in the group)
-      if (!this.groupBreaks.includes(activeBreak)) {
+      // Player adapters may deserialize a new object for the same break on every event,
```
stonko1994: Let's not introduce information about mobile details in the web UI.


```suggestion
      // On mobile, the Player may deserialize a new object for the same break on every event,
```

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- Ad counts no longer increase incorrectly when player adapters provide freshly deserialized instances of the same ad break.
```
```suggestion
- Ad count in `AdCounterLabel` no longer increases incorrectly when used on mobile SDKs
```

src/ts/utils/AdBreakTracker.ts
```diff
@@ -129,8 +129,12 @@ export class AdBreakTracker {
...
+      // so use the stable break ID instead of object identity for deduplication.
+      const activeBreakAlreadyRetained = this.groupBreaks.some(adBreak =>
```
stonko1994: Nit nit nit: The current naming is a mouthfull

```suggestion
      const containsAdBreak = this.groupBreaks.some(adBreak =>
```

stonko1994 (commented):
I tested the current changes in the iOS SDK with and without those changes and I couldn't see any issues 🤔 Can you elaborate on what exactly is the comparison issues (just for understanding - not saying the changes don't make sense)

jarhoax:
> I tested the current changes in the iOS SDK with and without those changes and I couldn't see any issues 🤔 Can you elaborate on what exactly is the comparison issues (just for understanding - not saying the changes don't make sense)

It could happen if multiple breaks are scheduled at the same position, containing multiple ads.
https://github.com/bitmovin-engineering/player-ios/pull/4860/changes#diff-4804764380c720971069307e21b11a38130781d1596f24e43683c4dfad302f5eR109

We also put a known issue in the changelog until this is fixed:
https://github.com/bitmovin-engineering/player-ios/pull/4860#discussion_r3934581430

src/ts/utils/AdBreakTracker.ts
jarhoax: 99eee224711fcab35e7ade42bdd2e1b8802e395c

CHANGELOG.md
jarhoax: 7e1328c59346e7b3a89fa9c4fb2b456ea978045a

src/ts/utils/AdBreakTracker.ts
jarhoax: 052c464b66ca0fe4cec47920ba78b2c2fe13d40b

stonko1994 (approved):

PR #941, opened 2026-09-08
Fix settings panel overflow with long option names

jarhoax:
## Description
Fix settings panels overflowing the player or making setting labels unreadable when option names are long. Keep setting labels visible, truncate selected values, and wrap full names in the options list. 

Also remove the available subtitle count from the selected track label.

## Validation
TypeScript and Sass lint, TypeScript checks, and development build passed. 
Manual testing of the layout fix worked well

| Before | After |
| --- | --- |
| Before | After |

- [x] `CHANGELOG` entry

### DIFF
--- a/src/ts/components/settings/DynamicSettingsPanelItem.ts
+++ b/src/ts/components/settings/DynamicSettingsPanelItem.ts
@@ -1,9 +1,8 @@
 import { Label, LabelConfig, LabelStyle } from '../labels/Label';
 import { UIInstanceManager } from '../../UIManager';
 import { PlayerAPI } from 'bitmovin-player';
-import { i18n, LocalizableText } from '../../localization/i18n';
+import { LocalizableText } from '../../localization/i18n';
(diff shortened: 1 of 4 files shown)


src/ts/components/settings/DynamicSettingsPanelItem.ts
```diff
@@ -110,13 +109,7 @@ export class DynamicSettingsPanelItem extends InteractiveSettingsPanelItem<Dynam
       return;
     }
 
-    let selectedOptionLabelText = selectedItem.label;
-    if (this.settingComponent instanceof SubtitleSelectBox) {
-      const availableSettings = this.settingComponent.getItems().length;
-      selectedOptionLabelText =
-        i18n.performLocalization(selectedOptionLabelText) + ' (' + (availableSettings - 1) + ')';
```
jarhoax, line 117: I don't think this adds much value, but rather confuses people. There is no label or description showing what this actually is, so I removed it again. If it needs to stay, let me know. Can simply revert the commit.

stonko1994 (approved):

